PR #21, opened 2017-08-18
Battery bug fixes

nickdiorio:
Several fixes for flow battery, general battery behavior


nickdiorio:
Will build beta and have user give feedback on whether bugs are sufficiently addressed

PR #36, opened 2017-09-19
Fix a bug flagged by the warning: left operand of comma operator has no effect

### LINKED ISSUE #35: bug in Heliostat::getPanelById
As far as I can see, this function has never worked properly. g++ warns about this problem with the following messages:

../solarpilot/Heliostat.cpp:748:18: warning: left operand of comma operator has no effect [-Wunused-value]
../solarpilot/Heliostat.cpp:749:23: warning: left operand of comma operator has no effect [-Wunused-value]
../solarpilot/Heliostat.cpp:756:20: warning: left operand of comma operator has no effect [-Wunused-value]
../solarpilot/Heliostat.cpp:764:23: warning: left operand of comma operator has no effect [-Wunused-value]

I think this is a genuine bug, so submitting this issue (and I'll also submit a PR that fixes it).

bje-:
Fixes #35. 

* solarpilot/Heliostat.cpp (getPanelById): matrix_t::operator[] only
  takes a single int and returns the address of the row. Instead, use
  the public at() method to properly access matrix elements by row
  and column number.

### DIFF
--- a/solarpilot/Heliostat.cpp
+++ b/solarpilot/Heliostat.cpp
@@ -745,15 +745,14 @@ Reflector *Heliostat::getPanelById(int id){
 
 	for (int j=0; j<(int)ncantx; j++) {
 		for (int i=0; i<(int)ncanty; i++) {
-			if (_panels[j,i].getId() == id){
-				return &_panels[j,i];
+		  	if (_panels.at(j,i).getId() == id) {
+			  	return &_panels.at(j,i);
 			}
 		}
 	}
 
 	//#####call an error here
-
-	return &_panels[0,0];
+	return &_panels.at(0, 0);
 }
 
 Reflector *Heliostat::getPanel(int row, int col){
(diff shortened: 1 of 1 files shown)


jdpipe:
well spotted, ben!

solarpilot/Heliostat.cpp
```diff
@@ -745,23 +745,22 @@ Reflector *Heliostat::getPanelById(int id){
 
 	for (int j=0; j<(int)ncantx; j++) {
 		for (int i=0; i<(int)ncanty; i++) {
-			if (_panels[j,i].getId() == id){
-				return &_panels[j,i];
```
nickdiorio: Seems like a fairly insidious bug, from the perspective that an operator [] with two operands is not defined for a matrix_t, and so only the right operand gets evaluated according to https://stackoverflow.com/questions/54142/how-does-the-comma-operator-work .

nickdiorio: I agree, nice find!

PR #38, opened 2017-09-21
Fix warnings: statement has no effect [-Wunused-value].

bje-:
There are many instances in this patch of taking the address of a variable and casting the pointer to `void *` in the hope of silencing compiler warnings about the variable being unused. It would be far better to find out why the variable isn't used and fixing the root cause than to mask the problem with `void *` casts.


nickdiorio:
@mjwagner2 , do you have time to look at this?  Otherwise, I can go through and fix these warnings.  Darice had taken the unused returns and cast them to (void*) to get rid of the unused variable warnings, but this is somewhat of a hack.  Thanks!

mjwagner2:
I will take a look. We have sometimes kept these "unused" items to assist in debugging, as return values may not be used in the code but are useful in identifying a problem when stepping through the program. It is not a bad solution to cast as void in most cases, though some cases can be cleaned up further.

nickdiorio:
Great, thanks for looking.  I wonder if we could setup a debug log or some other mechanism to record these values if needed.

PR #45, opened 2017-09-28
Fix more warnings in the tcs directory

bje-:
An ad-hoc set of fixes for g++ warnings. Sending them in batches!

### DIFF
--- a/tcs/cavity_calcs.cpp
+++ b/tcs/cavity_calcs.cpp
@@ -141,12 +141,12 @@ void Cavity_Calcs::OuterPanel_Floor( double * F_AF )
 	double Ptheta, Pphi, theta, phi, r1, r2, x, y, y_i, z_i;
 	point p;
 	int hit;
-	int ray_counts = 1;
+	int ray_counts;
 
 	// Only set the seed once!
 	srand( (unsigned int)time( (time_t*)NULL) ); //Seed with time on init
 
-	for( ray_counts; ray_counts <= m_n_rays; ray_counts++ )
+	for( ray_counts = 1; ray_counts <= m_n_rays; ray_counts++ )
 	{
 		
 		for( int i = 0; i < m_n_nodes; i++ )
@@ -234,12 +234,12 @@ void Cavity_Calcs::InnerPanel_Floor( double * F_BF  )
 	double Ptheta, Pphi, theta, phi, r1, r2, x, y, y_i, z_i;
 	point p;
 	int hit;
-	int ray_counts = 1;
+	int ray_counts;
 
 	// Only set the seed once!
 	srand( (unsigned int)time( (time_t*)NULL) ); //Seed with time on init
 
-	for( ray_counts; ray_counts <= m_n_rays; ray_counts++ )
+	for( ray_counts = 1; ray_counts <= m_n_rays; ray_counts++ )
 	{
 		
 		for( int i = 0; i < m_n_nodes; i++ )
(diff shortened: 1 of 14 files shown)


tcs/cavity_calcs.cpp
```diff
@@ -141,12 +141,12 @@ void Cavity_Calcs::OuterPanel_Floor( double * F_AF )
 	double Ptheta, Pphi, theta, phi, r1, r2, x, y, y_i, z_i;
 	point p;
 	int hit;
-	int ray_counts = 1;
```
nickdiorio: Do these initializations cause warnings?

tcs/csp_solver_core.cpp
```diff
@@ -3109,9 +3108,6 @@ void C_csp_solver::Ssimulate(C_csp_solver::S_sim_setup & sim_setup)
 				double tol_C = 1.0;								//[C]
 				double tol = tol_C / m_cycle_T_htf_hot_des;		//[-]
 
-				double relaxed_tol_mult = 5.0;				//[-]
```
nickdiorio: These are defined above as "relaxed_tol_multiplier" and "relaxed_tol"

tcs/csp_solver_core.cpp
```diff
@@ -5028,8 +5023,6 @@ void C_csp_solver::Ssimulate(C_csp_solver::S_sim_setup & sim_setup)
 			mc_kernel.mc_sim_info.ms_ts.m_time_start = mc_kernel.mc_sim_info.ms_ts.m_time;	//[s]
 			mc_kernel.mc_sim_info.ms_ts.m_time = mc_kernel.get_baseline_end_time();			//[s]
 			mc_kernel.mc_sim_info.ms_ts.m_step = mc_kernel.mc_sim_info.ms_ts.m_time - mc_kernel.mc_sim_info.ms_ts.m_time_start;	//[s]
-
-			double blah = 1.23;
```
nickdiorio: These were probably used as debugging breakpoints, can always be added if necessary later

tcs/heat_exchangers.cpp
```diff
@@ -781,10 +779,6 @@ void NS_HX_counterflow_eqs::solve_q_dot_for_fixed_UA_enth(int hot_fl_code /*-*/,
 			throw(C_csp_exception("Off-design heat exchanger method failed"));
 		}
 	}
-	else
```
nickdiorio: This control structure isn't needed, it operates in the event that an exception isn't generated, which can be done by simply putting code at the end of the try-catch block

tcs/sam_lf_st_pt_type232.cpp
```diff
@@ -2673,7 +2673,6 @@ void PipeFlowCavity( double Re, double Pr, double LoverD, double relRough, doubl
 	    // given Reynolds number (Re), Prandtl number (Pr), the pipe length diameter ratio (LoverD)
 	    // and the relative roughness
 	    double Gz = Re*Pr/LoverD;	// Eq. 5-79 Nellis and Klein
-	    double x = LoverD/Re;		// Eq. 5-58 Nellis and Klein
```
nickdiorio: I will just comment out, as it's possible it's used for debugging

bje-:
On Mon, Oct 02, 2017 at 05:52:38PM +0000, Nick DiOrio wrote:

> > @@ -141,12 +141,12 @@ void Cavity_Calcs::OuterPanel_Floor( double * F_AF )
>  	double Ptheta, Pphi, theta, phi, r1, r2, x, y, y_i, z_i;
>  	point p;
>  	int hit;
> -	int ray_counts = 1;
> 
> Do these initializations cause warnings? 

No, but I movwd the initialisation of the loop index variable
(ray_counts) into the for loop. It's much clearer and eliminates
warnings about statements with no effect, eg:

  for (ray_counts; ray_counts < max; ray_counts++);

Cheers, Ben

nickdiorio:
Great, thanks!  Just checking for my own understanding.  Your changes have been merged.

PR #49, opened 2017-10-04
Fix compilation error on Linux (strcmp was not declared).

bje-:
* solarpilot/Toolbox.cpp (Toolbox::getDelimiter): Do not use strcmp. If we want to use the C++ standard library, then we should stop trying to escape to C. :-)


denise2201:
+1, we've been using this patch since day 1

PR #52, opened 2017-10-10
Using `my_isnan` macro for cross-platform support

jerluc:
It looks like at some point the `my_isnan` macro was defined to better support MS Visual Studio toolchains, but was never actually integrated into the `isnan` call later in the code. Then, it looks like that same code was modified to support builds on Linux in 36cb9cc3d43e85ff456e913c973db49f2a04baee.

This PR simply replaces that call with the macro instead so as to include support for both MSVSC and Linux.

### DIFF
--- a/shared/lib_weatherfile.cpp
+++ b/shared/lib_weatherfile.cpp
@@ -1436,7 +1436,7 @@ bool weatherfile::has_data_column( size_t id )
 }
 
 bool weatherfile::has_calculated_data(size_t id){
-	return !std::isnan(m_columns[id].data[0]);
+	return !my_isnan(m_columns[id].data[0]);
 }
 
 bool weatherfile::convert_to_wfcsv( const std::string &input, const std::string &output )


dguittet:
Thanks for catching that. It's a good fix to ensure compatibility with older visual studio versions, and I'll merge it. Additionally, the function is not being used anymore, as it was implemented for an error check that was not necessary in the end, so I will delete it and the one for weatherdata as well.

nickdiorio:
Thanks @jerluc for the contribution.  Would you mind sending me an email (nicholas.diorio@nrel.gov) with: 

I agree to contribute to SAM. I agree to the following terms and conditions for my contributions: First, I agree that I am licensing my contributions under the terms of the current SAM license. Second, I agree that, in order to conform to any future open source software license(s) under which SAM may be provided, the terms of my license may be modified without any notice to me and without my consent. Third, I represent and warrant that I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes my contributions, I represent and warrant that I have received permission to make contributions and grant the required license on behalf of my employer.

Thanks, and best regards.
Nick

PR #53, opened 2017-10-11
Fix more warnings in the tcs directory

bje-:


tyneises:
Modified and merged here: https://github.com/NREL/ssc/commit/7491b1ff483e4a14ba292895880b2b6064e95458

PR #61, opened 2017-10-20
Check that fscanf() returns exactly 3 input items

bje-:
This change silences a warning from g++ about not checking the result of fscanf(). The same problem persists in the current upstream version of lpsolve, so I have sent the same patch to the upstream maintainer.

Also, I suggest not compiling 3rd party packages with -Wall as it's better to ignore those warnings by default and fix them upstream.


nickdiorio:
That's a good call on -Wall for 3rd party packages, we should also do the same thing for nlopt, and possibly google test.

Issue #56, opened 2017-10-17
Possible logic bug in SolarField::radialStaggerPositions?


bje-:
g++ warns that:

    ../solarpilot/SolarField.cpp:2473:40: warning: logical not is only applied to the left hand side of comparison [-Wlogical-not-parentheses]

That line reads:

     && !Htv->is_round.mapval() == var_heliostat::IS_ROUND::ROUND) {

It's not clear to me if what is meant was to invert the mapval() result and then test equality with ROUND or if what was meant was !(expr == ROUND) or, simply, expr != ROUND.

If someone can please clarify, I'll fix it either way.

bje-:
I dug in some more and found that mapval() returns an enum that has only two possible values (0, 1). Hence the code happens to work either way -- you can invert the enum value and test it against ROUND or you can test the original enum value against ROUND and invert the equality result.  I will submit a patch to clear it up.

PR #65, opened 2017-10-20
Fix warning in solarpilot/SolarField.cpp (radialStaggerPositions).

bje-:
This patch fixes issue #56.


bje-:
I think this one may have been overlooked. Reopening.

PR #66, opened 2017-10-20
Fix memory leaks

bje-:
`delete [] a, b, c;` only deletes `a`. The comma operator applies to b and c, so those operands on the right-hand side of the commas are unused values.


nickdiorio:
Good catch

PR #71, opened 2017-10-25
Fix warning: ‘physics::AIR_DENSITY_SEA_LEVEL’ defined but not used

bje-:
This constant is defined because it calls `CelciusToKelvin`. Mark it (and `KelvinToCelcius`) as `constexpr` to silence this warning.


jdpipe:
PS "Celsius" with an "s"!!

nickdiorio:
Just found that Visual Studio 2013 doesn't support `constexpr`, https://stackoverflow.com/questions/20264644/constexpr-not-compiling-in-vc2013, but we are in the process of upgrading to Visual Studio 2017.  For now, I'll either undo, or ifdef around these.

PR #73, opened 2017-10-25
* lib_windfile.cpp: Rewrite trim() and locate2()

bje-:
I have rewritten these two functions to use C++ `strings` and `vector<string>` to make the code much simpler and clearer. There is no now arbitrary limitation of 128 columns in `locate2()`.


nickdiorio:
Note the continuous integration fails due to a failure in an unrelated file to `include <limits>`.  That problem has been fixed in b15d3f2b798000df22e4e72f6e1dd95ef0afad93

nickdiorio:
I inadvertently merged this PR (without the associated change) to fix a problem building `constexpr` with VS2013.  So this change is actually not merged.  It appears I cannot reopen.  I can either comment on the closed pull request.  Can you send another pull request with these changes?  Sorry for the confusion.

Issue #54, opened 2017-10-12
control reaches end of two non-void functions


bje-:
There are two functions that return `double`, but have execution paths that do not return a value. The warnings are:

    ../tcs/sco2_power_cycle.cpp:2061:1: warning: control reaches end of non-void function [-Wreturn-type]
    ../tcs/sco2_power_cycle.cpp:2067:1: warning: control reaches end of non-void function [-Wreturn-type]

Should these functions return a quiet NaN, perhaps, if frame == NULL?

bje-:
I'll close the issue as the patch covers it.

PR #74, opened 2017-10-27
Fix warning: control reaches end of non-void function

bje-:
Given that `frame` should never be `NULL`, I think it's fine to return `NaN` if it is. Fixes issue #54.


nickdiorio:
I merged aae9ea8, but it looks like 322cc74 is handled by another pull request, so will look at that one separately.  Thanks!

PR #75, opened 2017-10-27
Fix warning: deleting object  of polymorphic class type ‘AutoPilot_S’ which has non-virtual  destructor might cause undefined behaviour

bje-:
It is unsafe to delete an instance of a derived class through a pointer to a base class if the base class does not have a virtual destructor. This patch fixes the AutoPilot declaration and silences the warning,


nickdiorio:
Agreed, good catch and fix

PR #77, opened 2017-10-30
sCO2 power cycle new configurations

luiscoco:
sCO2 power cycle RC with ReHeating, pending the off-design performance


luiscoco:
pending comments

PR #78, opened 2017-10-30
lib_windfile.cpp: Rewrite trim() and locate2()

bje-:
I have rewritten these two functions to use C++ strings and vector<string> to make the code much simpler and clearer. There is no now arbitrary limitation of 128 columns in locate2().

### DIFF
--- a/shared/lib_windfile.cpp
+++ b/shared/lib_windfile.cpp
@@ -82,7 +82,7 @@ static void trim(std::string &buf)
 	  	buf.pop_back();
 }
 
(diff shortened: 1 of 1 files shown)


nickdiorio:
I merged this in 07a40f6790e2dc06328e9586800a9aca22eae121 .  I added some exception handling, as the examples I was running were failing due to invalid stoi arguments.  Essentially, it appears as if `stoi` returns that exception in the event that it cannot convert the input string to an integer.

PR #82, opened 2017-11-13
* build_linux/Makefile-sdktool, Makefile-tcsconsole: Add -Werror now.

bje-:
The -Werror option can now be added for sdktool and tcsconsole directories.


nickdiorio:
Very nice!

PR #84, opened 2017-11-19
Change auto_ptr to unique_ptr throughout.

bje-:
`auto_ptr` has been replaced by the `unique_ptr` class in C++11. These changes fix warnings of the form: `‘template<class> class std::auto_ptr’ is deprecated`.


bje-:
Any chance of getting these PRs approved? I'd like to prune some branches. :-) Thanks.

nickdiorio:
Hi Ben, yes, sorry for the delay, these should be handled this week!  One internal challenge is supporting internal systems that are on pre-C++11 compilers (Centos 6.9).  NREL is working on upgrading these systems (to Centos 7.3), but won't have it complete until March.  So, while we officially support C++11, we're trying to figure out how to support those old systems too.

nickdiorio:
FYI @sjanzou

nickdiorio:
Implemented a templated typedef called 'smart_ptr', that should detect C++11 support and use appropriate auto_ptr or unique_ptr

PR #110, opened 2017-12-30
Remove fgets

bje-:
This set of patches eliminates the use of the C fgets() function in favour of C++ streams. This eliminates lots of error-prone pointer arithmetic. This eliminates the last of the G++ warnings, so the last patch in this set enables -Werror.

Note that I renamed `locate` to `split` (as that is what it does!) and removed the arbitrary limitation on the maximum number of columns that can be handled.

### DIFF
--- a/build_linux/Makefile-shared
+++ b/build_linux/Makefile-shared
@@ -1,7 +1,7 @@
 VPATH = ../shared
 CC = gcc
 CXX = g++
-WARNINGS = -Wall -Wno-strict-aliasing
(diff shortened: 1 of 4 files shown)


nickdiorio:
Of note, this failed all tests.  Please check in the future to fix the errors.

I went through and added an ifs.clear() before all calls to ifs.seekg(0), according to: http://www.cplusplus.com/forum/beginner/30644/.

Without this, all calls to open weather files fail.  Thanks.

bje-:
Thanks. Consider my wrist slapped!

PR #111, opened 2017-12-31
Fix up nlopt build system

bje-:
Commit 6fefc274287e5530d13dd2ec478fe3c8fea93f2c by sjanzou doesn't observe the `configure` script and the Autoconfy way of doing things, which is to have "feature" tests, not platform tests. First, `config.h` and `config.h.in` are generated files and should not be modified by hand. I have cleaned this up and put any Windows-specific workarounds into the `.cpp` files as required.

I also had to give the sys subdirectory a new name as this clashes with the include search path to the system <sys/blah.h> headers.

I have submitted the changes in logical chunks to make review easier.


nlopt/config.h
```diff
@@ -55,12 +55,18 @@
 /* Define to 1 if you have the <sys/stat.h> header file. */
 #define HAVE_SYS_STAT_H 1
 
+/* Define to 1 if you have the <sys/time.h> header file. */
+#define HAVE_SYS_TIME_H 1
```
nickdiorio: Windows does not have sys/time, so this always fails.  We need to be able to maintain one file that when checked out, compiles on both Windows, Linux, and OSX.  I will add a conditional statement around this that sets correctly for windows.

nickdiorio:
@sjanzou , just FYI.  I'm unable to build in Windows.  I will try and work on more soon

bje-:
Sure, no worries. Remember that config.h is a generated file and, so, really, `configure` should be run on Windows as well to probe the system and write out a `config.h` with HAVE_SYS_TIME_H defined to 0, for example. It does complicate the builds on Windows, but having said that, running `configure` scripts on Windows today is easier than it has ever been.

nickdiorio:
Good point Ben, perhaps we should have a config_win.h, config_linux.h, config_max.h, and then selectively include depending on the platform.  Then we could run configure on each platform to correctly generate the config

bje-:
Unfortunately, that would eliminate the whole point of the configure script, which is to dynamically disocover all of this at compile time.  So, for example, I might get a different `config.h` on a 32-bit vs 64-bit Linux system. It should not be necessary to check in different static versions of the file. Is it not possible to run the `configure` script from the Visual Studio build system?

nickdiorio:
That is a possibility I hadn't considered.  We can look into how that could be accomplished.

sjanzou:
I have been going through the option of using MinGW to run ./configure in nlopt as outlined here https://stackoverflow.com/questions/4179800/the-equivalent-of-configure-in-windows Currently, this will configure the generated files for g++ in MinGW but I have not yet finished configuring for nmake and cl for Visual Studio. As >90% of our users are Windows based, I am not sure this extra configuration will be well received. We may need to come up with an alternate plan for each of the platforms. This is how the configuration runs in the MinGW shell on Windows 10: 
image

bje-:
That `configure` run looks good. One possibility is to run it on a Windows 10 system, rename the generated file to `config-windows.h` or somesuch and then conditionally include that on Windows. Check that version into git. It's not ideal, but it is a pragmatic solution that keeps the burden down on Windows users trying to build.

bje-:
BTW you should be able to use the Visual Studio compiler by running `CC=cl ./configure`.

sjanzou:
Ben, Thanks for the update. We discussed at our meeting today and we are going to make a patch to our public release the first week of February and then will revisit this issue with a workaround with a symbolic link in Linux most likely. We are open to other options but the MinGW and configure script on windows with modifications necessary for every update to nlopt using #ifdefs is currently not feasible..

jdpipe:
Another option for resolving this is to expect NLOPT to already be installed somewhere, not to attempt to automate its build within SSC. On Ubuntu/Debian, there is also a usable version of NLOPT which so far seems to work. For MinGW/MSYS2, it would be easy to give instructions for quickly compiling/building/installing NLOPT (although I note that MSYS2 already has an NLOPT package). For MSVS on Windows, perhaps supply a dedicated vcproj? https://github.com/NREL/SAM/issues/76

nickdiorio:
Sorry for the delay in addressing this, but the following solution has been implemented: https://github.com/NREL/ssc/commit/474eb27edccb751c26740d0fd6faa9c3ffd1eb8f

1.  Files in nlopt which included `config.h` now include `platform-config.h`, which has an `#ifdef` for the platform, selectively including `config-windows.h` or `config.h`.
2. On Windows, the `config-windows.h` has been written to work on Windows.  The `configure` script unfortunately does not generate a file which works by default.
3. On Linux/OSX, `Makefile-nlopt` runs `configure`, which generates `config.h`.  This file is not checked into version control since it is platform dependent.

This is the cleanest way I could think to do this while preserving the intent of the `configure` script.  Even using MINGW, running `configure` on Windows does not produce a working `config.h`.

bje-:
Sounds reasonable!

PR #126, opened 2018-02-21
Cleanup code for solarpilot app open source

mjwagner2:
removing extraneous code, cleaning up header files, removing plant and TES from solarpilot calculations to focus only on solar field


mjwagner2:
Build fails on linux

PR #148, opened 2018-03-21
VSCode project files

### LINKED ISSUE #147: VSCode project files needed
VSCode is the most popular IDE, and it would be nice to provide project files for SSC on Linux.

mjwagner2:
This creates project files in the folder ".vscode" in the main project directory. These files configure the build to use the default makefiles, and provide a launcher for debugging. Fixes #147.

I have only tested this on one Ubuntu 16.04 machine (native). It would be helpful for others to review and see where things need to be generalized.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,47 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "includePath": [
+                "/usr/include/c++/5",
+                "/usr/include/x86_64-linux-gnu/c++/5",
+                "/usr/include/c++/5/backward",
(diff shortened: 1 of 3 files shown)


nickdiorio:
Hi Mike, I'm not familiar with VSCode, but from my perspective if you've tested and it looks good to you, then it's okay with me.  One thing to keep in mind, is that adding support for more IDE's increases the amount of maintenance within the ssc repo.  We already have a hard time keeping up with the Linux and OSX builds, but it is good that it largely uses the existing Makefiles and provides IDE support for OSX and Linux.  @sjanzou , any thoughts?

nickdiorio:
@mjwagner2 , I'll approve, as I'm sure this will be a valuable addition for Linux users in particular.  Thank you for going through the effort.

PR #156, opened 2018-05-08
Weather forecast

nickdiorio:
A ton of work, including adding new dispatch options, revamped battery dispatch and charge controllers.

Creating PR to run tests on before merging


nickdiorio:
Fixed test issues and build issues, local Linux build suggests everything should be good.

PR #174, opened 2018-06-18
Bugfix: return SP_DATTYPE

jerluc:
This bugfix improves type compatibility within the solarpilot module.


nickdiorio:
@mjwagner2 , can you take a quick look?  The CI failure is not related to this commit.

nickdiorio:
Since this was approved, I think we just dropped the ball on merging it.  Merging now.

PR #177, opened 2018-06-20
Interaction limit parameter for SolarPILOT

### LINKED ISSUE #161: Parameter to specify heliostat block/shade interaction limit
Currently, the limit for calculating interaction between two heliostats is hard-coded at 100 times the height of the heliostat. If shadowing and blocking are present in heliostats beyond that distance (at very low sun angles), SolarPILOT ignores the interaction. 

In order to test the effect of changing this limit, the value should be an input parameter that the user can adjust. This should likely be done in the detailed options on the SolarPILOT GUI, but left at a default value of 100 for SSC and SAM applications.

mjwagner2:
The interaction limit defines the radius around any given heliostat that blocking or shadowing is considered from any other heliostat. Providing a limit reduces the computational requirement of evaluating shadowing and blocking, but has the potential to introduce inaccuracy in assessing these losses. 

Different heliostat layouts or geometries may require different limits to balance accuracy and computational expense. This PR introduces a parameter that replaces the hard-coded limit of 100 x (heliostat height). In SAM and SSC, the default value is assumed, and the parameter is not exposed in the SAM interface.

Closes #161.


mjwagner2:
Travis issue appears to have caused build failure. Going ahead with merge.

PR #180, opened 2018-06-20
Restructure test

dguittet:
Migrate ssc integration tests to use TestInfo/TestResult structure


dguittet:
Testing shows little performance gain in using this new structure, which takes longer to setup than the old version. Closing as unnecessary.

PR #186, opened 2018-07-16
Inverter thermal derate model

dguittet:
Added inverter thermal derate curve to SharedInverter
SharedInverter->CalculateACPower() function now has a third parameter, ambientT, used to calculate whether derating should be applied

Added unit tests for SharedInverter and an integration test for pvsamv1. 

Changed definition in pvsamv1 of output variable "annual_ac_inv_eff_loss_percent" to reflect actual inverter efficiency reduction

@nickdiorio dispatch_manual_t and battstor use CalculateACPower(). Do they need to be aware of potential powerAC reductions due to temperature? In the code, I have changed ambientT to be 0.0 to reflect (assumed) independence.


ssc/cmod_pvsamv1.cpp
```diff
@@ -3158,10 +3213,14 @@ void cm_pvsamv1::exec( ) throw (compute_module::general_error)
...
+	percent = 0.;
+	if (annual_dc_net > 0) percent = 100 * annual_inv_tdcloss / annual_dc_net;
+	assign("annual_ac_inv_tdc_loss_percent", var_data((ssc_number_t)percent));
+
 	sys_output = annual_dc_net;
-	sys_output -= (annual_inv_cliploss + annual_inv_pntloss + annual_inv_psoloss);
+	sys_output -= (annual_inv_cliploss + annual_inv_pntloss + annual_inv_psoloss + annual_inv_tdcloss);
 	percent = 0.;
-	if (sys_output > 0) percent = 100 * (sys_output - annual_ac_gross) / sys_output;
+	if (annual_dc_net > 0) percent = 100 * (annual_dc_net - sys_output) / annual_dc_net;
```
dguittet: redefine "annual_ac_inv_eff_loss_percent" 

in a test, I derated the inverter to ~50% for 50% of the year but this value only showed 8% efficiency loss, whereas with new definition it shows 20% efficiency loss

nickdiorio: I think we could modify the battery model (DC-connected) to also derate the inverter for temperature, but we can wait to do that while we focus on this initial integration

shared/lib_shared_inverter.cpp
```diff
@@ -8,9 +9,93 @@ SharedInverter::SharedInverter(int inverterType, int numberOfInverters,
...
+bool sortByVoltage(const double i[], const double j[])
+{
+	return (i[0] < j[0]);
+}
+
+bool SharedInverter::setTempDerateCurves(double* curve1, double* curve2, double* curve3)
+{
+	if (curve1[0] < 0 || curve1[1] <= -99 || curve1[2] >= 0) return false;
+
+	// if multiple curves provided, partition operating V range
```
janinefreeman: Partitioning, or just sorting them to be in order?

shared/lib_shared_inverter.cpp
```diff
@@ -8,9 +9,93 @@ SharedInverter::SharedInverter(int inverterType, int numberOfInverters,
...
 
-void SharedInverter::calculateACPower(const double powerDC_Watts, const double DCStringVoltage)
+bool sortByVoltage(const double i[], const double j[])
+{
+	return (i[0] < j[0]);
+}
+
+bool SharedInverter::setTempDerateCurves(double* curve1, double* curve2, double* curve3)
+{
+	if (curve1[0] < 0 || curve1[1] <= -99 || curve1[2] >= 0) return false;
```
janinefreeman: Where do these limits come from? Worth a comment. Also possibly worth creating variables for the limits if they're going to apply to multiple derate curves, so that if we want to change the limits, we only have to change them in one place.

shared/lib_shared_inverter.cpp
```diff
@@ -8,9 +9,93 @@ SharedInverter::SharedInverter(int inverterType, int numberOfInverters,
...
+	vParts[0] = m_tempV[0];
+	vParts[1] = m_tempV[1];
+	startC[0] = m_tempStartC[0];
+	startC[1] = m_tempStartC[1];
+	startC[2] = m_tempStartC[2];
+	slope[0] = m_tempSlope[0];
+	slope[1] = m_tempSlope[1];
+	slope[2] = m_tempSlope[2];
+}
+
```
janinefreeman: Add some comments above this function to explain what it’s doing, what units are of inputs/outputs, etc.- for example, include that it modified pAC, doesn’t just return a calculated value

shared/lib_shared_inverter.cpp
```diff
@@ -8,9 +9,93 @@ SharedInverter::SharedInverter(int inverterType, int numberOfInverters,
...
+void SharedInverter::calculateTempDerate(double V, double T, double& pAC, double& eff, double& loss)
+{
+	if (eff == 0. || pAC == 0.) return;
+
+	double slope = 0.0;
+	double deltaT = 0.0;
+
+	if (m_tempV[0] == 0 || (m_tempV[0] != 0 && V <= m_tempV[0])){
```
janinefreeman: Why checking for 0 here, and again in line 79? What if there's only one curve specified? This seems to assume all three curves exist. What if there's only one curve, but the user inputs V > m_tempV[0]? What happens then? One possible logic fix would be to work backwards from the biggest existing curve until you find one with m_tempV[i] < V, otherwise use the first curve. 

Also totally possible I'm completely misunderstanding what's happening here. :)

shared/lib_shared_inverter.cpp
```diff
@@ -8,9 +9,93 @@ SharedInverter::SharedInverter(int inverterType, int numberOfInverters,
...
+		}
+		else{
+			if (T <= m_tempStartC[2]) return;
+			slope = m_tempSlope[2];
+			deltaT = T - m_tempStartC[2];
+		}
+	}
+
+	double pDC = pAC/eff;
+	eff += deltaT*slope;
```
janinefreeman: worth clarifying here that the slope will be negative, otherwise it looks like the logic is wrong. Also worth clarifying units on slope- what are they here (presumably it would be .01 rather than 10) and when did they get set that way?

shared/lib_shared_inverter.cpp
```diff
@@ -8,9 +9,93 @@ SharedInverter::SharedInverter(int inverterType, int numberOfInverters,
...
+			slope = m_tempSlope[1];
+			deltaT = T - m_tempStartC[1];
+		}
+		else{
+			if (T <= m_tempStartC[2]) return;
+			slope = m_tempSlope[2];
+			deltaT = T - m_tempStartC[2];
+		}
+	}
+
```
janinefreeman: Additional comment here on the logic of lines 91-95 might be helpful (although redundant). It looks correct to me, but had to write down an example to verify.

shared/lib_shared_inverter.cpp
```diff
@@ -8,9 +9,93 @@ SharedInverter::SharedInverter(int inverterType, int numberOfInverters,
...
+	double slope = 0.0;
+	double deltaT = 0.0;
+
+	if (m_tempV[0] == 0 || (m_tempV[0] != 0 && V <= m_tempV[0])){
+		if (T <= m_tempStartC[0]) return;
+		slope = m_tempSlope[0];
+		deltaT = T - m_tempStartC[0];
+	}
+	else{
```
janinefreeman: Looks like it uses the derate curve whose voltage is >= the input voltage. Is that true? Worth a comment on that too. Is that what the manufacturers recommend? Or do they say that you should interpolate between curves?

ssc/cmod_pvsamv1.cpp
```diff
@@ -372,38 +372,50 @@ static var_info _cm_vtab_pvsamv1[] = {
...
+	{ SSC_INPUT,        SSC_NUMBER,      "inv_pd_pnt",                                 "AC power consumed by inverter at night",                   "Wac",     "",                     "pvsamv1",       "inverter_model=2",                    "",                              "" },
+	{ SSC_INPUT,        SSC_NUMBER,      "inv_pd_vdco",                                "DC input voltage for the rated AC power rating",           "Vdc",     "",                     "pvsamv1",       "inverter_model=2",                    "",                              "" },
+	{ SSC_INPUT,        SSC_NUMBER,      "inv_pd_vdcmax",                              "Maximum DC input operating voltage",                       "Vdc",     "",                     "pvsamv1",       "inverter_model=2",                    "",                              "" },
+
+	{ SSC_INPUT,		SSC_NUMBER,		 "en_inv_tdc",								   "Enable inverter temperature derating curve(s)",			   "0/1",     "",					  "pvsamv1"		   "*",								  "",							   "" },
```
janinefreeman: Perhaps worth investigating whether we should use an SSC_MATRIX here, so that they all show up as one table/one variable rather than 9 input variables. Will depend a little bit on how the UI input is set up, so it makes sense to wait on that decision until you add the UI piece.

ssc/cmod_pvsamv1.cpp
```diff
@@ -1524,6 +1539,51 @@ void cm_pvsamv1::exec( ) throw (compute_module::general_error)
 	// The shared inverter of the PV array and a tightly-coupled DC connected battery
 	std::unique_ptr<SharedInverter> sharedInverter(new SharedInverter(inv_type, num_inverters, &snlinv, &plinv));
 
+	// Inverter thermal derate curves
```
janinefreeman: This section might also change if we make use of an SSC_MATRIX

ssc/cmod_pvsamv1.cpp
```diff
@@ -1524,6 +1539,51 @@ void cm_pvsamv1::exec( ) throw (compute_module::general_error)
...
+	bool en_inv_tdc = as_boolean("en_inv_tdc");
+	if (en_inv_tdc) {
+		double* curve1 = new double[3];
+		curve1[0] = as_double("inv_tdc_V1");
+		curve1[1] = as_double("inv_tdc_T1");
+		curve1[2] = as_double("inv_tdc_S1");
+		
+		double* curve2 = new double[3];
+		try {
```
janinefreeman: Try/catch is inconsistent with how we;ve typically error-checked things in the past, not necessarily a bad thing but I don't know enough about those functions. Will they build across all platforms for the right C++ standard (11)? @nickdiorio  your input on this is welcome

ssc/cmod_pvsamv1.cpp
```diff
@@ -2711,11 +2772,12 @@ void cm_pvsamv1::exec( ) throw (compute_module::general_error)
 	*********************************************************************************************** */
 	idx = 0; ireport = 0; ireplast = 0; percent_baseline = percent_complete;
 	double annual_battery_loss = 0;
+	wdprov->rewind();
```
janinefreeman: Why did reading the weather file need to be added here? Seems like there must be an existing variable containing weather file temperature here...

ssc/cmod_pvsamv1.cpp
janinefreeman: This change should be reverted- this definition is basically the efficiency of converting from DC to AC, not including clipping losses, nighttime losses, or power draw losses, which is why the annual_ac_gross term needs to be in there. Since we'll want to show inverter thermal losses as their own loss stream separate from the inverter efficiency, line 3221 is correct.

Definitely not the most transparent calculation in the world, and a big area we'd like to fix with some code rewriting. :)

dguittet: reverted!

ssc/cmod_pvsamv1.cpp
dguittet: The inverter's CalculateACPower function requires wf. tdry, and the previous instance of weather file was inside the previous for loop (DC) and so is not available in this loop (AC)

ssc/cmod_pvsamv1.cpp
dguittet: try catch is a standard phrase and is used in ssc/core.cpp for instance. It was implemented in case there was no value assigned to some of these variables (a SDK user who only needed one curve so did not bother to assign the rest)

But I changed to use a matrix and that simplified this section

ssc/cmod_pvsamv1.cpp
dguittet: Done!

ssc/cmod_pvsamv1.cpp
dguittet: For the UI input, I needed a separate set of inputs for each inverter type (since it doesn't allow multiple variables with the same name, as happened when I tried to attach the collapsible page onto each of the inverter models), so I changed them to a matrices for simplicity.

shared/lib_shared_inverter.cpp
dguittet: Good point!

shared/lib_shared_inverter.cpp
dguittet: Good question. The first comes from the fact that the voltage needs to be above 0 in order for any derating to be applied. The third comes from the fact that the derating needs to be negative. The second one I changed to be <=-273 since that's the lowest possible temperature in C :)

shared/lib_shared_inverter.cpp
dguittet: It was partitioning, but now that you mention it, interpolating might be better. What do you think?

shared/lib_shared_inverter.cpp
janinefreeman: I think the partitioning method makes sense, barring any recommendations from the manufacturer (maybe we could reach out, or see if any other tools have this feature implemented and if so, how they do it). If you add some comments here explaining which curves get used when, I think that would be sufficient.

ssc/cmod_pvsamv1.cpp
janinefreeman: Ohhh. That makes sense, but is a bummer. Perhaps something we'll need to figure out/simplify in a future version.

ssc/cmod_pvsamv1.cpp
janinefreeman: What about using the tdry output variable here?

Hopefully something that will be addressed with @nickdiorio 's IOManager. :)

PR #199, opened 2018-10-15
Add output for AC capacity factor

nickdiorio:
Adds an output (in data tables, not metrics table) for a capacity factor based upon the nameplate AC capacity of the system.  Calculation is: `capacity_factor_ac = annual_energy_ac / nameplate_ac`.

This references: https://github.com/NREL/SAM/issues/103.

To demonstrate, consider residential PV system with:
image

SAM calculates:
image


janinefreeman:
This looks good to me. Thanks for adding this @nickdiorio !

PR #200, opened 2018-10-15
Fixes NREL/SAM#25

nickdiorio:
This pull request attempts to make it less likely that a programmer will inadvertently try to use or assign to un-allocated outputs by forcing the allocation process to occur for all structures at the end of the construction of PVIOManager.

This pull request also creates an intermediate variable to store the lifetime dc degradation factor, which was previously being allocated at an output but calculated directly in the constructor of the PVSystem.  This was problematic once all output allocation was moved to to the end.  Typically outputs should be calculated with the compute module itself (pvsamv1), not in the IOManager, which is responsible for setting up inputs to the model and allocated the memory required for storing outputs.


dguittet:
Nick, this looks fine to me!

PR #201, opened 2018-10-16
flag class for pv_io_manager using safe bool idiom

dguittet:
Added flag class to lib_pv_io_manager.h as a wrapper for a bool/int value that is used for equality comparisons in pv_io_manager. The class implements an initialization check before each comparison is evaluated. Common comparison usages such as (flag), (flag == x) and (flag != x) overloaded with initCheck.

Mostly follows https://www.artima.com/cppsource/safeboolP.html. Disallowing comparisons to other types using templated overload was not used in order to allow comparisons given implicit conversion of enum values to int (otherwise a comparison with an enum could use both the int and templated version, resulting in error).

Since the flag class is intended for *_IO_Managers, I left it in lib_pv_io_manager.h for now. Perhaps if we had several *_IO_Managers in the future, then we could separate it.


nickdiorio:
@dguittet , I think this looks great, very sophisticated!

PR #226, opened 2019-01-16
Warning fixes

nickdiorio:
Team please review changes here, both Windows and Linux platforms build (sorry OSX), and unit tests run.

Recommendations
- Change remaining Makefiles to use more generic way to generate object files.  Shouldn't have to add new files to Makefile, just new directories.
- Fix remaining warnings in SSC and move to treat warnings as errors.


nickdiorio:
Fixed all windows warnings with Visual Studio 2017, updating to latest patch to ensure nothing was broken

Issue #228, opened 2019-01-20
Standardize/update module efficiency calculations across all module models


caseyzak24:
Currently, the definition of module efficiency (passed out of the module model as `out.Efficiency` and accessed as `subarrayX_modeff`) when using the mlm module model is inconsistent with the definition of all the other available module models.

In the mlm model, efficiency = output_power / (collector_area\***poa_eff**), where poa_eff is the total irradiance after module cover effects (in keeping with the output definitions in the latest release). In all other module models, efficiency = output_power / (collector_area\***poa_shaded_soiled**), such that the given module efficiency includes module cover effects.

A simple solution would be to change the implementation of the mlm model, but I think this brings up the larger question of how module efficiency *should* be defined in ssc/SAM. When I first used ssc I was surprised to find that efficiency was defined relative to poa_shaded_soiled, since other software (namely PVsyst) defines it relative to poa_eff. Also, defining it relative to poa_eff aligns with the way the name plate rating of a module is defined (it also does not consider angle of incidence effects).

I am sure there are good reasons to define it relative to poa_shaded_soiled, but I hope this request will eventually result in a standardization of the calculation.

nickdiorio:
Thanks @caseyzak24 , you bring up some excellent points.  We do have the plan to re-evaluate these calculations this year, in addition to breaking more losses out of module model.  We'll definitely keep your recommendations in mind, and I agree that it would make sense if we align an efficiency definition across models.

janinefreeman:
Great explanation @caseyzak24 . Assuming that it will take a while for various tools to agree on a definition of efficiency, I can see advantages to reporting either efficiency in the meantime. On one hand, being consistent with the definition reported by the other module models in SAM is helpful, but on the other hand, consistency with output values reported by pvsyst is arguably the primary reason this model option is used in SAM. I'd be curious to hear @caseyzak24 , @timorichert, and @cpaulgilman 's opinions on this question prior to merging in @caseyzak24 's fix for that particular line.

Thanks for starting an interesting discussion! :)

caseyzak24:
@janinefreeman you bring up an interesting point; I hadn't thought about people expecting the output of the model to line up with PVsyst outputs...

Long term, I'd like to see the definition of efficiency across all module models changed to align with PVsyst, simply because it is the most clean definition of efficiency and it aligns with the calculation of nameplate capacity for modules. Or, perhaps, there should be 2 reported efficiencies, something like "cell conversion efficiency" and "module assembly efficiency".

[...]

Issue #227, opened 2019-01-20
mlm model doesn't output out.AOImodifier


caseyzak24:
The mlm module model in lib_mlmodel.cpp does not output `out.AOIModifier`. As a result, when using the mlm module model in conjuction with `pvsamv1`, the output signals `subarray1_poa_front` and `subarray1_poa_eff` are zero.

nickdiorio:
We can investigate adding this, though it likely won't be for a few months.  If you'd like to submit a pull request, please feel free.

nickdiorio:
Closed with #230

PR #230, opened 2019-02-04
fix(): Address bugs and increase usability of ml module model

### LINKED ISSUE #229: mlm module model support for irradmode=3 (poa from reference cell)
The mlm module model currently only considers poa components (i.e. `input.Ibeam`, `input.Idiff`, and `input.Ignd`) but will not use directly supplied user POA reference cell data.

It would be nice if the same logic found in `iec61853_module_t` were included at the beginning of `mlm_module_t`, something like:
```
if( input.radmode != 3 ){ // Skip module cover effects if using POA reference cell data 
		double S = (f_IAM_beam * input.Ibeam + f_IAM_diff * input.Idiff + groundRelfectionFraction * f_IAM_gnd * input.Ignd) * f_AM;
        }
else if(input.usePOAFromWF){ // Check if decomposed POA is required, if not use weather file POA directly
		double S = input.poaIrr;
	} 
else { // Otherwise use decomposed POA
		double S = (f_IAM_beam * input.Ibeam + f_IAM_diff * input.Idiff + groundRelfectionFraction * f_IAM_gnd * input.Ignd) * f_AM;
	}
```

caseyzak24:
Fixes #229, #227 and provides a short-term fix for #228. For now, changed ml model so that it calculates efficiency like the other models, i.e. 

eff = P / (module_area * POA_shaded_soiled)

However, the larger question of how module efficiency _should_ be calculated consistently remains open.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -337,6 +337,8 @@ build_linux/config.status
 linux_test\.txt
 
 nlopt/config\.h
+nlopt/config.h
+build_osx/config.status
 
 doc/html/*
 doc/latex/*
--- a/shared/lib_mlmodel.cpp
+++ b/shared/lib_mlmodel.cpp
@@ -214,7 +214,16 @@ bool mlmodel_module_t::operator() (pvinput_t &input, double T_C, double opvoltag
 	}
 
 	// Total effective irradiance
-	double S = (f_IAM_beam * input.Ibeam + f_IAM_diff * input.Idiff + groundRelfectionFraction * f_IAM_gnd * input.Ignd) * f_AM;
+	double S;
+	if(input.radmode != 3){ // Skip module cover effects if using POA reference cell data
+		S = (f_IAM_beam * input.Ibeam + f_IAM_diff * input.Idiff + groundRelfectionFraction * f_IAM_gnd * input.Ignd) * f_AM;
+    }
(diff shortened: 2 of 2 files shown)


nickdiorio:
Thanks @caseyzak24 for submitting this, that's great.  We'll take a look and let you know if we see anything that looks off.

caseyzak24:
@nickdiorio I rebased this branch off of a more recent version of `develop` and then made a small change to the .gitignore. I'm not sure why this is causing the build to fail, but it the logs indicate it's related to the sdktool, which I didn't edit. I will try and dig around on my end and figure out why it's failing but if you could take a look as well that would be great.

caseyzak24:
@nickdiorio I figured out what the issue was, my last commit deleted the sdktool folder for some unknown reason. The simplest solution was to roll back one commit and force push (the more recent commit was just to fix the repo index anyways). I'm not sure how to rerun the build but if you do it should pass.

nickdiorio:
@caseyzak24 , no problem, my guess is that it's being rebuilt now, as it takes about 45 minutes on the Travis CI.  I think it will probably pass.  In any case, we'll try to get this integrated in the next week.

nickdiorio:
Built and test locally, no issues.

PR #239, opened 2019-02-20
Generic battery

nickdiorio:
Removing battery from generic system and chaining within the startup script instead.


sjanzou:
Committed and closed with ssc patch commit with manual merges 7cb4deae595c01ae40ed2c01646a5432c49ef9cb

nickdiorio:
Did we not merge this?  It looks like you rewrote two files, but the rest of the changes were not included.

PR #244, opened 2019-03-23
Updates to fix sunrise/sunset bug at very high latitudes

janinefreeman:
Updated previous fix to be a more robust fix for sunrise and sunset calculations at high latitudes. This includes calculating sunrise and sunset correctly and allowing power to be calculated when the sunset occurs on the next day (e.g. sunset is at 1:30AM from the previous day, sunrise is at 2:30AM that day). Added a test for the days where the high latitude locations proved problematic. In this update, also changed the nighttime sun angles that are reported so that they represent real values calculated at the actual time in question, and updated the NightCase irradproc tests accordingly (previously, the angles were either set to -999 or leftover values from calculating at noon that day).

I would appreciate a review on these changes and some testing by @cpaulgilman to confirm that they make sense.

### DIFF
--- a/shared/lib_irradproc.cpp
+++ b/shared/lib_irradproc.cpp
@@ -1097,7 +1097,7 @@ int irrad::calc()
 	double t_sunrise = sunAnglesRadians[4];
 	double t_sunset = sunAnglesRadians[5];
 
-	if (t_sunset > 24) //sunset is legitimately the next day, so recalculate sunset from the previous day
+	if (t_sunset > 24.0 && t_sunset != 100.0) //sunset is legitimately the next day but we're not in endless days, so recalculate sunset from the previous day
 	{
 		double sunanglestemp[9];
 		if (day > 1) //simply decrement day during month
@@ -1106,12 +1106,15 @@ int irrad::calc()
 			solarpos(year, month - 1, __nday[month - 2], 12, 0.0, latitudeDegrees, longitudeDegrees, timezone, sunanglestemp); //month is 1-indexed and __nday is 0 indexed
 		else //on the first day of the year, need to switch to Dec 31 of last year
 			solarpos(year - 1, 12, 31, 12, 0.0, latitudeDegrees, longitudeDegrees, timezone, sunanglestemp);
+		//on the last day of endless days, sunset is returned as 100 (hour angle too large for calculation), so use today's sunset time as a proxy
+		if (sunanglestemp[5] == 100.0)
+			t_sunset -= 24.0;		
 		//if sunset from yesterday WASN'T today, then it's ok to leave sunset > 24, which will cause the sun to rise today and not set today
-		if (sunanglestemp[5] >= 24)
+		else if (sunanglestemp[5] >= 24.0)
 			t_sunset = sunanglestemp[5] - 24.0;
 	}
 
-	if (t_sunrise < 0) //sunrise is legitimately the previous day, so recalculate for next day
+	if (t_sunrise < 0.0 && t_sunrise != -100.0) //sunrise is legitimately the previous day but we're not in endless days, so recalculate for next day
 	{
 		double sunanglestemp[9];
 		if (day < __nday[month - 1]) //simply increment the day during the month, month is 1-indexed and __nday is 0-indexed
@@ -1120,8 +1123,11 @@ int irrad::calc()
 			solarpos(year, month + 1, 1, 12, 0.0, latitudeDegrees, longitudeDegrees, timezone, sunanglestemp);
 		else //on the last day of the year, need to switch to Jan 1 of the next year
 			solarpos(year + 1, 1, 1, 12, 0.0, latitudeDegrees, longitudeDegrees, timezone, sunanglestemp);
+		//on the last day of endless days, sunrise would be returned as -100 (hour angle too large for calculations), so use today's sunrise time as a proxy
+		if (sunanglestemp[4] == -100.0)
+			t_sunrise += 24.0;		
 		//if sunrise from tomorrow isn't today, then it's ok to leave sunrise < 0, which will cause the sun to set at the right time and not rise until tomorrow
-		if (sunanglestemp[4] < 0)
+		else if (sunanglestemp[4] < 0.0)
 			t_sunrise = sunanglestemp[4] + 24.0;
 	}
 
--- a/shared/lib_irradproc.cpp
+++ b/shared/lib_irradproc.cpp
@@ -219,30 +219,46 @@ void solarpos(int year,int month,int day,int hour,double minute,double lat,doubl
 		E = E - 24.0;
 
 	arg = -tan(lat)*tan(dec);
-	if (arg >= 1.0)  /* No sunrise, continuous nights */
+	if (arg >= 1.0)  // No sunrise, continuous nights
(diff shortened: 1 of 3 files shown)


shared/lib_irradproc.cpp
```diff
@@ -219,30 +219,46 @@ void solarpos(int year,int month,int day,int hour,double minute,double lat,doubl
...
-		if (sunset > 24) sunset -= 24;
-		else if (sunset < 0) sunset += 24;
+		//now a bunch of error checks to try to correctly catch weird behavior
+		//both sunrise and sunset may be shifted by 24 hours (example: Fiji- positive tz negative lng), if so, roll them both back
+		if (sunrise > 24.0 && sunset > 24.0)
+		{
+			sunrise -= 24.0;
+			sunset -= 24.0;
+		}
+		/*//if only sunset is greater than 24, it actually sets the next day, but ws not quite big enough to be caught (example: Point Hope, AK)
```
dguittet: It seems like there are 4 possible odd cases:
sunrise > 24 and sunset > 24
sunrise < 0 and sunset < 0
sunrise > 24 and sunset < 0
sunrise < 0 and sunset > 24
The last two seem to be handled in lines 222-232 whereas the first two inside 234-262, but is it possible that |arg| < 1.0 but we still end up in the last two cases?

shared/lib_irradproc.cpp
```diff
@@ -1090,12 +1106,25 @@ int irrad::calc()
 	double t_sunrise = sunAnglesRadians[4];
 	double t_sunset = sunAnglesRadians[5];
 
+	if (t_sunset > 24) //sunset is legitimately the next day, so recalculate sunset from the previous day
+	{
+		double sunanglestemp[9];
+		solarpos(year, month, day - 1, 12, 0.0, latitudeDegrees, longitudeDegrees, timezone, sunanglestemp);
```
dguittet: What happens if the timestep is 3/1/2018 and (day -1 = 0)?

shared/lib_irradproc.cpp
```diff
@@ -1090,12 +1106,25 @@ int irrad::calc()
...
+		t_sunset = sunanglestemp[5] - 24.0;
+	}
+
+	if (t_sunrise < 0) //sunrise is legitimately the previous day, so recalculate for next day
+	{
+		double sunanglestemp[9];
+		solarpos(year, month, day + 1, 12, 0.0, latitudeDegrees, longitudeDegrees, timezone, sunanglestemp);
+		t_sunset = sunanglestemp[5] + 24.0;
```
dguittet: Before adding or subtracting 24, I would check that it is indeed greater than or less than 24, unless we're at that unlucky day where yesterday sunset was at 11:59 pm and today sunset is tomorrow at 12:01 am.

dguittet:
I tried to follow the logic of the algorithm and think of edge cases where it might not work. I don't have much experience in solar position algorithms, so some of my comments may be off, but overall I think there are a few odd cases left that should have error handling.

Are there weather files in these locations with multiple all-days or all-nights that we can test by running irrad_proc over the entire year?

shared/lib_irradproc.cpp
```diff
@@ -1134,11 +1162,9 @@ int irrad::calc()
 	else
 	{	
 		// sun is down, assign sundown values
-		sunAnglesRadians[0] = -999*DTOR; //avoid returning a junk azimuth angle (return in radians)
-		sunAnglesRadians[1] = -999*DTOR; //avoid returning a junk zenith angle (return in radians)
-		sunAnglesRadians[2] = -999*DTOR; //avoid returning a junk elevation angle (return in radians)
-		timeStepSunPosition[0] = 0;
-		timeStepSunPosition[1] = 0;
+		solarpos(year, month, day, hour, minute, latitudeDegrees, longitudeDegrees, timezone, sunAnglesRadians);
```
dguittet: Is this calculation necessary since timeStepSunPosition[2] = 0; and we will go straight to line 1237?

shared/lib_irradproc.cpp
```diff
@@ -1134,11 +1162,9 @@ int irrad::calc()
...
+		timeStepSunPosition[0] = hour;
+		timeStepSunPosition[1] = (int)minute;
```
dguittet: If we are in continuous days, then starting at 1109, t_sunrise = -1 and t_sunset = 25, so then we will get t_sunset = 25-24 = 1, and in line 1120, t_sunset = 25. Then for the entire day, we are in line 1157 sunup case. So this looks right. 

In continuous nights, we have t_sunrise = 25 and t_sunset = -1, then we will go to line 1165 so that makes sense too.

But when we are in the last night of continous days, say sunset is hr 20 but will sunrise then be -1 from solarpos?

If so then, when we get from line 1120, t_sunset = 20+24, effectively pushing it back day after day until sunrise is no longer calculated as -1.

shared/lib_irradproc.cpp
janinefreeman: The logic here is that in case 1 or case 2, the equations just aren't working for a weird longitude/time zone combination (which is what we've found with test cases historically), so they legitimately both just need to be rolled by 24 hours. I don't think Case 3 is possible without it being caught by the |arg| < 1.0 check (or something going VERY wrong with the inputs), because sunrise= 12-x and sunset=12+x. Case 4 I would say is two separate cases: sunrise < 0 OR sunset > 24. Because cases 1 and 2 have hopefully ruled out cases of the equations not working for a weird tz/long combo, then likely Case 4A and 4B are legitimate sunrises/sunsets that occur on a different day in high latitude locations. Since they are legit, we want solarpos to return these times as they are, and have subsequent logic able to deal with them (lines 1140-1156).

shared/lib_irradproc.cpp
janinefreeman: Ah, good point! I will update this logic. Thanks!

shared/lib_irradproc.cpp
janinefreeman: Good point on both of these pieces, updating the code accordingly.

shared/lib_irradproc.cpp
janinefreeman: Yes, this calculation is necessary because timeStepSunPosition gets reported as a result, so it must be assigned using the right hour or they will report the results from noon (which is why previously they were overwritten with -999 and 0).

shared/lib_irradproc.cpp
janinefreeman: Continuous nights won't trigger lines 1109 and 1116 because t_sunrise = 25 !< 0 (line 1109) and t_sunset = -1 !> 24 (line 1116). This will correctly trigger nighttime per line 1155 because t_cur will never be < -1 or > 25. 

Just to be safe, I'm also updating 225-232 to set sunrise/sunset to +- 100 on continuous nights/continuous days, so that even if they get rolled by 24 hrs, they'll still be out of the range of 0-24. 

In adding the check you suggested to check that sunrise/sunset from the next/previous day are also actually <0 or >24, this will fix the first/last day of continuous days problem. Good catch!

shared/lib_irradproc.cpp
dguittet: Ok, thanks for clarifying!

PR #246, opened 2019-04-03
Update sscapi.cpp

yassineennassiri:
I've just added, cm_alkaline_electrolyzer to line 160 and 247 in order to make the compute module that I'm developing accessible.


janinefreeman:
Hi @yassineennassiri ! Looks like you're making progress on your branch! Usually we would not integrate a pull request until the feature is fully developed, although this may have been an unintentional pull request. I'll close it for now, and we look forward to seeing your new model as it gets closer to completion. As a side note, as you go along you'll want to make sure that your code is built on the "develop" branch of SSC, since that is the branch to which we add new features. Thanks!

PR #255, opened 2019-04-30
PySam

dguittet:
Adds CMakeLists.txt which produces each target insides its own folder. Can be run as its own project or as part of a larger (system_advisor_model) project. Names are the same across platforms.

SSC var table changes.

Cleaned up some code also.

Building in Visual Studios 2017 on Windows and CLion on MacOSX is working.

Currently makes no changes to build_vs2017 slns, so each build process is independent.

After merging, some evaluation of using CMake is required to decide if we'd like to switch or maintain both.


dguittet:
Refreshing the PR to sort out the recent merge conflicts that have popped up since the start of this PR.

PR #256, opened 2019-05-03
Interconnection

nickdiorio:
This pull request adds:

- cmod_grid.cpp, a model to take incoming total site generation and enforcing a grid interconnection limit. 
- Tests for the interconnection model


janinefreeman:
@nickdiorio Looks good to me. I see that you've implemented this grid compute module and the fuel cell one a little bit differently than we historically have (header file with constructor), maybe you can walk me through that decision at some point.

nickdiorio:
@janinefreeman , the reason for implemented the compute modules this way is primarily:

- To encapsulate IO management within another structure accessible by the compute module.  Similar to the IO manager, this should make the main compute module code more readable and shorter.


Thanks for the review!

Issue #236, opened 2019-02-09
Apply diffuse self-shading in all racking/shade-behavior scenarios


caseyzak24:
Currently diffuse self-shading in SAM is applied as shown in the following graph

![image](https://user-images.githubusercontent.com/29411281/52514304-418fb700-2bc5-11e9-8eac-002a9df2aec2.png)

where the green scenarios have it applied and the blue scenarios do not.

However, the current diffuse self-shading model, which is based on a simple 2-D unlimited sheds geometry, seems applicable as an approximation to all scenarios. In particular for backtracking SAT systems, diffuse shading becomes important due to the higher tilt angles experienced and we've seen it contribute up to a 2% loss to energy production.

janinefreeman:
Thanks for creating the issue with all the helpful details for this, @caseyzak24 ! The SAM team agrees that diffuse self-shading should be applied in the blue scenarios as well. If you get a chance to make some improvements (or if anyone else is interested in tackling it!), let us know!

caseyzak24:
@janinefreeman I am currently working on this issue and I think I have it implemented, just working out a few kinks. I will submit a PR when it is complete.

janinefreeman:
Awesome! Looking forward to it! Thanks @caseyzak24 !

janinefreeman:
Closed in 27a1793664c34b64b06d06ba055579353f4b55fc

PR #280, opened 2019-06-21
feat(): Apply diffuse self-shading in all racking/shade-behavior scenarios and improve self-shading algorithm

caseyzak24:
There are 3 major updates in this PR:
- Resolve Issue #236 by applying the currently existing diffuse self-shading model in all scenarios
- Improve the diffuse self-shading algorithm so that the results more closely match PVsyst
- Correct miscalculation of self-shading albedo derate

After making the necessary changes so that diffuse self-shading could be applied to SAT systems, I ran a representative simulation and compared the self-shading diffuse derate to that calculated for the same scenario in PVsyst:

_5MW Backtracking SAT system in Oregon with 355 Mono modules (ML model) and a 30% GCR_
Below are the results:
sam_diffuse_shading_vs_PVsyst

The results show poor agreement that varies along with the magnitude of the derate.

In order to improve the agreement with the PVsyst model, I implemented the [PlantPredict diffuse self-shading model](https://plantpredict.com/algorithm/optical-losses/#near-diffuse-shading-factor). After doing so, for the same representative scenario, the agreement had significantly improved:
sam_vs_PVsyst_with_PPshading

Now the variability that trended with the derate magnitude is significantly reduced and the ssc derate is now consistently slightly higher than the PVsyst derate.

To confirm that this model still results in acceptable behavior for FT systems and across GCR and location, we ran the model for a number of scenarios. Below are plots of monthly relative error between the ssc model and PVsyst. 
Mono_agreement
Note that the linear shading (e.g. thin film) scenario is not represented here, but the diffuse shading model should be the same.

Finally, I also noticed that the ground shading derate had very poor agreement with PVsyst. After review, I noticed that the math described in the paper referenced by the technical documentation did not match the technical documentation or the code and fixed that as well.

### The Rub ###
I have made the fixes explicitly to improve agreement of the ssc model with PVsyst, because for our purposes (a Utility-scale developer) that is how the software brings us the most value, but I understand that PVsyst shoudl not necessarily be considered truly representative of real system performance. In this case, I do think that there are probably some incorrect trig calculations being used in the current ssc model. Regardless, I am happy to retract this PR and submit one that only addresses Issue #236 if that is what the team would prefer.

### This PR does not pass all tests ###
As far as I can tell, the only tests it is failing are ones with result-based tests, which should fail now that the model results in reduced DC production. As discussed with @janinefreeman I don't have the expertise to efficiently fix these tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -339,7 +339,6 @@ nlopt/config.status
 build_osx/config.status
 build_sdk/sdk-release/linux64/dependencies\.txt
 build_sdk/sdk-release/linux64/filetypes\.txt
-
 doc/html/*
 doc/latex/*
 build_linux/linux_errors
--- a/build_osx/config.status
+++ b/build_osx/config.status
@@ -427,7 +427,7 @@ Copyright (C) 2012 Free Software Foundation, Inc.
 This config.status script is free software; the Free Software Foundation
 gives unlimited permission to copy, distribute and modify it."
 
-ac_pwd='/Users/imacuser/Public/Projects/GitHub/NREL/ssc/build_osx'
+ac_pwd='/Users/caseyzak/repos/sam_dev/ssc/build_osx'
 srcdir='../nlopt'
 test -n "$AWK" || AWK=awk
 # The default lists apply if the user does not specify any file.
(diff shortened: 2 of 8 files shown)


nickdiorio:
Thanks @caseyzak24 , that looks really useful.  We will review and try to get merged asap.

PR #282, opened 2019-06-25
New unit tests for power tower molten salt model

kieranfung:
@Matthew-Boyd , can you take a look at this for me?


test/ssc_test/cmod_tcsmolten_salt_test.cpp
```diff
@@ -71,19 +71,687 @@ TEST_F(CMTcsMoltenSalt, DefaultSingleOwnerFinancialModel) {
...
+		ssc_data_get_number(data, "h_tower", &h_tower);
+		EXPECT_NEAR(h_tower, 193.458, 193.458 * m_error_tolerance_lo) << "Tower Height";  // choose either m_error_tolerance_lo or m_error_tolerance_hi
+
+		//ssc_number_t VARIABLE;
+		//ssc_data_get_number(data, "VARIABLE", &VARIABLE);
+		//EXPECT_NEAR(VARIABLE, EXP_VAL, EXP_VAL * m_error_tolerance_lo) << "DESCRIPTION";  // choose either m_error_tolerance_lo or m_error_tolerance_hi
+	}
+}
+
+/// Test tcsmolten_salt with alternative condenser type: Radiative
```
Matthew-Boyd: Alternative flow pattern?

test/ssc_test/cmod_tcsmolten_salt_test.cpp
```diff
@@ -71,19 +71,687 @@ TEST_F(CMTcsMoltenSalt, DefaultSingleOwnerFinancialModel) {
...
-//    { "conversion_factor",                      NR,                 89.4472,                0.1 },  // Gross to Net Conversion Factor
-//    { "N_hel",                                  NR,                 8790,                   0.1 },  // Number of heliostats
-//    { "rec_height",                             NR,                 21.6029,                0.1 },  // Receiver height
-//    { "A_sf",                                   NR,                 1.26905e6,              0.1 },  // Solar Field Area
-//    { "D_rec",                                  NR,                 17.65,                  0.1 },  // The overall outer diameter of the receiver
-//    { "annual_total_water_use",                 NR,                 97591.1,                0.1 },  // Total Annual Water Usage: cycle + mirror washing
-//    { "csp.pt.cost.total_land_area",            NR,                 1892.04,                0.1 },  // Total land area
-//    { "h_tower",                                NR,                 193.458,                0.1 },  // Tower height
-//};
+/// Test tcsmolten_salt with all defaults and the single owner financial model
```
Matthew-Boyd: Sliding pressure?

test/ssc_test/cmod_tcsmolten_salt_test.cpp
```diff
@@ -71,19 +71,687 @@ TEST_F(CMTcsMoltenSalt, DefaultSingleOwnerFinancialModel) {
...
+		//EXPECT_NEAR(VARIABLE, EXP_VAL, EXP_VAL * m_error_tolerance_lo) << "DESCRIPTION";  // choose either m_error_tolerance_lo or m_error_tolerance_hi
+	}
+}
+
+/// Test tcsmolten_salt with alternative Location: Tucson, Arizona
+/// Rest default configurations with respect to the single owner financial model
+TEST_F(CMTcsMoltenSalt, Rankine_Location_SingleOwner) {
+
+	ssc_data_t data = ssc_data_create();
+	int test_errors = tcsmolten_salt_daggett_Tucson_AZ(data);
```
Matthew-Boyd: daggett?

kieranfung:
That's my mistake. You're correct, this is a comment that was copied from
the test above (radiative cooling) intended for alternative flow pattern.
Good eye.

test/tcs_test/tcsmolten_salt_cases.h
```diff
@@ -22,4 +25,287 @@ int tcsmolten_salt_daggett(ssc_data_t &data)
...
+{
+	tcsmolten_salt_default(data);
+
+	ssc_data_set_number(data, "tech_type", 3);
+	
+	int status = run_module(data, "tcsmolten_salt");
+
+	//single_owner_default(data);
+	//status += run_module(data, "singleowner");
+
```
Matthew-Boyd: You don't need to replicate these two commented lines. Just keep in the first test and remove all the rest.

test/tcs_test/tcsmolten_salt_common_data.h
```diff
@@ -29,7 +29,7 @@ void tcsmolten_salt_default(ssc_data_t &data)
     //  The actual weather file used by SAM could change and thus change the UI output values (different input (i.e., weather file) -> different outputs)
 	int n1 = sprintf(solar_resource_path, "%s/test/input_cases/moltensalt_data/daggett_ca_34.865371_-116.783023_psmv3_60_tmy.csv", std::getenv("SSCDIR"));
 
-    ssc_data_set_string(data, "solar_resource_file", solar_resource_path);
+	ssc_data_set_string(data, "solar_resource_file", "C:/Users/kfung/Documents/SAM/sam_dev/SAM/deploy/solar_resource/daggett_ca_34.865371_-116.783023_psmv3_60_tmy.csv");
```
Matthew-Boyd: Reference solar_resource_path like before, and update that variable if necessary, noting that the path should stay within the ssc repository.

test/tcs_test/tcsmolten_salt_cases.h
```diff
@@ -22,4 +25,287 @@ int tcsmolten_salt_daggett(ssc_data_t &data)
...
+// Rest default configurations 
+int tcsmolten_salt_daggett_radiative_condenser(ssc_data_t &data)
+{
+	tcsmolten_salt_default(data);
+	// Start of radiative-cooling metrics
+	ssc_data_set_number(data, "h_ctes_tank_min", 1);
+	ssc_data_set_number(data, "ctes_tshours", 15);
+	ssc_data_set_number(data, "ctes_field_fl", 4);
+	ssc_data_set_number(data, "h_ctes_tank", 30);
+	ssc_data_set_number(data, "u_ctes_tank", 0.40000000000000002);
```
Matthew-Boyd: Identify when the generated code value is trying to approximate a less significant [digit] value. For this instance change to 0.4. Update all other similar instances and remember to ensure that the test(s) still pass.

test/tcs_test/tcsmolten_salt_common_data.h
kieranfung: Matt, I believe I also made this path error within tcsmolten_salt_cases.h for <int tcsmolten_salt_Tucson_AZ(ssc_data_t &data)>.

Can you verify?

Matthew-Boyd: What you did here is change where to find the weather file, from the path given by solar_resource_path to "C:/User/kfung/...", where solar_resource_path is "...test/input_cases/moltensalt_data..." . You made this change when it was found that the weather file had intrinsically changed and was causing some of the test values to change. Instead of changing which weather file to use, you instead just copied the weather file actually used in the UI (from the sam folder/repository) and overwrote the one pointed to here. You didn't need to make any changes to the code here, because you changed the actual weather file instead. So, you need to undo your code change in this instance, making the green highlighted line look like the red highlighted line.

Any other code changes you think you may have made will be highlighted in red or green like the above.

kieranfung: Matt, this change from green to red triggers the original issue we had in the beginning of this project, of an incorrect output value being produced via the weather file. When the red line of code is implemented, the test fails. Back to our original example, when implemented annual_energy is evaluated at 5.77e8 instead of matching the SAM UI output value of 5.81e8.

PR #296, opened 2019-07-19
Increase char array storage from [256] to [512] 

kieranfung:
@Matthew-Boyd , would you mind checking this out when you get the chance?


sjanzou:
Please fix Linux build error on Travis: 
../tcs/csp_solver_core.cpp:4843:55: error: call of overloaded ‘isnan(double&)’ is ambiguous
         if (W_dot_tes_pump < 0 || isnan(W_dot_tes_pump)){

PR #312, opened 2019-10-07
Batt fixes & unit tests

dguittet:
Changes in batt_fixes

1. Capacity models

     - Update the state of charge after applying lifetime or thermal losses, computing relative to min(qmax_lifetime, qmax_thermal)

     - Thermal losses were not being added to capacity loss

      - KiBAM parameter init changes both qmax and qmax_thermal

      - Replacement should reset both qmax and qmax_thermal

      - Replacement by percent adds new capacity by percent replaced

2. Lifetime 

      - Lifetime_t doesn't require capacity_t, only SOC & chargeChanged

3. Battery thermal

     - Battery resistance never changes, fix as constant

     - Analytic heat transfer solution to replace numerical solution, see below

4. Lib_util

     - yearOneIndex simpler implementation

5. Battery losses

     - Capacity model shouldn't be required since only the charge mode is needed

     - Capacity update from lifetime model should occur outside losses class

-------------

Analytic thermal model 

The model is the same as in the NREL report, but the solution for the temperature of the battery given an initial temp, a room temperature and a current as source term has been implemented analytically. The numerical solution was using too large of a time step, resulting in over and under estimating changes across dt_hour = 3600. An example bug effect was a change in current could result in battery being cooled to below room temperature.

The new function solves the first order lumped system analysis equation whenever the initial conditions change, ie room temp changes. After enough time (t > t_threshold) with a given room temp, the diffusion term becomes negligible and only the source term (resistance dissipating heat) affects the temperature.

The thermal model is now more sensitive to current and the model parameters such as h, heat coefficient [Wm2K], area and resistance. Incorrectly defining the thermal model can result in extreme heating leading to capacity falling to 0.

--------------

Unit tests for battery_t, thermal_t, lifetime_t, lifetime_cycle_t, lifetime_calendar_t, losses_t, capacity_t, voltage_t

--------------

@cpaulgilman The 'batt_replacement_schedule' input is used differently in the cmod than described in help.

`batt_replacement_schedule`

    - described as "Battery bank replacements per year (user specified)" [number/yr]

    - In `battstor::check_replacement_schedule()`, replacements are done at n-equal intervals per year and the amount 

        - `batt_replacement_schedule_percent`

            - Not used in SAM

            - Described as ""Percentage of battery capacity to replace in year" [%]

            - Works as % of Nominal

   - but in SAM help, it's "type a '1' for each year in which the batteries will be replaced" so [bool/yr]


shared/lib_battery.cpp
```diff
@@ -419,17 +450,24 @@ double capacity_kibam_t::q20(){return _q20;}
...
 
 void capacity_lithium_ion_t::replace_battery(double replacement_percent)
 {
+    replacement_percent = fmax(0, replacement_percent);
+    double qmax_old = _qmax;
 	_qmax += _qmax0 * replacement_percent * 0.01;
 	_qmax = fmin(_qmax0, _qmax);
-	_q0 = _qmax * _SOC_init * 0.01;
 	_qmax_thermal = _qmax;
+	_q0 += (_qmax-qmax_old) * _SOC_init * 0.01;
```
janinefreeman: confused by this- why are you subtracting the previous qmax here? I see it in a couple places.

shared/lib_battery.cpp
```diff
@@ -1260,7 +1302,7 @@ void lifetime_calendar_t::runTableModel()
 	double capacity_lo = 100;
 	double capacity_hi = 0;
 
-	// interpolation mode
+	// interpolation mode f
```
janinefreeman: is the f supposed to be there?

shared/lib_battery.cpp
```diff
@@ -1301,33 +1343,63 @@ void lifetime_calendar_t::replaceBattery(double replacement_percent)
...
-	double Cp,  double h, std::vector<double> T_room, 
-	const util::matrix_t<double> &c_vs_t ) : _dt_hour(dt_hour), _mass(mass), _length(length), _width(width), _height(height),
-	_Cp(Cp), _h(h), _T_room(T_room), _cap_vs_temp(c_vs_t)
+thermal_t::thermal_t(double dt_hour, double mass, double length, double width, double height, double R, double Cp,
+                     double h, std::vector<double> T_room_K, const util::matrix_t<double> &c_vs_t)
+        : dt_sec(dt_hour * 3600), _mass(mass), _length(length), _width(width), _height(height),
+	_Cp(Cp), _h(h), T_room_K(T_room_K), _cap_vs_temp(c_vs_t)
 {
-	_R = 0.004;
+	_R = R;
```
janinefreeman: is R defined elsewhere, or a new input?

shared/lib_battery.h
```diff
@@ -433,7 +433,7 @@ class lifetime_calendar_t
 {
 public:
 	lifetime_calendar_t(int calendar_choice, util::matrix_t<double> calendar_matrix, double dt_hour, 
-		float q0=1.02, float a=2.66e-3, float b=7280, float c=930);
+		float q0=1.02, float a=2.66e-3, float b=-7280, float c=930);
```
janinefreeman: why would this become negative?

shared/lib_battery.cpp
dguittet: So in this, I assumed the idea with the replacement_percent x was that we're replacing up to x percent of the battery. So if the current max capacity has degraded down to 50% of original and the replacement percent is 70, then we'll be adding a new 20% of original capacity. Hence, the capacity would increase by (70-50) * initial SOC.

I couldn't find much info on exactly what is meant in replacing the battery. But another interpretation would be to replace that 50% plus an additional 20%. Then the capacity would be equal to 70 * initial SOC.

Now that you've mentioned it, I think the second interpretation makes more sense. Especially with how replacement costs are probably calculated using that 70% rather than the 20% new capacity. What do you think?

shared/lib_battery.cpp
dguittet: nope, thanks!

shared/lib_battery.h
```diff
@@ -573,48 +573,50 @@ class thermal_t
...
 	thermal_t();
-	thermal_t(double dtHour, double mass, double length, double width, double height,
-		double Cp, double h, 
-		std::vector<double> T_room,
-		const util::matrix_t<double> &cap_vs_temp);
+	thermal_t(double dt_hour, double mass, double length, double width, double height, double R, double Cp,
+              double h, std::vector<double> T_room_K, const util::matrix_t<double> &c_vs_t);
 
-	// deep copy
+    thermal_t(const thermal_t& thermal);
```
janinefreeman: are lines 579, 583, and 586 duplicative? why getting rid of the const if it will be cloned/copied?

shared/lib_battery.cpp
dguittet: R actually is an SSC input already, batt_resistance

shared/lib_battery.h
dguittet: In the UI it's negative, and since lifetime_calendar_t was always called with all the inputs provided from the UI, the incorrectly provided default value (7280) wasn't ever used

ssc/cmod_battery.cpp
```diff
@@ -794,20 +794,21 @@ battstor::battstor(compute_module &cm, bool setup_model, size_t nrec, double dt_
...
-		batt_vars->batt_Cp, // [J/kgK]
-		batt_vars->batt_h_to_ambient, // W/m2K
-		batt_vars->T_room, // K
-		batt_vars->cap_vs_temp);
+            dt_hr,
+            batt_vars->batt_mass, // [kg]
+            batt_vars->batt_length, // [m]
+            batt_vars->batt_width, // [m]
+            batt_vars->batt_height,
+            batt_vars->batt_resistance, // [m]
```
janinefreeman: battery resistance in m, not Ohms?

shared/lib_battery.h
dguittet: 583 and 586 are the original functions for making a copy, and I added 579. In the refactoring, I'm getting rid of 583 ad 586. 

getting rid of the const?

ssc/cmod_battery.cpp
dguittet: Oops! thanks for catching that!

test/input_cases/battery_common_data.h
```diff
@@ -90,11 +90,11 @@ namespace {
 		ssc_number_t p_replacement_cost[1] = { 68 };
 		ssc_data_set_array(data, "om_replacement_cost1", p_replacement_cost, 1);
 		ssc_data_set_number(data, "batt_mass", 3102.717041015625);
-		ssc_data_set_number(data, "batt_length", 1.0689570903778076);
-		ssc_data_set_number(data, "batt_width", 1.0689570903778076);
-		ssc_data_set_number(data, "batt_height", 1.0689570903778076);
-		ssc_data_set_number(data, "batt_Cp", 1004);
-		ssc_data_set_number(data, "batt_h_to_ambient", 5);
+		ssc_data_set_number(data, "batt_length", 7);
```
janinefreeman: why did these inputs change?

test/shared_test/lib_battery_properties.h
```diff
@@ -89,14 +89,13 @@ class BatteryProperties : public ::testing::Test
 
 		// thermal
 		mass = 507;
-		length = 0.58;
-		width = 0.58;
-		height = 0.58;
+		length = 1.1;
```
janinefreeman: why did these input values change?

test/ssc_test/cmod_battery_test.cpp
```diff
@@ -21,7 +21,7 @@ TEST_F(CMBattery, CommercialLifetimePeakShaving_cmod_battery) {
 		// roundtrip efficiency test will ensure that the battery cycled
 		ssc_number_t roundtripEfficiency;
 		ssc_data_get_number(data, "average_battery_roundtrip_efficiency", &roundtripEfficiency);
-		EXPECT_NEAR(roundtripEfficiency, 94.129, 2);
+		EXPECT_NEAR(roundtripEfficiency, 90.06, 2);
```
janinefreeman: can you explain again why this efficiency would decrease so much with the changes you've implemented?

janinefreeman:
Love all the new tests! Have several questions, many of them probably dumb questions. :) I see that you only changed the answer to one existing test and asked for some more details there. How many of the tests that you added changed answers from their original values with your bug fixes, and can we review any answers that did change at the SAM meeting tomorrow?

@sjanzou if you have time to take a look at this as well, I'd appreciate that.

shared/lib_battery.h
janinefreeman: Not knowing exactly what b is, does it make sense for it to be negative? Just for thoroughness :)

shared/lib_battery.h
janinefreeman: gotcha, makes sense then if you'll be getting rid of 583 and 586. Isn't 579 casting what was passed in as a const thermal_t to not be const anymore, or am I misunderstanding that (highly likely)?

test/input_cases/battery_common_data.h
dguittet: Using the new thermal model, when the room temperature doesn't change, the battery temperature becomes primarily a function of the source term, I^2*R / surf_area / h. 

With the inputs provided here, the battery dimensions are too small for its power. It's current can go up to 3037 A for instance (batt_current_charge_max) and so it was overheating the battery way too much.

janinefreeman: So I imagine it was causing the test to fail, but only because it was a highly non-physical test to begin with?

test/shared_test/lib_battery_properties.h
dguittet: Same reason as above. I don't know any good estimates for thermal parameteres but this (paper)[https://iopscience.iop.org/article/10.1088/1757-899X/53/1/012014/pdf] provides an example set of parameters where the ratio of mass to surface area is lower than provided here originally

Screen Shot 2019-10-09 at 4 00 38 PM

test/ssc_test/cmod_battery_test.cpp
dguittet: average_roundtrip_efficiency = 100.*(_e_discharge_accumulated / (_e_charge_accumulated + _e_loss_system));

1. the thermal model results in more heating of the battery, and adds to the losses

2. in void capacity_kibam_t::updateCapacityForThermal(double capacity_percent), thermal losses were not being added to I_loss before.

dguittet:
All the unit tests were written during the bug fixes so I don't have any results for what values the models may have provided previously.

Mostly the changes are in the thermal model, the losses in capacity and replacements, which you've identified in your comments.

I had a since overwritten test showing why the thermal model was buggy:

Screen Shot 2019-10-02 at 10 13 02 AM

The room temp is 20C. In the third test, the battery has been cooled to -23 C.

shared/lib_battery.h
dguittet: class_name(const class_name& ) is the signature for the copy constructor, so it's how you would construct a new class_name from an existing one:

class_name a;
class_name b(a); 

In the copy constructor, you define how all the attributes of the class should be copied over. Some attributes can be done trivially (like if they are doubles) but if they are another class or a pointer, you'll need to specify how to make that copy.

test/input_cases/battery_common_data.h
dguittet: I think so

shared/lib_battery.h
dguittet: Good question though, thanks for all your questions so far!

shared/lib_battery.h
dguittet: The degradation graph produced looks reasonable. The model is based on this paper here, https://ieeexplore.ieee.org/document/7963578, but I don't see the relevant parameter from a quick skim.

shared/lib_battery.cpp
dguittet: I think I'll change it so that if you specify replacement of 70% that means you get rid of the old battery entirely and add a new battery of 70% of original battery

dguittet: Well actually the original code does this, where qmax0 is the nominal max capacity

```
_qmax += replacement_percent * 0.01* _qmax0;
_qmax = fmin(_qmax, _qmax0);
_q0 = _qmax*_SOC_init*0.01;
```
So we're adding to the degraded capacity some percent of the nominal, up to 100% of the nominal. This replacement follows a schedule or by a minimum capacity after which replacement occurs. 

However the replacement costs are not actually calculated using how much replacement has occurred. In lib_battery_dispatch.cpp 1613:

```
double capacityPercentDamagePerCycle = _Battery->lifetime_model()->cycleModel()->estimateCycleDamage();
m_cycleCost = 0.01 * capacityPercentDamagePerCycle * m_battReplacementCostPerKWH;
```
where estimateCycleDamage estimates the capacity lost in running the next cycle, whose range is the same as the average range thus far. This is a very different quantity than how much ACTUAL replacement has occurred * the cost to replace.

What I'm proposing:
1. track how much capacity is actually replaced
2. cycling cost = actual replaced capacity * cost per kwh

test/ssc_test/cmod_battery_test.cpp
janinefreeman: Have you found anything on round trip efficiency in the literature? Is this closer to expected than the higher number?

shared/lib_battery.h
janinefreeman: Thanks for the detailed explanations :)

test/ssc_test/cmod_battery_test.cpp
dguittet: I don't think this number reflects anything real world because we don't have real world temperature coefficients. But battery efficiency is generally higher, given 20C operating conditions and not too much heating inside the battery.

As an example from https://ieeexplore.ieee.org/document/4544139 :

image

shared/lib_battery.cpp
janinefreeman: Hmm. Maybe we can ask around about what it means to replace part of a battery. I don't know which of your proposed capacity calculations (or the original definition) make the most sense. But I do think that your proposed cycle cost makes sense.

dguittet:
Much of this was completed in the `batt_refactor` merge

#362

PR #314, opened 2019-10-15
Expanding bifacial capabilities and bringing bifacial modeling into alignment with PVsyst

caseyzak24:
As I've discussed with collaborators in the past, for SAM/`ssc` to be useful to us at CCR it has to be able to generate results that are similar to the current dominant bankable energy model, which is, (for now) PVsyst.

This PR is the result of my efforts to bring the 2 models into alignment. Using these changes, we are seeing good agreement between the 2 models in monthly DC production across a range of locations and designs (**see figure below, error as a decimal is on the y-axis**).
image

As noted in the commit messages, all of these changes may not be the right choice for `ssc` (or even the right approach in general). I implemented PVsyst's approach to frontside ground diffuse contribution. **Note this will result in differences between monofacial and bifacial sims that have nothing to do with backside phenomena**. I changed the IAM model used for diffuse in the bifacial model to ASHRAE (again likely not as realistic but resulted in better agreement). I also changed the transpo model used from perez to isotropic.

I submitted all these changes together because that was easiest and it is the version I have benchmarking results for; please feel free to pick and choose the parts that make the most sense for this project.


dguittet:
Thanks for the contribution! We'll have to look closely at these changes and discuss which ones make sense to add in.

janinefreeman:
As an update, we will probably be reviewing these changes after our next SAM release, unless there are any bug fixes in here that you think we should include @caseyzak24 . Thanks again for your contributions!

PR #315, opened 2019-10-15
Fix: iec61853 module model errors

caseyzak24:
I found a handful of errors/typos in the iec61853 moduel model that were significantly affecting results and corrected them. I also changed the IAM model to consider sky and ground diffuse effects using the same approach as the cec and mlm models.

### DIFF
--- a/shared/lib_iec61853.cpp
+++ b/shared/lib_iec61853.cpp
@@ -1,22 +1,22 @@
 /**
 BSD-3-Clause
 Copyright 2019 Alliance for Sustainable Energy, LLC
-Redistribution and use in source and binary forms, with or without modification, are permitted provided 
+Redistribution and use in source and binary forms, with or without modification, are permitted provided
 that the following conditions are met :
-1.	Redistributions of source code must retain the above copyright notice, this list of conditions 
+1.	Redistributions of source code must retain the above copyright notice, this list of conditions
 and the following disclaimer.
-2.	Redistributions in binary form must reproduce the above copyright notice, this list of conditions 
+2.	Redistributions in binary form must reproduce the above copyright notice, this list of conditions
 and the following disclaimer in the documentation and/or other materials provided with the distribution.
-3.	Neither the name of the copyright holder nor the names of its contributors may be used to endorse 
+3.	Neither the name of the copyright holder nor the names of its contributors may be used to endorse
 or promote products derived from this software without specific prior written permission.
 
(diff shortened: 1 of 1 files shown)


dguittet:
Great, thanks for fixing those errors! I'll take a look at these.

shared/lib_iec61853.cpp
```diff
@@ -897,7 +902,7 @@ bool iec61853_module_t::operator() ( pvinput_t &input, double TcellC, double opv
 		double Ilop = tpoa/1000*(Il + alphaIsc*(Tc-298.15));
 		double Egop = (1-0.0002677*(Tc-298.15))*Egref;
 		double Ioop = Io*pow(Tc/298.15,3.0)*exp( 11600 * (Egref/298.15 - Egop/Tc));
-		double Rsop = D1 + D2*(Tc-298.15) + D3*( 1-tpoa/1000.0)*pow(1000.0/poa,2.0);
+		double Rsop = D1 + D2*(Tc-298.15) + D3*( 1-tpoa/1000.0)*pow(1000.0/tpoa,2.0);
```
janinefreeman: Good catch, thank you!

shared/lib_iec61853.cpp
```diff
@@ -858,27 +858,32 @@ bool iec61853_module_t::operator() ( pvinput_t &input, double TcellC, double opv
...
+		iamf_gnd = iam( theta_gnd, GlassAR );
+
+        tpoa = iamf_beam * input.Ibeam + iamf_diff * input.Idiff + iamf_gnd * input.Ignd;
+        if( tpoa < 0.0 ) tpoa = 0.0;
+        if( tpoa > poa ) tpoa = poa;
 	
 		// spectral effect via AM modifier
 		double ama = air_mass_modifier( input.Zenith, input.Elev, AMA );
 		tpoa *= ama;
+		AOIModifier = tpoa/poa;
```
janinefreeman: This looks overall consistent with other models. From a code maintenance perspective, would it make sense to call calculateIrradianceThroughCoverDeSoto in lib_pv_incidence_modifier.cpp instead of re-implementing these equations? I didn't review in enough detail to see if there are any differences.

caseyzak24: @janinefreeman the calculateIrradianceThroughCoverDeSoto function does not consider the effects of antireflective coatings. Since most installed modules have them these days, it's important to capture those effects. It does seem worth it to modify the calculateIrradianceThroughCoverDeSoto function to account for ARC and then use this function in the iec61853, pvwatts and cec module models.

PR #327, opened 2019-12-19
add out-of-memory exception handling

dguittet:
add try-catch block around resiliency to catch out of memory killer from bad_alloc

have two possibilities in tests: either out of memory killed or check all values

--------

edit: further testing shows memory leaks are probably the actual issue


dguittet:
This check doesn't work on linux, which overcommits memory as explained here: https://stackoverflow.com/questions/51674575/c-bad-alloc-exception-not-catched

On linux, the way to prevent a hard crash of SAM would be to change the system settings... `sudo sh -c 'echo 2 > /proc/sys/vm/overcommit_memory'`

dguittet:
thanks @daniel-thom for tips on how to start tracking down the problem!

fixed a bunch of memory leaks in battery-related class destructors

battery & resilience tests passing with no leaks on windows, linux & mac

PR #331, opened 2020-02-06
Batt bug

dguittet:
From bugs reported in NREL/sam#240 and NREL/sam#239, there were a few battery bugs:

1. the battery was discharging even when the state of charge had reached its minimum limit

- This was fixed by propagating changes to the current in `battery_t::run` so that the state of charge limits enforced

2. the pv + battery system output wasn't properly accounting for the energy pulled from the grid, which would reduce the system output (making it negative), which resulted in utility rate not accounting for this purchased energy

- This was fixed by re-adding the (negative) battery charge power into the `gen` output. Tests were written for `BatteryPowerFlow` which revealed some operational differences between AC and DC connected batteries. 

- For instance, when the battery is told to charge, in AC systems the battery will charge only after load has been met, but in DC systems, the battery will prefer to charge first from PV then grid and meeting the load from the grid.

### DIFF
--- a/shared/lib_battery.cpp
+++ b/shared/lib_battery.cpp
@@ -1564,6 +1564,7 @@ void thermal_t::updateTemperature(double I, double R, double dt, size_t lifetime
 		_T_battery = implicit_euler(I, dt*HR2SEC, lifetimeIndex);
 	else
 		_message.add("Computed battery temperature below zero or greater than max allowed, consider reducing C-rate");
+	_T_battery = fmax(_T_battery, _T_room[util::yearOneIndex(_dt_hour, lifetimeIndex)]);
 }
 
 double thermal_t::f(double T_battery, double I, size_t lifetimeindex)
@@ -1857,7 +1858,7 @@ double battery_t::calculate_max_discharge_kw(double *max_current_A) {
     return _voltage->calculate_max_discharge_w(q, qmax, _thermal->T_battery(), max_current_A) / 1000.;
 }
 
-double battery_t::run(size_t lifetimeIndex, double I)
+double battery_t::run(size_t lifetimeIndex, double &I)
 {
 	// Temperature affects capacity, but capacity model can reduce current, which reduces temperature, need to iterate
 	double I_initial = I;
--- a/shared/lib_battery_dispatch.cpp
+++ b/shared/lib_battery_dispatch.cpp
@@ -774,7 +774,6 @@ void dispatch_automatic_t::dispatch(size_t year,
 
 bool dispatch_automatic_t::check_constraints(double &I, size_t count)
 {
-    m_batteryPowerFlow->calculate();
     // check common constraints before checking manual dispatch specific ones
 	bool iterate = dispatch_t::check_constraints(I, count);
 
@@ -784,13 +783,13 @@ bool dispatch_automatic_t::check_constraints(double &I, size_t count)
 		double P_battery = I * _Battery->battery_voltage() * util::watt_to_kilowatt;
 		double P_target = m_batteryPower->powerBatteryTarget;
 
-		// Comomon to automated behind the meter and front of meter
+		// Common to automated behind the meter and front of meter
 		iterate = true;
-		
-		
+
+
 		// Don't respect target if bidirectional inverter efficiency is low while charging
-		if (m_batteryPower->connectionMode == dispatch_t::DC_CONNECTED && 
-			m_batteryPower->sharedInverter->efficiencyAC <= m_batteryPower->inverterEfficiencyCutoff && 
+		if (m_batteryPower->connectionMode == dispatch_t::DC_CONNECTED &&
+			m_batteryPower->sharedInverter->efficiencyAC <= m_batteryPower->inverterEfficiencyCutoff &&
 			P_target < 0)
 		{
 			iterate = false;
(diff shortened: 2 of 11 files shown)


shared/lib_battery.cpp
```diff
@@ -1564,6 +1564,7 @@ void thermal_t::updateTemperature(double I, double R, double dt, size_t lifetime
 		_T_battery = implicit_euler(I, dt*HR2SEC, lifetimeIndex);
 	else
 		_message.add("Computed battery temperature below zero or greater than max allowed, consider reducing C-rate");
+	_T_battery = fmax(_T_battery, _T_room[util::yearOneIndex(_dt_hour, lifetimeIndex)]);
```
sjanzou: Why maxed at room temperature or battery temp? Why can't the battery be below room temperature? Is this based on Troom being a conditioned area for battery storage?

shared/lib_battery.cpp
```diff
@@ -1857,7 +1858,7 @@ double battery_t::calculate_max_discharge_kw(double *max_current_A) {
     return _voltage->calculate_max_discharge_w(q, qmax, _thermal->T_battery(), max_current_A) / 1000.;
 }
 
-double battery_t::run(size_t lifetimeIndex, double I)
+double battery_t::run(size_t lifetimeIndex, double &I)
```
sjanzou: Can double &I be changed to const double &I?

shared/lib_battery_dispatch.cpp
```diff
@@ -774,7 +774,6 @@ void dispatch_automatic_t::dispatch(size_t year,
 
 bool dispatch_automatic_t::check_constraints(double &I, size_t count)
 {
-    m_batteryPowerFlow->calculate();
```
sjanzou: Why does the batteryPowerFlow no longer have to be calculated when calling the constraint check? Is the power flow ran outside of the call?

shared/lib_battery_powerflow.cpp
```diff
@@ -267,20 +267,18 @@ void BatteryPowerFlow::calculateACConnected()
 
 	// Compute total system output and grid power flow
 	P_grid_to_load_ac = P_load_ac - P_pv_to_load_ac - P_batt_to_load_ac - P_fuelcell_to_load_ac;
-	P_gen_ac = P_pv_ac + P_fuelcell_ac + P_inverter_draw_ac - P_system_loss_ac;
```
sjanzou: Is this where the improper accounting was fixed with new variables?

test/shared_test/lib_battery_powerflow_test.cpp
```diff
@@ -7,7 +7,7 @@
 #include "lib_shared_inverter.h"
 
 
-void BatteryPowerFlowTest::SetUp()
+void BatteryPowerFlowTest_lib_battery_powerflow::SetUp()
```
sjanzou: Thanks for adding the flow test!

test/ssc_test/cmod_battery_test.cpp
```diff
@@ -23,7 +23,7 @@ TEST_F(CMBattery_cmod_battery, CommercialLifetimePeakShaving) {
 		// roundtrip efficiency test will ensure that the battery cycled
 		ssc_number_t roundtripEfficiency;
 		ssc_data_get_number(data, "average_battery_roundtrip_efficiency", &roundtripEfficiency);
-		EXPECT_NEAR(roundtripEfficiency, 82.75, 2);
+		EXPECT_NEAR(roundtripEfficiency, 94.42, 2);
```
sjanzou: Quite a jump in efficiency - is this because of the previously lost energy?

shared/lib_battery.cpp
dguittet: I don't think the battery operations can cool the battery, which was what was happening due to the large step size in the numerical methods: if the current decreased dramatically, the temperature could decrease so much (as the approximate derivative was applied across the whole hour) as to be cooling the battery. If the current falls precipitously to zero, the temperature of the battery would simply to be room temp.

shared/lib_battery.cpp
dguittet: No, it needs to pass back to the caller of `run` the changes that are made to `I`. The bug this addresses is how the battery was sometimes discharging phantom power: https://github.com/NREL/SAM/issues/240#issuecomment-581515175

The reason that was happening is that the current was too high, and the `run` function would reduce it upon checking the SOC constraints, but that information was not passed back out, so the battery itself did not discharge any current, but the battery controller thought the original `I` was used.

Ideally, this logic would be performed in the dispatch classes' `check_constraints` method, but due to the iterative looping of `check_constraints` with `run`, the current would bounce back and forth rather than settling on a number which satisfies all constraints. I didn't want to modify the `check_constraints` without having more time to do so, so prioritizing the constraints in `run` seemed like the best option for now.

shared/lib_battery_dispatch.cpp
dguittet: Right, I'd added this line here during debugging a few commits ago and forgot to remove it actually. But yup, the power flow calculation is run by `battery:run` before `check_constraints` is run.

shared/lib_battery_powerflow.cpp
dguittet: Yes, I pretty much restored it to what it was before my last changes when I broke the definition of gen while trying to fix the unallowed grid charging bug. So I just restored the original calculation of gen, didn't add any new variables, and refined the error-checking in line 276 so that it should grid-charging preference is enforced.

test/ssc_test/cmod_battery_test.cpp
dguittet: I think it's actually due to the temperature floor (battery capacity decreasing with temperature).

test/shared_test/lib_battery_powerflow_test.cpp
dguittet: Thanks so much for the review!

PR #336, opened 2020-02-26
code cleanup following .editorconfig

dguittet:
We've discussed coding style guidelines and generally agreed that something such as google's c++ guide would be worth trying out.

Using an .editorconfig file with EditorConfig plugin allows reformatting existing code and automatic consistency of new code.


janinefreeman:
Oooh I'm gonna have to agree strongly with Ty on the curly brace being on its own line...

Matthew-Boyd:
Yes, I've used in successfully. I copied the one from wxWidgets-3.1.1 and put it in sam_dev/ssc/ . I did this to convert Tabs to spaces, as the VS2019 options for doing so weren't being respected. I later had to change end_of_line in the .editorconfig file to crlf as git was always warning me. (I'm on Windows.)

brtietz:
Closing due to age. May attempt again after upcoming release. Keeping branch around in case there's anything useful.

PR #342, opened 2020-03-18
PV Battery Integration Tests

brtietz:
Status:

Ready to review?

*  Integration tests have been developed for all dispatch modes, with both residential and PPA controllers, while checking dispatch powers, energies, and cycles per day.
*  I've added unit tests for manual dispatch and automatic behind the meter controllers. Darice has added front of meter controller tests.
*  Some fixes have been cherry-picked from patch so this MR might be a little cleaner if patch is merged into develop first.


test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -0,0 +1,448 @@
...
+			auto annualDischargeEnergy = data_vtab->as_vector_ssc_number_t("batt_annual_discharge_energy");
+			EXPECT_NEAR(annualDischargeEnergy[0], expectedBatteryDischargeEnergy[i], m_error_tolerance_hi) << "Battery annual discharge energy.";
+
+			auto batt_power = data_vtab->as_vector_ssc_number_t("batt_power");
+			daily_battery_stats batt_stats = daily_battery_stats(batt_power);
+
+			EXPECT_NEAR(batt_stats.peakKwCharge, peakKwCharge[i], m_error_tolerance_lo);
+			EXPECT_NEAR(batt_stats.peakKwDischarge, peakKwDischarge[i], m_error_tolerance_lo);
+			EXPECT_NEAR(batt_stats.peakCycles, peakCycles[i], m_error_tolerance_lo);
+			EXPECT_NEAR(batt_stats.avgCycles, avgCycles[i], 0.0001); // Runs once per year
```
dguittet: Does this "Run once per year" (per the comment)? Looks more frequent than that.

test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -0,0 +1,448 @@
...
```
dguittet: I really like this. I was wondering if you could compute `daily_battery_stats` on `custom_dispatch_residential_schedule` and see how the stats differ? It probably won't be very much due to the low activity.

Here's another dispatch csv you can try instead. I pulled it from the reopt sizing function for pvsamv1 residential a while back, and while it's not fitted to the parameters here, I think it's about the right battery size for the load.
[custom_dispatch_residential_hourly.csv.zip](https://github.com/NREL/ssc/files/4383903/custom_dispatch_residential_hourly.csv.zip)

test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+	ssc_number_t expectedEnergy = 37321520;
+	ssc_number_t expectedBatteryChargeEnergy = 2030;
+	ssc_number_t expectedBatteryDischargeEnergy = 3175;
+
+	ssc_number_t peakKwCharge = -1020.4;
+	ssc_number_t peakKwDischarge = 958.7;
+	ssc_number_t peakCycles = 1;
+	ssc_number_t avgCycles = 0.0027;
+
```
dguittet: [custom_dispatch_singleowner_hourly.csv.zip](https://github.com/NREL/ssc/files/4383925/custom_dispatch_singleowner_hourly.csv.zip)

Here's another custom dispatch you can try. Again, not perfect but gives a bit more signal to test

test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -0,0 +1,448 @@
...
+	singleowner_defaults(data);
+
+	ssc_number_t expectedEnergy[3] = { 37322265, 37320217, 37321282 };
+	ssc_number_t expectedBatteryChargeEnergy[3] = { 0, 14062, 6668 }; // No rate model means battery use is low
+	ssc_number_t expectedBatteryDischargeEnergy[3] = { 1417, 13430, 7102 };
+
+	ssc_number_t peakKwCharge[3] = { 0, -1028.5, -1028.5 };
+	ssc_number_t peakKwDischarge[3] = { 967.5, 966.8, 966.8 };
+	ssc_number_t peakCycles[3] = { 0, 1, 1};
+	ssc_number_t avgCycles[3] = { 0, 0.0005, 0.0003};
```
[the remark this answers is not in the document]
brtietz: Thanks! Activity on these tests has increased, as expected.

test/ssc_test/cmod_battery_pvsamv1_test.cpp
brtietz: I like these sparse schedules as tests, since they give me insight into the battery stats over a longer period, and dispatch the system at capacity. I added your files to new tests, thanks for sending them!

test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -93,8 +93,8 @@ TEST_F(CMPvsamv1BatteryIntegration_cmod_pvsamv1, ResidentialACBatteryModelIntegr
 
 	ssc_number_t peakKwCharge[3] = { -2.81, -2.87, -2.25 };
 	ssc_number_t peakKwDischarge[3] = { 1.39, 1.30, 0.97 };
-	ssc_number_t peakCycles[3] = { 1, 1, 1 };
-	ssc_number_t avgCycles[3] = { 1, 0.9973, 0.4904 };
+	ssc_number_t peakCycles[3] = { 1, 2, 1 };
```
dguittet, line 96: Turns out this result was different on Mac vs Linux/Windows.

In Linux, some of the batt_power entries were `-0.000` and in Mac, this was simply `0.0000`, and this resulted in a different of which cycle state was being assigned where `-0.000` was being assigned to `cycleState=-1` and triggering an extra cycle.

brtietz:
We're seeing intermittent failures in the following tests, so I'm going to hold off merging this until that issue is solved:

* AutoFOM_lib_battery_dispatch.DispatchFOM_DCCustomCharge
* AutoFOM_lib_battery_dispatch.DispatchFOM_DCCustomChargeSubhourly
* ManualTest_lib_battery_dispatch.ManualGridChargingOnDCConnectedTest

Please comment here if you see additional tests that fail on this branch (intermittent or otherwise)

brtietz:
Merged via squash commit https://github.com/NREL/ssc/commit/caa527c897596181b4788ab8dbf6e726e6bbd207

Intermittent test failures were caused by inverter cutoff limit not being initialized in test code. Shouldn't affect compute modules.

Issue #346, opened 2020-03-25
Subhourly battery dispatch can switch between charging and discharging too frequently


brtietz:
**Describe the bug**
Under certain conditions, `dispatch_t`'s `_t_at_mode` variable can be updated twice in a given timestep, causing a controller to change from charging to discharging (or vice versa) sooner than specified.

**To Reproduce**
TBD how this error affects things at the GUI level, but a unit test at https://github.com/NREL/ssc/blob/pv_battery_integration_tests/test/shared_test/lib_battery_dispatch_test.cpp confirms the existence of the issue. This would only be seen in sub-hourly simulation modes.

**Expected behavior**
Dispatch should change at the interval specified in dispatch controller construction.

brtietz:
Closed in #351

PR #348, opened 2020-03-26
Merge patch into develop

janinefreeman:
Hi team- please review all things that you worked on in patch and **comment** on the pull request if something that Git has decided should be merged into develop is incorrect. (Do this by looking at "files changed"). Please comment on the PR once you have reviewed, so that I know when we're ready to merge. I will handle the final merging of patch into develop. After that, every time you update patch, please then merge your changes into develop.

PR #351, opened 2020-03-30
remove duplicate timestep iteration in switch_controller function. Fi…

brtietz:
Remove the extra increment to _t_at_mode when changing controllers. Tests on pv_battery_integration_test. Solves #346.

Let me know if this should go to patch or develop.


brtietz:
Merged into patch. I'm going to let https://github.com/NREL/ssc/pull/348 merge this into develop, since that will resolve the merge conflict in test/CMakeLists.txt

PR #352, opened 2020-04-01
LandBOSSE Cost Model Integration

dguittet:
Features
In PPA configurations of Wind, add LandBOSSE v2.2.7.3 cost model in System Costs page.

- Installation triggered by "enable" button
- Can be disabled by same button from simulation
- Run within equations, parametrics & regular simulation

**Code changes**
- Add jsoncpp library
- add cmod_wind_landbosse.cpp
- add set_python_path to sscapi
- Add conversion functions between var_table & var_data types to json
- Add tests for wind_landbosse
- Add var_table::merge function

part of https://github.com/NREL/SAM/issues/270


ssc/cmod_wind_landbosse.cpp
```diff
@@ -0,0 +1,350 @@
...
+std::string cm_wind_landbosse::call_python_module_windows(const std::string& input_dict_as_text) {
+	STARTUPINFO si;
+	SECURITY_ATTRIBUTES sa;
+	PROCESS_INFORMATION pi;
+    HANDLE g_hChildStd_IN_Rd = NULL;
+    HANDLE g_hChildStd_OUT_Wr = NULL;
+    HANDLE g_hChildStd_OUT_Rd = NULL;
+    HANDLE g_hChildStd_IN_Wr = NULL;
+    HANDLE stderr_rd = NULL;
+    HANDLE stderr_wr = NULL;  //pipe handles
```
daniel-thom, line 219: We need to use consistent variable names that do no imply they are global.

ssc/cmod_wind_landbosse.cpp
```diff
@@ -0,0 +1,350 @@
...
+            = std::chrono::system_clock::now() + std::chrono::seconds(60 * 5);
+
+    if(std::future_status::ready == f_completes.wait_until(time_passed))
+        return f_completes.get();
+    else
+        throw exec_error("wind_landbosse", "python handler error. Python process timed out.");
+}
+
+#ifdef __WINDOWS__
+std::string cm_wind_landbosse::call_python_module_windows(const std::string& input_dict_as_text) {
```
daniel-thom, line 210: Need to make this function use consistent alignment and curly braces.

ssc/cmod_wind_landbosse.cpp
```diff
@@ -0,0 +1,350 @@
...
+            else if (bread_last > 0)
+            {
+                break;
+            }
+        }
+
+        CloseHandle(pi.hThread);
+        CloseHandle(pi.hProcess);
+    }
+   error:
```
daniel-thom, line 287: This should be `done:` instead of `error:`.

ssc/cmod_wind_landbosse.cpp
```diff
@@ -0,0 +1,350 @@
...
+    if (!CreatePipe(&g_hChildStd_IN_Rd, &g_hChildStd_IN_Wr, &sa, 0)) {
+        goto error;
+    }
+    if (!CreatePipe(&g_hChildStd_OUT_Rd, &g_hChildStd_OUT_Wr, &sa, 0)) {
+        goto error;
+    }
+    if (!CreatePipe(&stderr_rd, &stderr_wr, &sa, 0)) {
+        goto error;
+    }
+
```
daniel-thom, line 244: Per the Microsoft example we should call SetHandleInformation to set g_hChildStd_IN_Wr, g_hChildStd_OUT_Rd, and stderr_rd to ensure that they are not inherited.

https://docs.microsoft.com/en-us/windows/win32/procthread/creating-a-child-process-with-redirected-input-and-output

ssc/cmod_wind_landbosse.cpp
```diff
@@ -0,0 +1,350 @@
...
+        for (;;)
+        {
+            PeekNamedPipe(g_hChildStd_OUT_Rd, buf, BUFSIZE - 1, &bread, &avail, NULL);
+            //check to see if there is any data to read from stdout
+            if (bread != 0)
+            {
+                if (ReadFile(g_hChildStd_OUT_Rd, buf, BUFSIZE - 1, &bread, NULL))
+                {
+                    output += std::string(buf);
```
daniel-thom, line 274: `output` vs `buf`?

ssc/cmod_wind_landbosse.cpp
```diff
@@ -0,0 +1,350 @@
...
+    for (HANDLE handle : handles) {
+        if (handle && handle != INVALID_HANDLE_VALUE) {
+            CloseHandle(handle);
+        }
+    }
+	return buf;
+}
+#endif
+
+void cm_wind_landbosse::cleanOutputString(std::string& output_json) {
```
daniel-thom, line 298: Can we push this logic to LandBOSSE?  Or make a wrapper function to JSON-ify the dictionary?  There could be other problems in the future if they add booleans (Python uses True/False instead of true/false).

dguittet: Yes, I think that's a good idea to have LandBOSSE use the json module and only return a valid json string.

@parangat94 Would it be possible to have this in the next version? Instead of returning a dictionary, could you convert it via the json module to a valid json string, to be returned from `run_landbosse`?

daniel-thom: Another option is for the run function to accept a `json` parameter that tells it to format the output as JSON.  Then it could support dictionaries (for native Python callers) or JSON.

ssc/cmod_wind_landbosse.cpp
dguittet: `buf` takes up segments separated by newlines. In the original code where std error and std cout were being read into the same buffer, this was used to continue reading from buffer after initial error lines. I guess we don't need that anymore since we will only be expecting single line from std cout as a json string.

daniel-thom: My comment was really about the fact that `output` is essentially unused.  The function returns `buf`.

dguittet: removed `output`

ssc/cmod_wind_landbosse.cpp
dguittet: added

ssc/cmod_wind_landbosse.cpp
dguittet: renamed

ssc/cmod_wind_landbosse.cpp
dguittet: edited

ssc/cmod_wind_landbosse.cpp
dguittet: renamed

ssc/cmod_wind_landbosse.cpp
```diff
@@ -0,0 +1,354 @@
...
+	std::string input_dict_as_text = input_json;
+	std::replace(input_dict_as_text.begin(), input_dict_as_text.end(), '\"', '\'');
+
+    load_config();
+#ifdef __WINDOWS__
+	std::string output_json = call_python_module_windows(input_dict_as_text);
+#else
+    std::string output_json = call_python_module(input_dict_as_text);
+#endif
+
```
daniel-thom, line 337: Should we throw a specific error if `output_json` is emtpy?  That condition could happen in the Windows function but isn't really handled.

dguittet: Good point, we do that for the Unix function. Added.

dguittet:
All ssc tests passing

ssc/cmod_wind_landbosse.cpp
dguittet: I'll add this to the "run_cmd" in landbosse.json

dguittet: @parangat94 But the issue is that LandBOSSE prints a failure message. This is what gets read from std out:

```
"Fail SAM_Run FoundationCost\n{\"errors\": [\"Error in NegativeInputError: User entered a negative value for depth. This is an invalid entry\", \"Error in FoundationCost: cannot convert float NaN to integer\"]}\n"
```

We want it to just say
```
"{\"errors\": [\"Error in NegativeInputError: User entered a negative value for depth. This is an invalid entry\", \"Error in FoundationCost: cannot convert float NaN to integer\"]}\n"
```

dguittet: Adding this to Jira and resolving for now.

daniel-thom: @parangat94 Can you make LandBOSSE write all error messages to stderr instead of stdout?  Basically, you would add `, file=sys.stderr` to all your calls to `print` when there is an error.

PR #356, opened 2020-04-10
restore build_sdk files

dguittet:
Restoring the deleted build_sdk folder to patch & develop. @sjanzou sorry for mistakenly deleting these. These all belong in the ssc repo correct?


dguittet:
@sjanzou Could you take a look a this please?

sjanzou:
@dguittet, we no longer distribute the sdk from this build process. The potential reason to restore is to have some code examples that would potentially be lost.

sjanzou:
@cpaulgilman, @janinefreeman, please check through the 187 files that were removed and see if we would like to include any of the examples in the SAM-Documentation repo or if we want to keep the build_sdk folder in the ssc public repo or add the files to an ssc wiki.

janinefreeman:
@sjanzou Are the system requirements for the various wrappers located somewhere in the code generator code now?

The file  build_sdk/apiwrappers/python/openei_utilityrate_billcalculator.py and associated files could be worth keeping since it shows the details of downloading a rate from the URDB and calling a utility rate compute module, although it is utilityrate3 so maybe too outdated to be useful. I'll defer to @cpaulgilman on other files in the examples folder- a lot of them are old, but I'm not sure we have any other examples of the C wrappers etc.

sjanzou:
@janinefreeman  No, no system requirements in the generated code folders. There several user support issues involving code generation. We could keep the pdfs and copy them over to the code generation folder.

cpaulgilman:
@sjanzou Is this an old pull request that can be safely closed without merging?

sjanzou:
@cpaulgilman, yes, this is old and out of date and the website has many more recent examples.

PR #360, opened 2020-05-01
bring back the travis code that runs all the tests now that our timeo…

brtietz:
Travis extended our build times to 120 minutes, so revert the yml script so it now runs all the tests.

3 test appear to be failing on Mac - I assume that should be addressed on another branch?


brtietz:
Handled in https://github.com/NREL/ssc/commit/9b599e86e4bce90aaf98d79ef80989ef571530a4

PR #371, opened 2020-05-18
Comments for pysam

brtietz:
Comments in the labels field to address https://github.com/NREL/pysam/issues/31

Questions for review:
(1) Will this break anything?
(2) Should I give pvsamv1 the same treatment?
(3) Is there any easy way to get high level comments like I put at the top of the file into the readthedocs output? Nate had some PVWatts usage recommendations he wanted to include based on this, which I added to the cpp file. Is there a better spot for this?

### DIFF
--- a/ssc/cmod_pvwattsv7.cpp
+++ b/ssc/cmod_pvwattsv7.cpp
@@ -35,6 +35,11 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "lib_pv_incidence_modifier.h"
 #include "lib_cec6par.h"
 
+/*
+ A simplified pv model with fewer inputs than pvsamv1. 
+ Does not do detailed degradation or loss modeling. If those are important, please use pvsamv1
+ */
+
 class lossdiagram
 {
 	unordered_map< std::string, double > m_map;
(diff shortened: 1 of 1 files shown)


dguittet:
1. Modifying the label, units, or adding comments won't break anything. PySAM interface is broken by changing the name, type or group however.

2. PySAM reuses these labels, so these changes will be reflected.

3. Yes, there is actually! In the RTD, each model has a "<name> model description" (see [MerchantPlant](https://nrel-pysam.readthedocs.io/en/master/modules/Merchantplant.html)) which ideally will include such information. This documentation is read from this [export_config file](https://github.com/NREL/SAM/blob/004775513ae83ff62d6b5a5a9cbd10aebbb46003/api/api_autogen/builder_PySAM.cpp#L42) which was originally a copy of the configuration descriptions in [startup.lk](https://github.com/NREL/SAM/blob/004775513ae83ff62d6b5a5a9cbd10aebbb46003/deploy/runtime/startup.lk#L78). 

I used a copy rather than reading directly from `startup.lk` to enable PySAM-specific documentation, but now I realize I should not maintain a separate copy. But for now, would you mind adding those comments to both the long description in `startup.lk` and the `module_docs`?

janinefreeman:
@brtietz I do not think this accurately reflects what the code is actually doing. In lifetime mode, pvwattsv7 does apply DC degradation correctly on the DC side. I think this merge may need to be reverted. Same with pvsamv1. Can you please confirm?

PR #372, opened 2020-05-18
add "gen" as output to PVWatts

dguittet:
Expose "gen" in var table as lifetime system power output. Originally hidden since lifetime PVWatts was not used in GUI since the degradation isn't explicitly simulated, but now that it is being used in the GUI, the variable should be listed for export.

Addresses issue NREL/pysam#31. Also see #371

### DIFF
--- a/ssc/cmod_pvwattsv5.cpp
+++ b/ssc/cmod_pvwattsv5.cpp
@@ -85,7 +85,7 @@ static var_info _cm_vtab_pvwattsv5_part2[] = {
 
         { SSC_OUTPUT,       SSC_ARRAY,       "dc",                             "DC array power",                              "W",     "",                         "Time Series",      "*",                       "",                          "" },
         { SSC_OUTPUT,       SSC_ARRAY,       "ac",                             "AC inverter power",                           "W",     "",                         "Time Series",      "*",                       "",                          "" },
-
+        { SSC_OUTPUT,       SSC_ARRAY,       "gen",                            "AC system power (lifetime)",                  "kWh",       "",                                             "Time Series",      "*",                       "",                          "" },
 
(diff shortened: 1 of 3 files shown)


janinefreeman:
@dguittet Look at line 346 in pvwattsv7- this is conditionally declaring the "gen" output variable in common.cpp if a battery isn't enabled. It doesn't declare "gen" if the battery is enabled because "gen" gets declared by the battery compute module. So I think "gen" should already be available as a lifetime output in pvwattsv7 via the SDK- can you verify?

dguittet:
Thanks for the input. If `gen` isn't declared in an [added var_info](https://github.com/NREL/ssc/blob/ca2feffb1c1383f7a9b27aca3745e83fe393e65f/ssc/cmod_pvwattsv7.cpp#L278) for the compute module, then it's not visible to the SDK. So that's why PySAM doesn't have a gen output, nor does the SDKtool for PVWatts.

The variable can be added to multiple compute_modules' var_info tables but will point to the same underlying data. 

With a battwatts running after PVWatts, Battwatts.gen == PVWatts.gen. This means the battery will overwrite the gen.

janinefreeman:
It is an add_var_info, line 345-346: 

`		if (!as_boolean("batt_simple_enable"))
			add_var_info(vtab_technology_outputs);`

So seems that it should be showing up in PySAM?

dguittet:
Janine and I determined that the `gen` output should be declared in the initializer in order to show up for the SDKtool and PySAM.

@janinefreeman Should we still enable lifetime outputs for PVWatts?

janinefreeman:
I think it makes sense for "gen" to be enabled for lifetime in PySAM, but I'm not sure we can go through and systematically enable others right now. Are there other considerations I should be thinking of?

dguittet:
I think it's fine to just add it for PVwatts.

PR #375, opened 2020-05-21
Sam 315 energy accumulation fuel cell

brtietz:
pvwatts7 was throwing an error when run in a subhourly fuel cell configuration (see https://github.com/NREL/SAM/issues/315). Fix the error by migrating from accumulate_monthly to accumulate_monthly_for_year.

It looks like this solves a second bug, where even on an hourly fuel cell configuration, all of the energy was being placed in January. Before fix:

fuel_cell_before

After fix:

fuel_cell_after

Mysteries: why did pvwattsv7 work subhourly without the fuel cell?

Questions: do any additional instances of accumulate_monthly need to change?

### DIFF
--- a/ssc/cmod_pvwattsv7.cpp
+++ b/ssc/cmod_pvwattsv7.cpp
@@ -1217,11 +1217,10 @@ public:
 			}
 
 			wdprov->rewind();
-			if (y == 0) {
-				accumulate_monthly("gen", "monthly_energy", ts_hour);
-				accumulate_annual("gen", "annual_energy", ts_hour);
-			}
 		}
+
+		accumulate_monthly_for_year("gen", "monthly_energy", ts_hour, step_per_hour);
+		accumulate_annual_for_year("gen", "annual_energy", ts_hour, step_per_hour);
 
 		accumulate_monthly("dc", "dc_monthly", 0.001*ts_hour);
 		accumulate_monthly("ac", "ac_monthly", 0.001*ts_hour);


janinefreeman:
I think it makes sense that subhourly PVWatts would work, because that sounds like a bug in lifetime mode, and plain old PVWatts doesn't run in lifetime mode except through the SDK. What's a mystery to me is, why was hourly fuel cell working?

dguittet:
Thanks Brian for finding the issue.

Looks like `system_use_lifetime_output` is set to 0 for PVWatts but 1 for Fuelcell. So when it's just PVWatts, `gen` is 8760 but for Fuelcell it's 8760 * analysis_period.

It looks like in the original code, `accumulate_annual("gen", "annual_energy", ts_hour);` summed the entire 25 years into a single year, thinking it was subhourly data. The `accumulate_monthly` function makes a similar error, where if I give it 30 years of 15min data, it thinks the steps_per_hour is 120.

First issue seems to be that `accumulate_monthly` and `accumulate_annual` both expect single-year arrays.

Second issue is that the `ts_hour` input for the `scale` parameter is not being used correctly. From the rest of the code, it looks like `scale` is meant to scale between units rather than adjusting for different timesteps.

So it also looks like the edit you've made 
`accumulate_monthly_for_year("gen", "monthly_energy", ts_hour, step_per_hour);`

should be
`accumulate_monthly_for_year("gen", "monthly_energy", 1, step_per_hour, 1);`

Could you also take a look at whether the `ts_hour` parameter into `accumulate_monthly` makes sense?

brtietz:
I copied this code from pvsamv1. The reason we need to scale is that the gen units are in kW, so ts_hour changes them from kW to kWh accurately.

dguittet:
Oh that's right! Thanks for resolving that.

Are there open questions still?

brtietz:
No, I think this is ready to merge.

PR #382, opened 2020-06-02
Excess gen credit options

brtietz:
Implements two out of three enhancements in https://github.com/NREL/SAM/issues/16. Adds two new variables to cmod_utilityrate5:

- ur_nm_credit_month: Allows the user to change the month for net metering "true-up" to any month of the year. kWh or $ credits are carried over from year to year as needed
- ur_nm_credit_rollover: Allows the user to apply their net metering cash out credits to the following month, for example receiving a credit on the January bill instead of a check in December.

Note that annual minimums are still calculated at the end of December. This matches the assumption for some utilities where that is assessed 12 months after the PV interconnection (PG&E).  If we find an example of a utility with an annual minimum applied to a month besides Dec, or allow SAM to start simulation on a date other than Jan 1st, this will need to change.

Paired with SAM pull request https://github.com/NREL/SAM/pull/324

Currently contains code from https://github.com/NREL/ssc/pull/377 as well. It might be easier to review that one first. This contains additional tests for the new configurations.


cpaulgilman:
Hi Brian,

This looks good from the user interface perspective. I did some tests changing the month for credit rollover, and tried the new "Apply credits to future bills" options, and things are working and generating results that at least make intuitive sense. A couple of thoughts:

1. It would be helpful to have a new output for the "end of year" or "end of metering period" dollar amount. I would expect this to be a monthly array with zeros for all months except those where the dollar amount is paid.

2. For the "apply credits to future bills" option, it appears that the full credit amount is applied as a payment to the customer at the end of the month following the month for credit rollover. Is that the intended behavior? If the credit amount is larger than that month's total bill, should it be rolled over to subsequent months until the credit amount is less than the bill? In other words, the customer's would pay $0 until the credit runs out, instead of receiving a payment in the first month after the end of the period.

I approve the pull request -- these questions could be addressed in modifications to the code after this is merged into Develop.

I also changed the milestone to SAM 2021 -- That's our code name for the Fall 2020 (DOE FY2021) version.

Thanks,
Paul.

brtietz:
Thanks for the review!

For (1) how is this new output different from "Excess generation credit $ credit applied"?

Ah, for (2) I was going off of "You can leave the credit on your bill to apply toward future energy charges or contact PG&E to request a check.” from https://www.pge.com/en_US/small-medium-business/energy-alternatives/private-solar/understand-your-solar-bill/net-surplus-compensation.page Your interpretation of the bill not going below 0 makes sense. I'll try to get a sense of how long that change will take, since I feel like it's important to the battery dispatch project.

cpaulgilman:
I think "Excess generation credit $ credit applied" are the monthly rollover credits for the "Net energy metering with $ credits." For the "Net energy metering" option, the credits are in kWh, so it is zero for all months unless there is an end-of-metering-period credit. However, for the Net metering with $ credits option, it would store both the monthly credit and the end-of-period value, so would make it harder for users to check that the value is being calculated as they expect. It's a small detail, but might be nice to have that value reported separately.

I think it is fine for SAM to model the "request a check" option for now. I'm sure users will request the other option, but we can implement that later.

Thanks,
Paul.

Issue #384, opened 2020-06-04
pvwattsv7 10x slower than pvwattsv5


dguittet:
Thanks to @grantbuster for noticing that:

> For the pysam default runs, pvwattsv5 takes 70ms, pvwattsv7 takes 900 ms, and windpower takes 210 ms. This is just a small test on my local ubuntu instance.

I tested this on mac and found a 6.6x slowdown for v7. 

Need to see if it's related to the newer features in v7. Or this may be similar to a previous issue with significant slow down on linux with the windpower module which was resolved by moving a repeated mathematical operation out of the loop (which was not automatically optimized in linux).

janinefreeman:
@dguittet Was this also merged into develop? If so, I think we can close this issue! Thanks!

PR #386, opened 2020-06-05
Pvwattsv7 speed up shading

dguittet:
Pvwattsv7 slowdown from #384 is from the self-shading since turning it off by using fixed roof tracking results in speeds similar to v5.

Refactoring the self-shading code for memoization, I got a speed up for a single pvwattsv7 run with shading from 1.24902 to 0.249561 seconds with fixed tilt. This is still slower than v5 at 0.197782 s. For 1-axis tracking, v5 runs in 0.418802 s  and v7 in 0.458495 s.


shared/lib_pvshade.cpp
```diff
@@ -299,6 +278,50 @@ void selfshade_xs_horstr(bool landscape,
...
+
+double ssSkyDiff::compute(double tilt) {
+    if (gcr == 0)
+        throw std::runtime_error("ssSkyDiff::compute error: gcr required in initialization");
+    // sky diffuse reduction
+    double step = 1.0 / 1000.0;
+    double skydiff = 0.0;
+    double tand_stilt = tand(tilt);
+    double sind_stilt = sind(tilt);
+    double Asky = M_PI + M_PI / pow((1 + pow(tand_stilt, 2)), 0.5);
```
janinefreeman, line 297: looks like original function for Asky required tan of stilt in radians, not degrees?

shared/lib_pvshade.cpp
```diff
@@ -342,8 +365,9 @@ bool ss_exec(
 	double albedo,		// used to calculate reduced relected irradiance
 	bool trackmode,		// 0 for fixed tilt, 1 for one-axis tracking
 	bool linear,		// 0 for non-linear shading (C. Deline's full algorithm), 1 to stop at linear shading
-	double shade_frac_1x,	// geometric calculation of the fraction of one-axis row that is shaded (0-1), not used if fixed tilt 
+	double shade_frac_1x,	// geometric calculation of the fraction of one-axis row that is shaded (0-1), not used if fixed tilt
 
+	ssSkyDiff &skydiffs,
```
janinefreeman, line 370: please include detailed comment of what this is, so folks using the function know what they're expected to pass in and how to initialize it

shared/lib_pvshade.h
```diff
@@ -67,7 +67,7 @@ void diffuse_reduce(
 //	double phi0, // mask angle
 	double alb,
 	double nrows,
-
+	double skydiff0,
```
janinefreeman, line 70: Include detailed comment about what this is. See other comment about confusing variable name.

shared/lib_pvshade.h
```diff
@@ -140,8 +156,8 @@ bool ss_exec(
 	double albedo,		// used to calculate reduced relected irradiance
 	bool trackmode,		// 0 for fixed tilt, 1 for one-axis tracking
 	bool linear,		// 0 for non-linear shading (C. Deline's full algorithm), 1 to stop at linear shading
-	double shade_frac_1x,	// geometric calculation of the fraction of one-axis row that is shaded (0-1), not used if fixed tilt 
-	
-	ssoutputs &outputs);
+	double shade_frac_1x,	// geometric calculation of the fraction of one-axis row that is shaded (0-1), not used if fixed tilt
+    ssSkyDiff &skydiffs,
```
janinefreeman, line 160: same comment as above- include detailed comment and consider less confusing variable name

ssc/cmod_pvsamv1.cpp
```diff
@@ -984,6 +984,7 @@ void cm_pvsamv1::exec( ) throw (general_error)
 		else
 			b = Subarrays[nn]->selfShadingInputs.nmody * Subarrays[nn]->selfShadingInputs.width;
 		Subarrays[nn]->selfShadingInputs.row_space = b / Subarrays[nn]->groundCoverageRatio;
+		Subarrays[nn]->selfShadingSkyDiff.init(Subarrays[nn]->tiltDegrees, Subarrays[nn]->groundCoverageRatio);
```
janinefreeman, line 987: Curious if this could be included as part of a different function so users from other parts of the code don't have to remember to initialize it. Could it be part of the constructor?

ssc/cmod_pvsamv1.cpp
```diff
@@ -1454,7 +1455,10 @@ void cm_pvsamv1::exec( ) throw (general_error)
 						if (radmode == irrad::DN_DF || radmode == irrad::GH_DF) dhi_to_use = (ssc_number_t)wf.df;
 						else dhi_to_use = Irradiance->p_IrradianceCalculated[1][hour * step_per_hour]; // top of hour in first year
 
-						if (ss_exec(Subarrays[nn]->selfShadingInputs, stilt, sazi, solzen, solazi, beam_to_use, dhi_to_use, ibeam, iskydiff, ignddiff, alb, trackbool, linear, shad1xf, Subarrays[nn]->selfShadingOutputs))
+						if (ss_exec(Subarrays[nn]->selfShadingInputs,
+						        stilt, sazi, solzen, solazi, beam_to_use, dhi_to_use, ibeam, iskydiff, ignddiff, alb, trackbool, linear, shad1xf,
+						        Subarrays[nn]->selfShadingSkyDiff,
```
janinefreeman, line 1460: If this is part of the constructor, this function wouldn't have to require this variable.

ssc/cmod_pvwattsv7.cpp
```diff
@@ -557,6 +557,7 @@ class cm_pvwattsv7 : public compute_module
 			//   If 2 module per Y, then nmodx=nrows/2.
 			pv.nmodx = pv.nrows / pv.nmody;
 			pv.row_spacing = module.length * pv.nmody / pv.gcr;
+			ssval.init(pv.tilt, pv.gcr);
```
janinefreeman, line 560: If I'm understanding this function correctly, this could be sped up even further with an additional argument for fixed tilt systems- where there is no sweep of 1000 tilt values and the sky diffuse reduction is only calculated once. This would apply both to pvwattsv7 and pvsamv1.

shared/lib_pv_io_manager.h
```diff
@@ -418,26 +418,27 @@ struct Subarray_IO
 	double nameplateLossPercent;
 	double dcLossTotalPercent;			/// The DC loss due to mismatch, diodes, wiring, tracking, optimizers [%]
 
-	// Shading and snow	
+	// Shading and snow
 	flag enableSelfShadingOutputs;		// Choose whether additional self-shading outputs are displayed
 	int shadeMode;						// The shading mode of the subarray [0 = none, 1 = standard (non-linear), 2 = thin film (linear)]
 	flag usePOAFromWeatherFile;			// Flag for whether or not a shading model has been selected that means POA can't be used directly for that subarray
 	ssinputs selfShadingInputs;			// Inputs and calculation methods for self-shading of the subarray
+	ssSkyDiff selfShadingSkyDiff;         // Intermediate variables for self-shading
```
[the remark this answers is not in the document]
dguittet: Thanks Janine for the review!

I've changed the class name to `sssky_diffuse_table` and added comments to the class: https://github.com/NREL/ssc/pull/386/commits/fc8f8ebd7fd97fac1f14b771c6acc9f9068238e4#diff-f9e6ae6261e2f5713e598c91de6a8d4a

This class contains a lookup table that's dynamically populated (new values are only added when they are required) and so its memory needs to outlive the `ssexec` function and be as long as the simulation duration.

shared/lib_pvshade.cpp
```diff
@@ -159,21 +158,21 @@ double qromb(double (*func)(double,double,double,double), double a, double b, do
...
+    double Gdh,
+    double poa_sky,
+    double poa_gnd,
+    double gcr,
 //	double phi0, // mask angle
 	double alb,
-	double nrows,
-
+    double nrows,
+    double skydiff0,
```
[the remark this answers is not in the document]
dguittet: I agree. I've changed the function signature to use the lookup table itself so it's more obvious what this input is for. Comments in the class.

shared/lib_pvshade.cpp
```diff
@@ -299,6 +278,50 @@ void selfshade_xs_horstr(bool landscape,
 	}
 }
 
+double ssSkyDiff::lookup(double tilt) {
+    char buf[124];
```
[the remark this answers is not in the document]
dguittet: Added comment. The table is defined once the class is initialized. This class does need to be initialized however. But if it hasn't been initialized then the `runtime_error` will be thrown.

shared/lib_pvshade.cpp
```diff
@@ -299,6 +278,50 @@ void selfshade_xs_horstr(bool landscape,
...
+    sprintf(buf, "%.3f", tilt);
+    if (SkyDiffTable.find(buf) != SkyDiffTable.end())
+        return SkyDiffTable[buf];
+    return compute(tilt);
+}
+
+double ssSkyDiff::compute(double tilt) {
```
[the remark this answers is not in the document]
dguittet: Ah that's what `stilt` stands for! :) Thanks

shared/lib_pvshade.cpp
dguittet: Original version used tan(stilt * DTOR)
```
Asky = M_PI + M_PI / pow((1 + pow(tan(stilt * DTOR), 2)), 0.5)
```
whereas now it uses 
```
#define tand(x) tan( DTOR*(x) )
...
tand_stilt = tand(surface_tilt)
```

The reason I reorganized this code is because I initially started by trying to speed up the simulation by vectorizing the calculations (hence all the 1000-length arrays) and this resulted in some speed up, but not enough to make a dent. So then I implemented the whole look-up table, and decided to leave this as is.

ssc/cmod_pvsamv1.cpp
dguittet: Very good point! Initialized in Subarray_IO::Subarray_IO

ssc/cmod_pvsamv1.cpp
dguittet: Unfortunately ss_exec isn't a class

ssc/cmod_pvwattsv7.cpp
dguittet: I think for fixed tilt, the sky diffuse derate needs to be calculated at least once. A single calculation of a sky diffuse derate value requires the sweep of 1000 tilt values.

Btw, I also added an `en_self_shading` flag to help make the initialization of sssky_diffuse_table more clear.

dguittet:
Thanks so much for the thorough and helpful review, Janine. I think it looks in a much cleaner state!

The PV tests are passing on my machine. I'll wait for the Travis CI Check to finish my most recent commit.

dguittet:
Looks like it passed! Do you have any comments to add on?

janinefreeman:
Looks great to me! Thanks for the changes, feel free to merge!

Issue #385, opened 2020-06-04
Pvwattsv5_1ts not working


dguittet:
`cmod_pvwattsv5_1ts` is not working since the models and variables in `cm_pvwattsv5_base` are not set up.

dguittet:
It looks like the Pvwattsv5 single time step model wasn’t correctly updated with some recent refactoring.

I fixed this by calling initializing functions from base pvwattsv5 class.

Regarding `poa` and `tcell`, a comment says ["/* input/output variable: tcell & poa from previous time must be given */"](https://github.com/NREL/ssc/blob/4ef0b094ca72b66440510c80a3dd6553145c9eb8/ssc/cmod_pvwattsv5.cpp#L629), but these values are not used in the [code](https://github.com/NREL/ssc/blob/4ef0b094ca72b66440510c80a3dd6553145c9eb8/ssc/cmod_pvwattsv5.cpp#L668).

It looks like they could be used as part of `initialize_cell_temp`, which [can use](https://github.com/NREL/ssc/blob/4ef0b094ca72b66440510c80a3dd6553145c9eb8/ssc/cmod_pvwattsv5.cpp#L216) poa and tcell from the last time step.

Question for @janinefreeman, should poa and tcell of the last time step be required/used? If not, should they not be "required inputs", since users may not have a poa and tcell to start?

PR #387, opened 2020-06-05
Pvwattsv5 1ts

dguittet:
Copying comment over from #385 :

It looks like the Pvwattsv5 single time step model wasn’t correctly updated with some recent refactoring.

I fixed this by calling initializing functions from base pvwattsv5 class in the pvwattsv5_1ts: 4ef0b09

Regarding poa and tcell, a comment says 
>/* input/output variable: tcell & poa from previous time must be given */

but these values are not used in the code.

It looks like they could be used as part of `initialize_cell_temp`, which can use `poa` and `tcell` from the last time step.

Question for @janinefreeman, should `poa` and `tcell` of the last time step be required/used? If not, should they not be "required inputs", since users may not have these values to start?

### DIFF
--- a/ssc/cmod_pvwattsv5.cpp
+++ b/ssc/cmod_pvwattsv5.cpp
@@ -627,8 +627,8 @@ DEFINE_MODULE_ENTRY(pvwattsv5, "PVWatts V5 - integrated hourly weather reader an
 
 static var_info _cm_vtab_pvwattsv5_1ts_outputs[] = {
 	/* input/output variable: tcell & poa from previous time must be given */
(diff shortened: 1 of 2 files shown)


janinefreeman:
I think it makes sense for `poa` and `tcell` of the previous timestep to be used if they are provided, but not required. Using them if they are provided can also go on a future to-do list item, since we want to switch to pvwattsv7 anyways...

dguittet:
Got it, thanks. I've allowed tcell & poa from the last time step to be used in `initialize_cell_temp`.

PR #388, opened 2020-06-10
Don't initialise doubles with NULL in C++

bje-:
In fact, don't initialise doubles with NULL ever. :smile:

Besides being less clear to the reader, G++ emits the following warning:
```
   warning: converting to non-pointer type ‘double’ from NULL
   [-Wconversion-null]
```


ssc/cmod_csp_tower_eqns.cpp
```diff
@@ -303,7 +303,7 @@ void Tower_SolarPilot_Capital_Costs_MSPT_Equations(ssc_data_t data)
     ssc_data_set_number(data, "csp_pt_cost_storage_mwht", csp_pt_cost_storage_mwht);
 
     ssc_data_get_number(data, "p_ref", &p_ref);
-    demand_var = NULL;
+    demand_var = 0;
```
Matthew-Boyd, line 306: Can you assign this one and the others to:
  std::numeric_limits<double>::quiet_NaN();
instead of 0?

Matthew-Boyd: It looks like you merged develop into this branch to resolve the merge conflict, but can you still change all of the NULLs to quiet_NaNs instead of 0s?

bje-: Whoops! Fixed.

PR #389, opened 2020-06-10
Re-indent var_table::lookup to eliminate G++ misleading-indentation warning

bje-:
The code is correct, but the indenation makes it misleading (and G++ agrees!)

```
vartab.cpp: In member function ‘var_data* var_table::lookup(const string&)’:
vartab.cpp:322:5: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]
  322 |     if (it == m_hash.end())
      |     ^~
vartab.cpp:324:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
  324 |  if ( it != m_hash.end() )
      |  ^~
```

### DIFF
--- a/ssc/vartab.cpp
+++ b/ssc/vartab.cpp
@@ -318,13 +318,13 @@ bool var_table::rename_match_case( const std::string &oldname, const std::string
 
 var_data *var_table::lookup( const std::string &name )
 {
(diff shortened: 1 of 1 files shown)


dguittet:
Thanks! We have an old item for code reformatting to fix these issues that we'll probably revisit soon!

PR #397, opened 2020-06-11
Fix residual G++ warnings

bje-:
This change silences a couple of residual warnings due to some variables being set but never used, including `percent_complete` because the code that accesses this variable is commented out. I figured the best thing to do was to comment out the variable declaration and assignment as well. Perhaps the right fix is to just delete the commented out regions all together?


ssc/cmod_fuelcell.cpp
```diff
@@ -137,13 +137,15 @@ void cm_fuelcell::exec()
 {
 	double annual_energy = 0.0;
 	double annual_fuel = 0.0;
-	float percent_complete = 0.0;
+	// float percent_complete = 0.0;
```
janinefreeman, line 140: Lines 34 and 211 in this file should also be commented out if this is commented out. Could you please add those?

ssc/core.cpp
```diff
@@ -125,6 +125,7 @@ bool compute_module::evaluate()
                 return false;
             }
         }
+	return true;
```
janinefreeman, line 128: I do not think we want this line there- we want this function not to return anything if it doesn't return false. @sjanzou can you confirm?

sjanzou: @janinefreeman, the "return true;" actually should be there since the CallSscEquations is a Lambda function with an implicit boolean type that should return a boolean from all control paths. However, this will not make a difference in the results since the return value is not used anywhere... lines 130 and 144.

Currently, the function returns 1>C:\Projects\GitHub\NREL\develop\ssc\ssc\core.cpp(128): warning C4715: '<lambda_c23638134389f859b3dd7a343978d202>::operator()': not all control paths return a value

janinefreeman: Ok, thank you for clarifying @sjanzou !

PR #399, opened 2020-06-11
Silience G++ sequence point warnings

bje-:
Clean up initialisation of a number of class member variables to quiet NaN. Use C++ default member intialisers instead of chained assignment expressions which is a lot clearer to read, I think you'll agree. It also has the advantage of putting the initialisation right next to the member variable declaration.


bje-:
The original code has the constructors use a chain of assignments expressions to set each member variable to `std::numeric_limits<double>:::quiet_NaN()` in the more compact form:
```
var1 = var2 = var3 = var4 = var5 = var6 = std::numeric_limits<double>::quiet_Nan();
```

However, because the variables have long names and there are so many of them, duplicates have crept in. eg:
```
var1 = var2 = var3 = var1 = var4 = var5 = var6 = std::numeric_limits<double>::quiet_NaN();
```
This will produce a sequence point warning because one of the variables (in this case, `var1`) is repeated in the expression. While it would be easy to fix this by removing any duplicates, I think that initialising each variable in the structure declaration is clearer and more maintainable.

tcs/csp_solver_core.h
```diff
@@ -388,34 +388,29 @@ class C_csp_collector_receiver
...
-		S_csp_cr_inputs()
-		{
-			m_field_control = std::numeric_limits<double>::quiet_NaN();
-			m_adjust = std::numeric_limits<double>::quiet_NaN();
-
-			m_input_operation_mode = -1;
-		}
+		double m_field_control =		//[-] Defocus signal from controller (can PC and TES accept all receiver output?)
+		  std::numeric_limits<double>::quiet_NaN();
+	  	int m_input_operation_mode = 1;		//[-]
```
Matthew-Boyd, line 411: Please change to -1.

bje-: All of `double` class members previously assigned to quiet_NaN()?

bje-: Done.

Matthew-Boyd: No, sorry, I should have been more clear. The doubles were good as quiet_NaN's, you just missed the negative sign with the m_input_operation_mode as it was originally set to -1.

bje-: Oh, gotcha.

bje-: Sorry for the mixup. :smile:

Matthew-Boyd:
No problem. Thanks for contributing.

PR #401, opened 2020-06-16
ssc issue 400 transformer timestep loss units updated.

sjanzou:
The transformer timestep units were previously in kWh and are now in kW


sjanzou:
This will be merged into patch and develop after review.

PR #406, opened 2020-06-19
Fix dc_adjust issue SAM_335

cpaulgilman:
https://github.com/NREL/SAM/issues/335

In common.cpp, make dc_adjust:constant required, and :hourly and :periods optional. This is consistent with the adjustment_factors setup() function that requires :constant.

Remove dc_adjust from generic_system. SAM UI does not assign a value to dc_adjust for the generic system configurations. Maybe this was added for some internal testing of DC losses for a generic system / battery configuration.

Change pvsamv1 label for adjust to "AC adjustment".

### DIFF
--- a/ssc/cmod_generic_system.cpp
+++ b/ssc/cmod_generic_system.cpp
@@ -69,7 +69,6 @@ public:
 		add_var_info( _cm_vtab_generic_system );
 
 		// performance adjustment factors
-		add_var_info(vtab_dc_adjustment_factors);
 		add_var_info(vtab_adjustment_factors);
 		add_var_info(vtab_technology_outputs);
 	}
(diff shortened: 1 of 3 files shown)


ssc/cmod_generic_system.cpp
```diff
@@ -69,7 +69,6 @@ class cm_generic_system : public compute_module
 		add_var_info( _cm_vtab_generic_system );
 
 		// performance adjustment factors
-		add_var_info(vtab_dc_adjustment_factors);
```
sjanzou, line 72: Thanks for removing this - that was a great question on DC adjustments for the generic system - maybe DC generators?

cpaulgilman:
I did run defaults locally for all generic, generic battery, detailed PV, and PV battery configurations to make sure they generate results.

sjanzou:
Restore branch dc_adjust_SAM_335 and merge into develop and commit and push to fix Travis failing tests.

Issue #404, opened 2020-06-17
Vanadium flow battery model discharges more than it charges


dguittet:
from the issue on the SAM forum: https://sam.nrel.gov/forum/forum-general/3018, a .sam file [VRF_2020-06-16.sam.zip](https://github.com/NREL/ssc/files/4793989/VRF_2020-06-16.sam.zip) showing a Vanadium flow battery with manual dispatch is able to discharge more energy than it has charged:

![image](https://user-images.githubusercontent.com/29387736/84925404-105faa80-b087-11ea-9853-2fca0cd6aa18.png)

The issue seems to be that for a unit change in SOC, there is a larger energy kWh change for discharging than charging, so that it results in an efficiency > 100.

![image](https://user-images.githubusercontent.com/29387736/84928115-1fe0f280-b08b-11ea-9279-3497ecc5c7ff.png)

dguittet:
According to the [paper](https://ieeexplore.ieee.org/document/6971196) representing the implemented model, the open-circuit voltage and the voltage drop forced by the aggregated resistance (R_specific) of the battery stack is:

![image](https://user-images.githubusercontent.com/29387736/84958260-410d0780-b0ba-11ea-9f85-d7d8054b3723.png)

where the convention is that voltage and current are both negative during charging. So during charging, `V_stack < 0`, `I_stack < 0` and `R_specific > 0`, and the two terms accumulate in absolute value.

Current SAM code has the two terms with opposite directions when charging since the voltages always > 0 but current < 0 when charging:
```
_V_ref_50 + m_RCF * T * A + I_string * _R
```

This resulted in a different voltage versus SOC curve for charging and discharging:
![image](https://user-images.githubusercontent.com/29387736/84960264-6c91f100-b0be-11ea-990b-60dcde5451b4.png)

This resulted in a higher dE/dSOC magnitude for discharging than charging, and therefore led to incorrectly high battery efficiencies.

The original `voltage_vanadium_redox_t` code shows a negative for I * R: https://github.com/NREL/ssc/blob/2ed1ddbaf684346bf7ba8577c7e043a114b4216e/shared/lib_battery.cpp#L730

This seems like it could create the opposite effect and result in lower battery efficiencies. 

However I think this voltage model only intends to calculate the battery system's OC voltage and does not model overpotential. In which case, the OC V versus SOC curves should be the [same](Electrochemical) whether charging or discharging:
[...]

PR #407, opened 2020-06-19
Vanadium flow battery: voltage model fixes

dguittet:
Fixes:
 - Copied over some fixes from `develop` on `calculate_max_charge_w` and `calculate_max_discharge_w`
 - fix `-inf` potential from `voltage_model`
 - voltage drop across aggregated stack the same for charge or discharge
 - adapt test to show catch difference in using abs(I_string)

With the thermal model turned off: this is the yearly battery charging & discharging table:
image

### DIFF
--- a/shared/lib_battery.cpp
+++ b/shared/lib_battery.cpp
@@ -864,39 +864,41 @@ void voltage_vanadium_redox_t::updateVoltage(capacity_t * capacity, thermal_t * 
 
 double voltage_vanadium_redox_t::calculate_max_charge_w(double q, double qmax, double kelvin, double *max_current) {
     qmax /= _num_strings;
-    q = qmax - tolerance;
-    double max_I = qmax - q;
+    q /= _num_strings;
+    double max_I = (q - qmax) / dt_hr;
 
     if (max_current)
         *max_current = -max_I;
 
-    return -voltage_model(q, qmax, max_I, kelvin) * max_I * _num_strings * _num_cells_series;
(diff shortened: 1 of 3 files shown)


shared/lib_battery.cpp
```diff
@@ -925,22 +928,25 @@ double voltage_vanadium_redox_t::voltage_model(double q0, double qmax, double I_
 	double SOC_use = q0 / qmax;
 	if (SOC_use > 1. - tolerance)
 		SOC_use = 1. - tolerance;
+	if (SOC_use == 0)
+	    SOC_use = 1e-3;
 
 	double A = std::log(std::pow(SOC_use, 2) / std::pow(1 - SOC_use, 2));
-
-	return _V_ref_50 + m_RCF * T * A + I_string * _R;
+	return _V_ref_50 + m_RCF * T * A + fabs(I_string) * _R;
```
brtietz, line 935: Can you add a comment here about how our voltage assumptions are different from the paper? (justifying the fabs call)

dguittet: Sure, good point!

dguittet: // In constrast to the V_stack + I_stack * R_specific in the paper which follows the convention of negative voltages,
// here the abs(I_stack) is used to allow both terms to move in same direction (https://github.com/NREL/ssc/issues/404)

dguittet:
@cpaulgilman Yes, it would be a separate PR. It think it's easier to review and track changes when the Pull requests (and Issues) are smaller and more targeted.

I haven't gotten a chance to look at the Li-ion part, but I'll move it to an issue of its own about the voltage table.

PR #409, opened 2020-06-19
Hour ranges in weather file / data

dguittet:
The subhourly weather file [here](https://raw.githubusercontent.com/NREL/ssc/develop/test/input_cases/pvsamv1_data/LosAngeles_WeatherFile_15min.csv) used in `CMPvsamv1PowerIntegration_cmod_pvsamv1.NoFinancialModelCustomWeatherFile` counts the hours with (1-24) instead of (0-23).

If weather_record.hour is being used to calculate index, as in this code: 

https://github.com/NREL/ssc/blob/3c8195d35cbf75372ab3f77653fcd2a47be9c57a/ssc/cmod_pvsamv1.cpp#L1131

then, hour ranges should always be (0-23)

I assumed the weather file starts "recording" at the same time, but perhaps just has different indexing, so I subtracted `1` from each hour. This ended up shifting the interpretation of the data and resulted in the test changes shown in the commit.

Should these types of files be handled differently?

### DIFF
--- a/shared/lib_util.cpp
+++ b/shared/lib_util.cpp
@@ -905,7 +905,6 @@ size_t util::hour_of_year(size_t month, size_t day, size_t hour)
 		h += days_in_months[m] * 24;
 	h += (day - 1) * 24;
 	h += hour;
-	if (h > 8759) h = 8759;
 	return h;
 }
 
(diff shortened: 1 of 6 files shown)


janinefreeman:
Good catch! I think you are right about the correct behavior. We always expect SAM to be given the first hour of the year representing midnight-1am on Jan 1, so whether it's timestamped 0-23 or 1-24, it should represent the same thing. Thanks!

Issue #411, opened 2020-06-22
Incorperate load escalation into battery dispatch forecasts


brtietz:
Right now if load is specified as a set of values for year 1, these values are copied for years 2 - 25 for the battery forecast. This does not match what is done in the financial models, where load can be escalated on a percentage basis each year. In order to improve the accuracy of dispatch forecasting, including load escalation in cmod_pvsamv1 and cmod_battery.

PR #413, opened 2020-06-23
Unify Load Escalation Code

brtietz:
Prior to this PR, the utility rate code was the only module using the load_escalation variable. This could cause unintended effects in battery dispatch and grid curtailment. Incorporate load escalation into code that copies a single year of load inputs into a set of lifetime variables, and propagate throughout the codebase.

Notes on the effects on battery dispatch, for behind the meter, residential:

* Counter to my intuition, NPV goes down if the battery can only charge from the system (the default). I think what is happening is the peak shaving dispatch doesn't charge at some timesteps where it did before, since it now realizes that the PV isn't enough to cover the load.
* NPV goes up for a battery that can charge from the grid, so that makes sense.
* The battery typically lasts another year or two in both cases

Solves https://github.com/NREL/ssc/issues/411


ssc/cmod_pvsamv1.cpp
```diff
@@ -1128,7 +1132,7 @@ void cm_pvsamv1::exec( ) throw (general_error)
 			// if PV simulation is subhourly.  load is assumed constant over the hour.
 			// if no load profile supplied, load = 0
 			if (nload == 8760)
-				cur_load = p_load_in[hour];
+				cur_load = p_load_in[hour] * load_scale[iyear];
```
dguittet, line 1135: It looks like `cur_load` is being calculated with scaling and then stored in the lifetime vector `p_load_full`. Could  `single_year_to_lifetime_interpolated` be used from the start like in other cmods?

ssc/cmod_grid.cpp
```diff
@@ -85,6 +85,41 @@ void cm_grid::exec() throw (general_error)
...
+    single_year_to_lifetime_interpolated<double>(
+        system_use_lifetime_output,
+        analysis_period,
+        n_rec_lifetime,
+        load_year_one,
+        load_scale,
+        gridVars->loadLifetime_kW,
+        n_rec_single_year,
+        gridVars->dt_hour_gen);
+
```
dguittet, line 122: Why is the calculation of curtailment initialization but the load is in exec?

brtietz: Both used to be in initialization. I was favoring "move as little code as possible" but if you prefer to be consistent between the two I can move curtailment as well.

ssc/common.cpp
```diff
@@ -1463,3 +1464,28 @@ bool ssc_cmod_update(std::string &log_msg, std::string &progress_msg, void *data
...
+{
+    vt = v;
+}
+
+std::vector<double> scalefactors::get_factors(const char* name)
+{
+    size_t nyears = vt->as_integer("analysis_period");
+    size_t count, i;
+    std::vector<double> scale_factors(nyears); // TODO analysis period
+    ssc_number_t* parr = vt->as_array(name, &count);
```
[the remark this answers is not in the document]
brtietz: as_array would throw a general_error, which would show up in the UI.

Throwing an error seems appropriate to me, the defaults should be defining the factors that are used.

ssc/cmod_pvsamv1.cpp
brtietz: Thanks for the suggestion! Looks like cur_load was used in some commented out logging functions. I removed it and replaced it with something that matches the other cmods in the latest commit.

ssc/cmod_grid.cpp
dguittet: I do prefer consistency :) I think they belong together since they're setting up member variables.

ssc/cmod_grid.h
```diff
@@ -103,6 +93,7 @@ struct gridVariables
 			(size_t)analysis_period,
 			n_rec_lifetime,
 			curtailment_year_one,
+            scaleFactors,
```
sjanzou: This line has broken running PVWatts Residential and Commercial configurations in SAM since no load_scale passed to single_year_to_lifetime_interpolated but a load passed and causes access violations in lines 86 and 93 in lib_time.cpp

PR #414, opened 2020-06-24
Ssc 410 all ac power

brtietz:
Changes the PV forecast for DC connected batteries to the AC output of the PV system. Change the variable names so which variables are in AC power vs DC power is more consistent throughout the code. The DC charging power conversion in the btm dispatch algorithm has been updated to reflect this.

This allows the peak shaving algorithm to more accurately account for the inverter efficiency, leading to an increase of NPV in the SAM default case of 15% (we avoid a battery replacement).

Right now, clipped power is still DC, I think this works itself out in the constraints algorithm, but if I need to make further updates, let me know.

Unit tests were updated to reflect the new dispatch.


brtietz:
I just realized that p_systemACPower is not initialized when it gets called in this code, leading to a vector of zeroes for forecasting. I'll look into an appropriate conversion factor for p_systemDCPower prior to merging this.

ssc/cmod_pvsamv1.cpp
```diff
@@ -2013,6 +2013,7 @@ void cm_pvsamv1::exec( ) throw (general_error)
 						//DC batteries not allowed with multiple MPPT, so can just use MPPT 1's voltage
 						sharedInverter->calculateACPower(dcpwr_kw, PVSystem->p_mpptVoltage[0][idx], 0.0);
 						cliploss = sharedInverter->powerClipLoss_kW;
+                        PVSystem->p_systemACPower[idx] = sharedInverter->powerAC_kW;
```
dguittet, line 2016: This code is only executed if there is not `p_pv_dc_clipping_forecast`. So if a clipping forecast is provided, what does the dispatch expect? Would a vector of zeroes make sense there or not?

test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -303,13 +303,13 @@ TEST_F(CMPvsamv1BatteryIntegration_cmod_pvsamv1, ResidentialDCBatteryModelIntegr
 	set_array(data, "load", load_profile_path, 8760); // Load is required for peak shaving controllers
 
 	ssc_number_t expectedEnergy[3] = { 8634, 8637, 8703 };
-	ssc_number_t expectedBatteryChargeEnergy[3] = { 838.3, 807.2, 528.65 };
-	ssc_number_t expectedBatteryDischargeEnergy[3] = { 764.4, 735.4, 484.5 };
+	ssc_number_t expectedBatteryChargeEnergy[3] = { 1412.75, 1414.89, 253.2 };
+	ssc_number_t expectedBatteryDischargeEnergy[3] = { 1283.8, 1285.88, 226.3 };
```
dguittet, line 307: Looks like the PV forecast made quite a difference all the pv + dc battery results! Could you please add a test seeing what happens if you take the actual dc clipped power loss and use that as the clipping prediction, to make sure that the results are the same regardless of if there is a dc clipped power forecast or not?

ssc/cmod_pvsamv1.cpp
brtietz: Oh, whoops, thanks for spotting this. A vector of zeroes does not make sense. It looks like some of the code for allowing for a custom PV forecast got stomped on (for example `p_pv_dc_forecast` was not passed into the battery), so I'll move things around a bit to re-enable that feature.

ssc/cmod_pvsamv1.cpp
```diff
@@ -1020,13 +1020,13 @@ void cm_pvsamv1::exec( ) throw (general_error)
...
-	std::vector<ssc_number_t> p_pv_dc_use;
+	std::vector<ssc_number_t> p_pv_ac_use;
 
 	if (is_assigned("batt_pv_clipping_forecast")) {
 		p_pv_clipping_forecast = as_vector_ssc_number_t("batt_pv_clipping_forecast");
 	}
-	if (is_assigned("batt_pv_dc_forecast")) {
-		p_pv_dc_forecast = as_vector_ssc_number_t("batt_pv_dc_forecast");
+	if (is_assigned("batt_pv_ac_forecast")) {
+		p_pv_dc_forecast = as_vector_ssc_number_t("batt_pv_ac_forecast");
```
dguittet, line 1029: Could you please refactor this to be named `p_pv_ac_forecast` to avoid confusion?

test/ssc_test/cmod_battery_pvsamv1_test.cpp
dguittet: Great, thanks!

PR #415, opened 2020-06-26
Allow output for 300W and under system capacities for pvwattsv7

sjanzou:
The pv.modules on line 537 was zero for all dc_nameplate values less than or equal to the module.stc_watts = 300. Also, the pv.modx calculation on line 559 was zero. These two conditions caused the ssout.shade_frac_fixed to be calculated as -inf (line 956) which in turn caused the ibeam to be nan on line 981. 

The fix was to check both pv.modules >=1 (line 539) and pv.nmodx>=1 (line 561).


janinefreeman:
Thanks for fixing this, Steve! Feel free to merge.

PR #416, opened 2020-06-29
Fix issues with crashing SAM PVWatts distributed and PV non-lifetime configs

sjanzou:
All of the PVWatts Distributed configurations in SAM were failing after latest cmod_grid updates. Also, the PV non-lifetime configurations were failing (LCOE Calculator and None). 
First issue was with access violation in single_year_to_lifetime_interpolated with implied assumption that scale factors array was input of size analysis period.
Second issue was that analysis period was only required in pvsamv1 for lifetime mode but was explicitly used in scalefactors class called unconditionally in pvsamv1 line 1103 and used in common.cpp line 1475.
This branch was used with the develop branch of SAM to successfully run all default configurations on Windows


ssc/cmod_grid.cpp
```diff
@@ -128,20 +128,22 @@ void cm_grid::exec() throw (general_error)
         load_year_one = as_vector_double("load");
     }
 
+
+    /* Does nothing since "load_escalation" not an input to this compute module
```
brtietz, line 132: I like this comment, but don't see the reason to leave the code commented out. Why not just remove it?

ssc/cmod_grid.cpp
```diff
@@ -128,20 +128,22 @@ void cm_grid::exec() throw (general_error)
...
     }
-
+    */
     single_year_to_lifetime_interpolated<double>(
         system_use_lifetime_output,
         analysis_period,
         n_rec_lifetime,
         load_year_one,
-        load_scale,
+        scaleFactors,
```
dguittet, line 146: The escalation of the load in the pv and battery models was added for proper lifetime battery dispatch signals-- as the load goes up over time, the battery must also try to meet that load to capture the financial benefits. This would result in higher `gen` output over the lifetime since `gen` is prior to `load` subtraction. This is now applied for both BTM and FOM.

For proper financial calculations where `cmod_grid` is included, the interconnect and curtailment limitations are applied to `gridNet = gen - gridVars->loadLifetime_kW[i]` in line 162/4. I think both `gen` and `loadLifetime_kW` were not escalated. 

But now if it were a battery simulation, `gen` is escalated since it was dispatched with escalated loads, so this scaleFactor should not just be `std::vector<double> scaleFactors(analysis_period, 1.0);`. However, if it were not a battery simulation, then `gen` will not be escalated and neither should the load. This means the value of scaleFactors should depend on whether or not `load_escalation` is available and makes sense.... Or am I missing something here?

shared/lib_time.cpp
```diff
@@ -83,14 +83,14 @@ void single_year_to_lifetime_interpolated(
 		// Scale single year interpolated vector to lifetime
 		for (size_t y = 0; y < n_years; y++) {
 			for (size_t i = 0; i < n_rec_single_year; i++) {
-				lifetime_from_singleyear_vector.push_back(singleyear_sampled[i] * scale_factor[y]);
+				lifetime_from_singleyear_vector.push_back(singleyear_sampled[i] * (scale_factor.size() > y ?  scale_factor[y] : 1.0));
```
dguittet, line 86: I think this work-around of a too-small scale_factor vector is a step in the right direction of flexibility, but I'd prefer more explicit handling of what's expected from this new scale_factor parameter as this seems to potentially hide errors.

@brtietz What are expected lengths of scale_factor? Same length of `singleyear_vector`, length 0 and length 1 seem reasonable to me.

ssc/cmod_grid.cpp
brtietz: The escalated load parameters apply to all pvsamv1, cmod battery, battwatts, and solar water heater simulations, as well as the utility rate modules. So I think you're correct that `load_scale` should apply here when we're doing lifetime simulations. `gen` should already be lifetime length, and it would account for PV degradation in most cases.

The question then, is what to do when the analysis period is 1 year (and, as a result, is undefined), or we're doing AC side degradation in the financial models (like for PVWatts)? It's worth noting that the GUI has load escalation defined, so it might be worth requiring some additional variables in the vartable/in the SAM defaults.

shared/lib_time.cpp
brtietz: The original intent was `n_years`, which seems to have caused problems with PVWatts. scale_factors translates from the SSC inputs (which can be length 1 or length `n_years`). I'd prefer to avoid duplicating the code that's already in common.cpp, if possible.

brtietz:
Since Darice reminded me that we might want load escalation in cmod_grid, I created an alternative PR here: https://github.com/NREL/ssc/pull/417 I think which we choose depends on how load_escalation and cmod_grid should interact.

PR #417, opened 2020-06-29
Resolve PVWatts issues by making scalefactors more flexible

brtietz:
An alternative solution to the issue in #416. Allows three options for load_escalation and other scaling factors:

* Length zero: no scaling (multiply by 1.0)
* Length 1: scale at a constant percent rate each year
* Length n_years/analysis_period: scale based on the custom vector

Defaults to analysis_period of 1 year if it is not defined, preventing the access errors.

### DIFF
--- a/ssc/common.cpp
+++ b/ssc/common.cpp
@@ -1470,6 +1470,7 @@ scalefactors::scalefactors(var_table* v)
     vt = v;
 }
 
+// TODO: add a boolean for factoring in inflation when adding new financial models
 std::vector<double> scalefactors::get_factors(const char* name)
 {
     size_t nyears = 1;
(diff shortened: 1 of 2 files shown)


ssc/common.cpp
```diff
@@ -1472,15 +1472,24 @@ scalefactors::scalefactors(var_table* v)
...
     size_t count, i;
-    std::vector<double> scale_factors(nyears); // TODO analysis period
+    std::vector<double> scale_factors(nyears);
     ssc_number_t* parr = vt->as_array(name, &count);
     if (count == 1)
     {
         for (i = 0; i < nyears; i++)
             scale_factors[i] = (ssc_number_t)pow((double)(1 + parr[0] * 0.01), (double)i);
     }
+    else if (count == 0)
```
dguittet, line 1488: Okay, I see where the length of the array is coming from. So actually I think this 0 count will never happen since there can't be an SSC_ARRAY assigned that has length < 1. In the else branch, you could still have an error if the user provided a `load_escalation` array that is < analysis_period, in which case there should be a message returned to the user.

sjanzou:
Please use PR#417  to replace PR#416

ssc/common.cpp
brtietz: I agree it's rare, but there are some defaults such as `batt_pv_clipping_forecast` that are zero length arrays. Good thinking on the error message!

brtietz:
Good suggestions, thanks! A sample of the error message is below:

scaling_error

sjanzou:
Brian, Sorry about that! I had the wrong branch checked out... I am building and testing with your branch now and will have a simplified scalefactors class for your review soon (and in the more_fleible_scalefactors branch).

sjanzou:
more_flexible_scalefactors ssc branch updated and all ssc tests pass and all SAM configurations run and error above resolved along with unassigned input and any length input to scalefactor class. Please test and review.

brtietz:
I like the new is_assigned check, but I'm not keen on the behavior where if `count < nyears` year "count + 1" suddenly reverts back to the base year. I thought my original error message would be more clear for users. Sorry about not being clear that message was intentional!

@janinefreeman or @dguittet, which do you prefer?

janinefreeman:
I think an error message if the input array isn't the right length is the best path forward. We have some error messages like that in the UI already, so this would be consistent. Thanks!

sjanzou:
Not sure if error message or notice should be used...
Consider the use case, on the Electricity Loads page of a PVWatts Residential configuration.
Analysis period set to 25 years
User sets the load escalation to a schedule with 6 years of known escalation.
Should the simulation return an error or run with a notice that the no escalation was used for years 7 to 25?

brtietz:
I would think in that case the user would want year 6 escalation for years 7 through 25, and could set that in the schedule. Reverting to the base year (a scale factor of 1.0) doesn't seem like the right behavior.

PR #418, opened 2020-06-30
Enable non-annual simulations in pvsamv1 and pvwattsv7

janinefreeman:
Updates to enable non-annual (single timestep or multiple, non-8760 timestep) simulations via weather input data (NOT weather files) for both pvsamv1 and pvwattsv7. Added an annual simulation flag to the weather data provider class to identify continuous, hourly/subhourly, single year weather data, which is treated the same as it was previously, versus "non-annual" simulations, which produce timeseries data but do not produce annual results (including capacity factor). 

All calculations in both cmods are updated to include NO calculations dependent on a continuous 8760 weather dataset, but rather to be completely dependent on the timestamp in the weather data itself. If the weather data is non-annual, lifetime (nyears) is always 1. "Hour_of_year" is calculated based on timestamp in order to properly index hourly shading and loss arrays that are still 8760 inputs. 

This feature is NOT enabled for battery-coupled modeling.

Some timestep-dependent features (snow model, wind stow model) give warnings because the timestep will be assumed to be 1 hour rather than calculated by difference between timesteps, so effects can be overestimated.

Minute timestamps are now required for weather input data (weather files remain unchanged), and minute timestamps are now required to go from 0-59 for both weather data and weather files, rather than 1-60, due to shading calculations that must occur at the actual timestamp. One test weather file was updated for this new convention (previously had timestamps at 15,30,45,60) which resulted in some test answer changes.

Errors are thrown if hour column does not range from 0-23 or 1-24. Some tests updated accordingly. 

Beam shading inputs are now calculated using timestamps rather than indices. Timeseries beam shading inputs are disabled for non-annual simulations because it would be difficult to ensure that timeseries inputs are aligned with non-annual inputs.

All tests pass, and all default configuration results remain unchanged in SAM repo.

I recommend diffing code with something other than Git, because Git is struggling hard with whitespace/indenting changes. :)


shared/lib_pv_io_manager.cpp
```diff
@@ -123,26 +124,32 @@ Irradiance_IO::Irradiance_IO(compute_module* cm, std::string cmName)
...
+		throw exec_error(cmName, "subhourly and non-annual weather files must specify the minute for each record");
 
 	weatherDataProvider->header(&weatherHeader);
 
 	//total number of records in the weather file (i.e. 8760 * timestep)
 	numberOfWeatherFileRecords = weatherDataProvider->nrecords();
-	stepsPerHour = numberOfWeatherFileRecords / 8760;
-	dtHour = 1.0 / stepsPerHour;
+	dtHour = 1.0; //initialize these values to 1 for non-annual simulations
+	stepsPerHour = 1.0;
```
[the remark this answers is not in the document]
janinefreeman: Yes, I think this is going to be a tricky point for folks to follow. A PySAM example is a great idea. As far as I tried to program, everything uses the _timestamp_ correctly, but those things which would require the _time interval_ will not be done correctly, which we decided to do rather than calculate it from timestamps since folks might be entering discontinuous data.

I tried to include warnings or disallow cases where the time interval would be used incorrectly:
-batteries (not allowed with non-annual simulations)
-timestep beam shading calculations (not allowed)
-snow model (warning issued)
-wind stow model (pvwattsv7 only, warning issued)

I just realized I should add a warning that the transient thermal model in pvsamv1 won't be triggered, so I will take care of that. Are there any other pieces you can see that I'm missing? I searched for `dtHour`, `ts_hour`, and `steps_per_hour `in pvsamv1 to find all cases.

ssc/cmod_pvsamv1.cpp
```diff
@@ -2049,135 +2066,138 @@ void cm_pvsamv1::exec( ) throw (general_error)
...
-				{
-					batt->initialize_time(iyear, hour, jj);
-					batt->check_replacement_schedule();
-				}
+			// Battery replacement
+			if (en_batt && (batt_topology == ChargeController::DC_CONNECTED))
+			{
+				// calculate timestep in hour for battery models- this will work for annual simulations
+				// and non-annual simulations are not allowed for battery models, so this should work
+				size_t jj = idx % hour_of_year;
```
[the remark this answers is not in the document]
janinefreeman: Oooh good point on checking for zero. I merged patch in last week and got a pretty clean merge, but it will probably take some more work to then merge from patch to develop. Is it better to make another branch before merging into develop, or to wait til this is merged into patch (from a not confusing git perspective)?

shared/lib_pv_io_manager.cpp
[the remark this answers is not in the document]
dguittet: I don't recall seeing any other model. I'm definitely agree that all the warnings are essential. My suggestion was just to have that information more explicit for people when they are getting into the model to know what its limitations are-- even the list written here would be good for future reference. :)

ssc/cmod_pvsamv1.cpp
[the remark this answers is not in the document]
dguittet: I think it's fine to merge into a new branch on develop. Once the conflicts are resolved there, there shouldn't be any conflicts merging into develop-- I hope! I just wouldn't merge into patch without fixing the issue first.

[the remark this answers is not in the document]
janinefreeman: Will do!

shared/lib_util.cpp
```diff
@@ -896,6 +896,20 @@ size_t util::hour_of_day(size_t hour_of_year)
 	return (hour_of_year) % 24;
 }
 
+//returns the hour index in the year (0-8759) of a given month, day, and time in hours
+size_t util::hour_of_year(size_t month, size_t day, size_t hour)
+{
+	size_t h = 0;
```
sjanzou, line 902: Can we add an explicit check here that the input month is between 1 and 12, the input day is between 1 and days_in_month[month] and the input hour is between 0 and 23?

ssc/cmod_pvwattsv1.cpp
```diff
@@ -302,7 +302,7 @@ class cm_pvwattsv1 : public compute_module
 			p_sunup[i] = (ssc_number_t)sunup;
 
 			p_shad_beam[i] = 1.0f;
-			if ( shad.fbeam( i, solalt, solazi ) )
```
sjanzou, line 305: This had to be reverted before as all tests were failing with row translation calculations in the shading_factors classes. The results of the test were close but off and the old method had to be used.

ssc/cmod_pvwattsv5.cpp
```diff
@@ -532,7 +532,7 @@ class cm_pvwattsv5 : public cm_pvwattsv5_base
 					p_aoi[idx] = (ssc_number_t)aoi;
 
 					double shad_beam = 1.0;
-					if (shad.fbeam(hour, solalt, solazi, jj, step_per_hour))
+					if (shad.fbeam(hour, wf.minute, solalt, solazi))
```
sjanzou, line 535: This had to be reverted before as all tests were failing with row translation calculations in the shading_factors classes. The results of the test were close but off and the old method had to be used. Did the shading tests for pvsam_v1 and pvwatts and for the swh pass with this change? Also, are there shading tests for swh and pvwatts?

ssc/common.cpp
```diff
@@ -1077,25 +1078,25 @@ bool shading_factor_calculator::use_shade_db()
...
-	if (hr_step >= m_steps_per_hour) hr_step = m_steps_per_hour - 1;
-	if (hr_step < 0) hr_step = 0;
-	ndx += hr_step;
+	// this function needs to handle different simulation timesteps and timeseries shading input timesteps
+	// therefore, start the row index by multiplying the hour of the year by the number of timesteps per hour in the timeseries shading input
+	size_t ndx = hour_of_year * (size_t)m_steps_per_hour;
+
+	// then figure out how many row indices to add for the subhourly timeseries entries based on the minute stamp
+	ndx += floor((int)minute / (60 / (m_steps_per_hour)));
+
```
sjanzou, line 1089: How does this work if the minute column is specified as 30 for an hourly weather file? 
The CMPvsamv1PowerIntegration_cmod_pvsamv1.NoFinancialModelShading tests were failing with this change previously.
image

test/ssc_test/cmod_pvsamv1_test.cpp
```diff
@@ -416,7 +422,7 @@ TEST_F(CMPvsamv1PowerIntegration_cmod_pvsamv1, NoFinancialModelShading)
 		EXPECT_NEAR(calculated_value, annual_energy_expected[0], m_error_tolerance_hi);
```
sjanzou, line 422: If this test is passing with the annual_energy_expected as on line 396,402 above, then the size_t get_row_index_for_input in common.cpp is probably okay! Great job!

test/ssc_test/cmod_swh_test.cpp
```diff
@@ -16,7 +16,7 @@ TEST_F(CM_SWH, ResidentialDefault_cmod_swh) {
 }
 
 TEST_F(CM_SWH, ResidentialDefaultUsingData_cmod_swh) {
-    auto weather_data = create_weatherdata_array();
+    auto weather_data = create_weatherdata_array(8760);
     ssc_data_unassign(data, "solar_resource_file");
     ssc_data_set_table(data, "solar_resource_data", &weather_data->table);
```
sjanzou, line 21: Maybe add a shading test here and compare with patch branch results?

shared/lib_util.cpp
janinefreeman: Good suggestion! Done!

ssc/cmod_pvwattsv1.cpp
janinefreeman: This is different logic than was used before, and all tests pass with this logic.

ssc/cmod_pvwattsv5.cpp
janinefreeman: This is different logic than was failing before. All shading tests pass with this logic. I don't know that there are tests for swh and pvwatts shading.

ssc/common.cpp
janinefreeman: This is a different implementation of this logic than was used previously, and all tests pass with this logic (including CMPvsamv1PowerIntegration_cmod_pvsamv1.NoFinancialModelShading). Added a comment to explain the logic better, using the 30 minute example.

test/ssc_test/cmod_pvsamv1_test.cpp
janinefreeman: Haha thanks!

PR #421, opened 2020-07-04
Explicitly address SAM issue 343

sjanzou:
Update and create separate enable and validations for time step buy rates

### DIFF
--- a/ssc/cmod_utilityrate5.cpp
+++ b/ssc/cmod_utilityrate5.cpp
@@ -71,6 +71,8 @@ static var_info vtab_utility_rate5[] = {
 	// time step rates
 	{ SSC_INPUT, SSC_NUMBER, "ur_en_ts_sell_rate", "Enable time step sell rates", "0/1", "", "Electricity Rates", "?=0", "BOOLEAN", "" },
 	{ SSC_INPUT, SSC_ARRAY, "ur_ts_sell_rate", "Time step sell rates", "0/1", "", "Electricity Rates", "", "", "" },
+	// add separately to UI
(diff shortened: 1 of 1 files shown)


ssc/cmod_utilityrate5.cpp
```diff
@@ -1381,6 +1383,47 @@ class cm_utilityrate5 : public compute_module
...
+
+				// assign timestep values for utility rate calculations
+				size_t idx = 0;
+				ssc_number_t br;
+				br = 0;
+				size_t step_per_hour = m_num_rec_yearly / 8760;
+				//time step rates - fill out to number of generation records per year
+				// handle cases
+				// 1. if no time step rate  s
+				// 2. if time step rate  has 8760 and gen has more records
```
brtietz, line 1413: In the case where gen has > 8760 records and buy rate has 8760, this would not properly interpolate the subhourly records. For example, if gen had 4 steps per hour, all buy rates after March would be 0. Is this likely to be an issue or is this something the GUI checks for?

sjanzou: See lines 1417 through 1430 below. cnt is the number of buy rate records and ts_step_per_hour is the number of timesteps for the but rate records and step_per_hour is the number of timesteps of the yearly gen record. So, for the case where the yearly gen has > 8760 records, say steps_per_hour =4 from your example, and the buy rate has 8760=cnt records. Then, ts_steps_per_hour is 1 and idx only gets incremented once per the ii loop over step_per_hour. Thus, the buy rates would not run out after March. 

Make sense? A little convoluted but works without issue.

brtietz: That makes sense, thanks for the detailed explanation!

PR #422, opened 2020-07-06
Dependent inputs

dguittet:
Enabling & reviewing @Matthew-Boyd's `compute_module::exec()`'s call to new `evaluate` function that calls `ssc_equations` so that all interdependent inputs are correctly modified. Nice job!

This is part of the push to migrate equations and callbacks that modify SSC inputs from the GUI. This allows SDK and PySAM users to modify compute_module inputs without worrying that they are missing variable updates and logic in the GUI.

This works by looking up all ssc_equations that apply to a given cmod by matching the name-- hence a new name field to `compute_module`.

Added a field to `ssc_equation_entry` called `auto_eval` for equations that should be run in the `evaluate` loop. https://github.com/NREL/ssc/blob/1c065b2e53e99391e906c90e6ab2d8194378fff1/ssc/ssc_equations.h#L42

@Matthew-Boyd, could you retest these for all the CSP cmods which have associated equations?

@sjanzou @janinefreeman @brtietz FYI

### DIFF
--- a/ssc/core.cpp
+++ b/ssc/core.cpp
@@ -28,64 +28,57 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "core.h"
 #include "ssc_equations.h"
 
-const var_info var_info_invalid = {	0, 0, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL };
+const var_info var_info_invalid = {0, 0, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL};
 
-compute_module::compute_module( )
-	:  m_handler(NULL), m_vartab(NULL), m_infomap(NULL)
-{
-	/* nothing to do */
+compute_module::compute_module()
+        : m_handler(NULL), m_vartab(NULL), m_infomap(NULL) {
+    /* nothing to do */
 }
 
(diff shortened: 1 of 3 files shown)


dguittet:
Thanks! Now that we know it's working, adding @sjanzou to review code changes to `core.cpp`

ssc/core.cpp
```diff
@@ -226,7 +217,7 @@ bool compute_module::evaluate()
             }
             default:
             {
-                float time = -1.;
+//                float time = -1.;
                 //log(variable_name + " of data type " + var_data::type_name(variable_data->type) + " is not supported for ssc_equations", SSC_ERROR, time);
```
sjanzou, line 221: Why is this commented out? Ben Elliston mentioned this and was noted on his pull request

dguittet: I think throwing an error is a good idea! I'll add that in. I commented out the line since it doesn't do anything.

dguittet: Uncommented out the error message. Will merge once Travis tests pass

dguittet:
@Matthew-Boyd Do you mind reviewing this to make sure it works again? I did some refactoring to allow for comparison of SSC_TABLE variables since this was required to run a SingleOwner equation with the table variable `dc_adjust`. Otherwise, with the `default` case error message enabled, SingleOwner configurations were failing.

PR #423, opened 2020-07-07
Silence warning about unused 'time' variable

bje-:


ssc/core.cpp
```diff
@@ -225,7 +225,7 @@ bool compute_module::evaluate()
             }
             default:
             {
-                float time = -1.;
+                //float time = -1.;
```
sjanzou, line 228: @dguittet  why is the log error message commented out? If SSC_TABLE are used, should there be a separate case here and the default case still return false?

dguittet: I think that's a question for Matt, regarding this pull request: https://github.com/NREL/ssc/pull/422

Do you mind moving this question there?

dguittet: Resolving in #422

PR #424, opened 2020-07-07
core.cpp (compute_module::evaluate): Return true if all OK

bje-:
This function returns `bool`, but falls off the bottom of the function without a return value.


ssc/core.cpp
```diff
@@ -125,6 +125,7 @@ bool compute_module::evaluate()
                 return false;
             }
         }
+	return true;
```
sjanzou, line 128: Should return true even though return value not used.

Issue #412, opened 2020-06-22
Voltage table may result in roundtrip eff > 100


dguittet:
@brtietz Here's a link to a .sam file with a Li-ion case that shows the same issue as the vanadium flow case:

https://sam.nrel.gov/media/kunena/attachments/101439/VRF_2020-06-19.sam

And, for reference the forum post where it is posted:

https://sam.nrel.gov/forum/forum-general/3018-battery-efficiencies.html?start=6#8329

Thanks,
Paul.

_Originally posted by @cpaulgilman in https://github.com/NREL/ssc/issues/404#issuecomment-646855931_

---------

@cpaulgilman This looks like a general problem with the voltage table. Since the voltage table is only a function of DOD, the battery will spend some of its time charging at a lower voltage than it discharges.

@dguittet My intuition on this is to add a `-I * R` term to the voltage table's calculate_voltage_for_current function, to account for the cyclic losses similar to the Shephard-Tremblay electrochemical model. Thoughts?

_Originally posted by @brtietz in https://github.com/NREL/ssc/issues/404#issuecomment-647606082_

dguittet:
@brtietz I wonder if it makes sense to add an external term since I think of the voltage table model as incorporating everything that affects voltage-- polarization, internal or aggregated resistance, etc. So if I had the curve output from the vanadium model for instance and put that into the table, would it make sense to add another IR loss to it?

I think hysteresis was only explicitly added for Lead acid, NiMH, and NiCD in newer [versions](https://pdfs.semanticscholar.org/8f16/68ffef08c83a3a69f8f3c557d04cd9a5ffc2.pdf) of the Shephard Tremblay model. 

I wonder how much of the cyclic losses we see in SAM is an artifact of how the integral P = VI is calculated as a single right-hand Riemann sum using voltage at the end of the time step. With the actual integral, the roundtrip efficiency should always been 100% for the voltage table.

I remember you fitted the kokam data to the shephard tremblay parameters. Could you please run a few a cycles comparing SOC vs Voltage during charging and discharging using just `_E0 - _K * (Q_cell / (Q_cell - it)) + _A * exp(-_B0 * it) - R * I` and see how the cyclic losses compare?

brtietz:
Can you expand on what you're thinking as far as an external term? The voltage table as currently implemented can't model hysteresis, we would need to be able to specify different curves for charging and discharging. I'm open to that approach, but feel as though it's a larger burden on the user than simply specifying the internal resistance.

[...]

PR #425, opened 2020-07-07
Fix round trip efficiency issues for voltage table.

brtietz:
All efficiencies for https://github.com/NREL/ssc/issues/412 are now between 89% and 91%. A number of fixes from develop were needed, as well as an update to the starting assumption for DOD_best in case the projected SOC lies outside of the voltage table.

To test: run https://sam.nrel.gov/media/kunena/attachments/101439/VRF_2020-06-19.sam and check that the annual charged energy is greater than the annual discharged energy in all cases.

After this gets merged into patch, I'll setup a separate PR for merging into develop, since a lot of files have moved and many of these changes are already there.


dguittet:
I'd like to check these changes against the tests in develop as there are just so many more than on patch.

Could you please create that separate PR for merging into develop first and let's take a look at whether all the tests (voltage_table_lib_battery_voltage_test for instance, but all the dispatch ones also) pass on there before moving forward with this one?

PR #433, opened 2020-07-16
Batt subminute

### LINKED ISSUE #432: Sub-minute timesteps for BatteryStateful
Sub-minute timesteps were not enabled due to SAM simulation intervals being at least minutes. The pv models have moved away from this with the non-annual simulations. Need to investigate whether battery timesteps can be similarly made flexible.

dguittet:
Changes:
 - Remove error for battery timesteps less than a minute
 - Update initial guesses for `calculate_max_discharge_w`, `calculate_max_charge_w` and `calculate_current_for_target_w` for each voltage model when applicable
 - Add new `SubMinute` tests to dynamic voltage model, voltage table and vanadium voltage classes for 1/200 or 1/360 `dt_hr`
 - Modify a test result from `CMBattwatts_cmod_battwatts, ResidentialDefaultsLeadAcid`-- difference is due to change in initial guesses but not sure of the details

close #432

### DIFF
--- a/shared/lib_battery_voltage.cpp
+++ b/shared/lib_battery_voltage.cpp
@@ -36,8 +36,6 @@ bool voltage_state::operator==(const voltage_state &p) {
 void voltage_t::initialize() {
     state = std::make_shared<voltage_state>();
     state->cell_voltage = params->Vnom_default;
-    if (params->dt_hr < 1 / 60.)
-        throw std::runtime_error("Battery time step size must be greater than 1/60th of hour.");
 }
 
 voltage_t::voltage_t(int mode, int num_cells_series, int num_strings, double voltage, double dt_hour) {
--- a/shared/lib_battery_voltage.cpp
+++ b/shared/lib_battery_voltage.cpp
@@ -412,7 +412,8 @@ double voltage_dynamic_t::calculate_current_for_target_w(double P_watts, double 
     }
 
     double x[1], resid[1];
-    x[0] = solver_power / state->cell_voltage / 10;     // initial guess is important
+    double initial_guess_factor = 0.1;       // trial and error, this gave correct results for a variety of timescales
+    x[0] = solver_power / state->cell_voltage * initial_guess_factor;
     bool check = false;
 
     newton<double, std::function<void(const double *, double *)>, 1>(x, resid, check, f,
@@ -538,7 +539,8 @@ double voltage_vanadium_redox_t::calculate_current_for_target_w(double P_watts, 
                                                                 this, _1, _2);
 
     double x[1], resid[1];
-    x[0] = solver_power / state->cell_voltage / 10;
+    double initial_guess_factor = 0.1;       // trial and error, this gave correct results for a variety of timescales
+    x[0] = solver_power / state->cell_voltage * initial_guess_factor;
     bool check = false;
 
     newton<double, std::function<void(const double *, double *)>, 1>(x, resid, check, f,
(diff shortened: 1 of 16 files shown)


shared/lib_battery_voltage.cpp
```diff
@@ -414,7 +414,7 @@ double voltage_dynamic_t::calculate_current_for_target_w(double P_watts, double
     }
 
     double x[1], resid[1];
-    x[0] = solver_power / state->cell_voltage;
+    x[0] = solver_power / state->cell_voltage / 10;     // initial guess is important
```
brtietz, line 417: Can the comment explain why 10 is the right number?

shared/lib_battery_voltage.cpp
```diff
@@ -542,7 +540,7 @@ double voltage_vanadium_redox_t::calculate_current_for_target_w(double P_watts,
                                                                 this, _1, _2);
 
     double x[1], resid[1];
-    x[0] = solver_power / state->cell_voltage;
+    x[0] = solver_power / state->cell_voltage / 10;
```
brtietz, line 543: Given that it's in here twice, should 10 be a member variable of lib_battery_voltage?

test/shared_test/lib_battery_voltage_test.cpp
```diff
@@ -37,16 +37,67 @@ TEST_F(voltage_dynamic_lib_battery_voltage_test, updateCapacityTest){
...
+    cap->updateCapacity(I, dt_hour); // qmx = 10, q0 = 3
+    model->updateVoltage(cap->q0(), cap->qmax(), cap->I(), 0, dt_hour);
+    EXPECT_NEAR(model->cell_voltage(), 4.01, tol);
+    EXPECT_NEAR(cap->q0(), 5, tol);
+
+    I = -2;
+    cap->updateCapacity(I, dt_hour); // qmx = 10, q0 = 5
+    model->updateVoltage(cap->q0(), cap->qmax(), cap->I(), 0, dt_hour);
+    EXPECT_NEAR(model->cell_voltage(), 4.1, tol);
+    EXPECT_NEAR(cap->q0(), 5, tol);
```
brtietz, line 94: Can this test have a smaller tolerance? It would be nice to see the small changes in q0.

test/shared_test/lib_battery_voltage_test.cpp
```diff
@@ -205,16 +328,65 @@ TEST_F(voltage_table_lib_battery_voltage_test, updateCapacityTest){
...
+    cap->updateCapacity(I, dt_hour); // qmx = 10, q0 = 3
+    model->updateVoltage(cap->q0(), cap->qmax(), cap->I(), 0, dt_hour);
+    EXPECT_NEAR(model->cell_voltage(), 3.689, tol);
+    EXPECT_NEAR(cap->q0(), 5, tol);
+
+    I = -2;
+    cap->updateCapacity(I, dt_hour); // qmx = 10, q0 = 5
+    model->updateVoltage(cap->q0(), cap->qmax(), cap->I(), 0, dt_hour);
+    EXPECT_NEAR(model->cell_voltage(), 3.69, tol);
+    EXPECT_NEAR(cap->q0(), 5, tol);
```
brtietz, line 383: A lower tolerance here would be good as well.

shared/lib_battery_voltage.cpp
dguittet: I don't think so because it's not inherent... it didn't have to be the same here and in the previous function I think

test/shared_test/lib_battery_voltage_test.cpp
dguittet: done

PR #438, opened 2020-07-18
Include <cmath> for fabs()

bje-:
Not sure why this header isn't needed on other platforms. For a C++11 compiler, the standard says `<cmath>` must be included.


bje-:
Ping?

PR #439, opened 2020-07-20
Subminute thermal model

brtietz:
The previous version of the thermal model could be numerically unstable at small timesteps and large currents, due to the current term not having an exponential to smooth it. This update includes the current term in the exponential, such that the battery will approach the same steady state temperature at large and small timesteps.

Old behavior, 10 second timesteps:

batt_stateful_old_thermal_temp

New behavior:

image

Reference for converging to steady state temperature: https://web.mit.edu/16.unified/www/FALL/thermodynamics/notes/node129.html (the convection constants are different, but the underlying differential equations are the same)

### DIFF
--- a/shared/lib_battery.cpp
+++ b/shared/lib_battery.cpp
@@ -117,14 +117,15 @@ void thermal_t::updateTemperature(double I, size_t lifetimeIndex) {
     }
 
     // the battery temp is the average temp over that step, starting with temp from end of last timestep
-    double source = I * I * params->resistance / (params->surface_area * params->h) + state->T_room;
+    
+    double T_steady_state = I * I * params->resistance / (params->surface_area * params->h) + state->T_room;
     double diffusion = exp(-params->surface_area * params->h * dt_sec / params->mass / params->Cp);
     double coeff_avg = params->mass * params->Cp / params->surface_area / params->h / dt_sec;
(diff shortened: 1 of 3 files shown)


shared/lib_battery.cpp
```diff
@@ -117,14 +117,15 @@ void thermal_t::updateTemperature(double I, size_t lifetimeIndex) {
...
 
     // the battery temp is the average temp over that step, starting with temp from end of last timestep
-    double source = I * I * params->resistance / (params->surface_area * params->h) + state->T_room;
+    
+    double T_steady_state = I * I * params->resistance / (params->surface_area * params->h) + state->T_room;
     double diffusion = exp(-params->surface_area * params->h * dt_sec / params->mass / params->Cp);
     double coeff_avg = params->mass * params->Cp / params->surface_area / params->h / dt_sec;
-    state->T_batt = (state->T_batt_prev - state->T_room) * coeff_avg * (1 - diffusion) + source;
+    state->T_batt = (state->T_batt_prev - T_steady_state) * coeff_avg * (1 - diffusion) + T_steady_state;
     state->heat_dissipated = (state->T_batt - state->T_room) * params->surface_area * params->h;
```
brtietz, line 125: Looking at this again: does heat_dissipated need to be multiplied by t? Are the units on this supposed to be Watts or Joules?

dguittet: Units are supposed to be in Watts, and h is [W/m2K]

PR #444, opened 2020-07-31
Batt stateful updates

dguittet:
Bugs fixed:
 - Voltage model did not read all required voltage parameters
 - Initial guess for some numerical solutions for current calculations was NaN if voltage was 0
 - Voltage model initialization wasn't happening in all initialization paths, so move to `set_initial_soc` function that is called by `battery_t` during initialization

Changes:
 - `heat_dissipated` units from Watts to kW
 - Add `heat_dissipated` to thermal_t tests


test/shared_test/lib_battery_test.cpp
```diff
@@ -63,7 +63,7 @@ TEST_F(lib_battery_thermal_test, updateTemperatureTestSubMinute) {
     }
     avgTemp /= 600.0;
     auto s = thermal_state({ 93.49, 16.86, 16.85 });
-    EXPECT_NEAR(avgTemp, s.T_batt, 0.02) << "updateTemperatureTest: 1";
```
brtietz, line 66: I'm not opposed to adding the "compare state" function here, but the reason I was using EXPECT_NEAR was to ensure that the average temperature during the sub-minute steps was the same as the average temperature during the hourly steps. Can we do both comparisons for more accurate coverage?

test/ssc_test/cmod_battery_stateful_test.cpp
```diff
@@ -95,3 +95,17 @@ TEST_F(CMBatteryStatefulIntegration_cmod_battery_stateful, SubMinute) {
...
+
+    auto copy = json_to_ssc_data(js.c_str());
+
+    mod = ssc_stateful_module_create("battery_stateful", copy);
+    copy = json_to_ssc_data(js.c_str());
+    ssc_module_exec(mod, copy);
+
+    js = ssc_data_to_json(copy);
+
+    printf("%s\n", js.c_str());
```
brtietz, line 111: Does this test need to assert/expect something?

test/shared_test/lib_battery_test.cpp
dguittet: Gotcha, I'll add that back in. Thanks

test/ssc_test/cmod_battery_stateful_test.cpp
dguittet: Fixed

Issue #445, opened 2020-08-07
Battery Current Adjustment with SOC bug


dguittet:
**Describe the bug**
Assume there is 1 kW required from the battery and this is calculated at the beginning of the time step as requiring 2 A from the battery. When the thermal response is applied to the capacity model, reducing the capacity, the adjustment to the current can result in extreme current changes.

**To Reproduce**
Add these tests to `lib_battery_capacity_test.cpp`:

```
TEST_F(LiIon_lib_battery_capacity_test, CurrentChangedBySOCLimits) {
    double I = 35.;
    double dt = 1./60.;

    auto cap = std::make_shared<capacity_lithium_ion_t>(19984, 90, SOC_max, SOC_min, dt);
    cap->updateCapacityForThermal(90);
    cap->updateCapacity(I, dt);

    printf("I %f\n", I);                 // I is now 53956.800000
}

TEST_F(LiIon_lib_battery_capacity_test, CurrentChangedBySOCLimits1) {
    double I = -35.;
    double dt = 1./60.;

    auto cap = std::make_shared<capacity_lithium_ion_t>(19984, 10, SOC_max, SOC_min, dt);
    cap->updateCapacityForThermal(90);
    cap->updateCapacity(I, dt);

    printf("I %f\n", I);                 // I is now -41966.400000
}
```

**Expected behavior**
The current adjustment cannot be calculate correctly within `capacity_t` since more information is required for `battery_t` to provide the requested power. The current adjustment should be minor and should signal to the `battery_t` iteration that the required current needs to be recomputed.

dguittet:
closed in #446

PR #446, opened 2020-08-08
Fix battery current adjustment when hitting SOC limits

dguittet:
Please see #445 

Changes
 - only apply current adjustments to charge current if overcharged and discharge current if undercharged
 - update test results


dguittet:
Thanks, good point! Added sam test results changes to https://github.com/NREL/SAM/pull/369. Merging.

Issue #447, opened 2020-08-10
Reduced battery cycling with interpolated weather files


brtietz:
**Describe the bug**
When running subhourly peak shaving for behind the meter batteries, the grid power target slowly escalates over the course of the year. This is not the desired (or hourly) behavior, where this target resets each month.

Bug was introduced between 2020.2.29 r2 and 2020.2.29 r3, it also exists on develop.

**To Reproduce**
1. Run an hourly simulation with peak shaving dispatch, save the graphs
2. Subsample the weather file and the load to subhourly (say 15 min increments)
3. Re-run the simulation. The grid power target will escalate and dispatch will be decreased.

The root cause of the issue is that the first 30 minutes of each day in the interpolated weather file is listed as the previous day in the day column. This causes the battery code to miss it's monthly reset.

The ultimate solution to this might be updating a macro in the SAM repository, but SSC should either throw an exception or pass the correct hour_of_year to the battery code.

**Expected behavior**
Hourly and subhourly dispatch should approximately match, if the load and weather files have been interpolated.

PR #448, opened 2020-08-11
Throw exception if weatherfile goes backwards in time

brtietz:
Throw an exception for any weatherfile subject to issue https://github.com/NREL/ssc/issues/447

The exception occurs when running a simulation with an old interpolated weather file for all relevant cases, and if https://github.com/NREL/SAM/pull/364 is checked out, it will appear when selecting interpolated weather file on the battery modules.

To test:
- Check out and build SAM on this branch
- Select an interpolated weather file from 2020.2.29 r3 (or develop prior to today) and try to run it
- An exception should be raised about hour 0 occurring after hour 23 (or similar)

Weather files interpolated with the code on https://github.com/NREL/SAM/pull/364 should not trigger the exception.

### DIFF
--- a/shared/lib_weatherfile.cpp
+++ b/shared/lib_weatherfile.cpp
@@ -1059,6 +1059,7 @@ bool weatherfile::open(const std::string &file, bool header_only)
 	// from 1-24 standard to 0-23
 	int tmy3_hour_shift = 1;
 	int n_leap_data_removed = 0;
+    bool subtract_hour = false;
 
 	for (int i = 0; i < (int)m_nRecords; i++)
 	{
@@ -1338,7 +1339,7 @@ bool weatherfile::open(const std::string &file, bool header_only)
 		}
 		else if (m_type == WFCSV)
 		{
-
+            
 			for (;;)
 			{
 				getline(ifs, buf);
@@ -1376,6 +1377,11 @@ bool weatherfile::open(const std::string &file, bool header_only)
 					continue;
 				}
 
+                if (m_columns[HOUR].data[i] > 23)
+                {
+                    subtract_hour = true;
+                }
+
 				if (m_columns[MINUTE].data[i] > 59)
 				{
 					m_message = "minute column must contain integers from 0-59";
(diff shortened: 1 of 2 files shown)


brtietz:
Ah, my change to invoke.cpp in the SAM PR wasn't correct for that callback. I've reverted that change and updated the behavior in https://github.com/NREL/SAM/pull/368, let me know if that meets your expectations.

cpaulgilman:
Looks good now. I tweaked the lib_weatherfile error message a bit in 32f4bcb6.

shared/lib_weatherfile.cpp
```diff
@@ -1365,6 +1377,11 @@ bool weatherfile::open(const std::string &file, bool header_only)
 					continue;
 				}
 
+                if (m_columns[HOUR].data[i] > 23)
+                {
+                    subtract_hour = true;
+                }
+
 				if (m_columns[MINUTE].data[i] > 59)
```
dguittet, line 1385: Sorry I'm late to this discussion. During implementation of the time-steppable PV model, I added a function to make sure hours were between 0-23. https://github.com/NREL/ssc/commit/01ef286ea05211ab21390d64cfe2ba999481eb36#diff-2ad11e7c78af69e327ea269d6cfea2c8 Is this partially duplicating the behavior?

brtietz: It looks like there are two paths for solar resource data, named "solar_resource_file" and "solar_resource_data" in the vartable. "solar_resource_data" was making the adjustment properly prior to this PR, but "solar_resource_file" was not. My debugger never went through common.cpp, so I ignored that code when doing this work, hence the duplicate function.

I'm open to having one function for both, but would prefer it ended up in lib_weatherfile.

dguittet: There's the same function for both weather data and weatherfiles: https://github.com/NREL/ssc/blob/32f4bcb69a05681866b7a4358bfb921895c1ad63/shared/lib_weatherfile.cpp#L438 implemented differently since the two classes organize their data differently.

Perhaps moving `start_hours_at_0` to be called earlier like you are doing here?

PR #459, opened 2020-10-15
EPW file format check

mjprilliman:
epw files based on tmy and tmy2 formats show 60 in the minutes column. The meaning of the time stamp is the time following the hour, so the minute should be interpreted as 0. The minute is not carried over to the SAM csv converter macro. Changing 60 to 0 fixes issues posted to the SAM forum. 

Also added a warning based on a forum post where the hour inputs were not integers (0 - 23.5). 

Closes SAM issue NREL/SAM#365


mjprilliman:
Formatting changes (tried to update to .editorconfig file) created a large diff. Look at lines 1263 and 1649

mjprilliman:
SAM forum posts:
https://sam.nrel.gov/forum/forum-general/3102
https://sam.nrel.gov/forum/forum-general/3095

Discussion of meaning of minutes column for epw formats based on tmy and tmy2:
https://discourse.radiance-online.org/t/meaning-of-epw-files-minute-field-and-why-is-it-60-in-tmy2-based-epw-and-0-in-tmy3-based-epw/1462/3

mjprilliman:
@cpaulgilman @dguittet Running into behavior in the release and in develop where after converting the EPW file to SAM csv the results are not the same for the same simulation. Looks like GHI calculation and solar position angles are diverging at sunrise and sunset hours on some days. Time steps appear to be at 30 minutes for each hourly file. Still investigating.

mjprilliman:
@cpaulgilman @dguittet weather files based on TMY data would be considered the opposite of instantaneous right? So we should not iterate between the sunrise time and time step for those types of files?

cpaulgilman:
The TMY files from the NSRDB is contain instantaneous irradiance data measured on the half hour: https://nsrdb.nrel.gov/about/tmy.html.

Note that the table in that description appears to define Minute = 0, but the TMY files returned by the NSRDB have Minute = 30.

dguittet:
@cpaulgilman Did they change it to instantaneous recently? On page you posted, the TMY, TMY2 and TMY3 are lumped together , whereas I'm not sure whether the average vs instantaneous convention among them all is consistent, however I would guess that to be the case.

The documentation of TMY3 here https://www.nrel.gov/docs/fy08osti/43156.pdf states that the irradiance is average:

image

mjprilliman:
I'll do my own research as well. I was always under the (possibly incorrect) assumption that it was an average. Either way I should be able to come up with a fix

cpaulgilman:
The PSM data from the NSRDB is different from the old TMY3 and TMY2 data, although I'm having a hard time finding a reference to instantaneous measurements. The latest version of the PSM provides 5, 15, 30, and 60 minute data. The 60-minute data is not averaged from the higher resolution, but is taken from the 30-minute time stamp.

[This document](https://doi.org/10.1016/j.rser.2018.03.003) was recommended as the reference for the PSM in the recent NSRDB webinar. On p. 55, there is:

image

(The last sentence ends  "...from the GOES-East data to the NSRDB time stamps.")

I will follow up with someone from the NSRDB team to confirm.

dguittet:
I agree that the PSM is instantaneous. I guess I was missing that NSRDB's TMY data is not the original TMY data?

mjprilliman:
I think I may have overcomplicated this yesterday. Right now the instantaneous bool is set to true for any weather file that contains a minute column. This part of the code was not recognizing the minute column in the epw file because I was rewriting the column index that is checked in order to automatically calculate the correct minute value for epw files with 8760 60 values in the minute column that are being downloaded from EnergyPlus. This rewriting is most likely not necessary and I will revisit that code, or I can do what I am doing now and reset the index to the correct value after rewriting the minute column. 

Changing the instantaneous parameter behavior to account for all weather files types and input options concerning the minute column would likely be too much for this issue. It would likely need its own issue.

dguittet:
@mjprilliman @cpaulgilman what are next steps on this?

mjprilliman:
@dguittet @cpaulgilman  I think the original issue has been answered. The epw files are checked for values of 60 in the minute column and the minutes are rewritten if the files have 60 repeating. The converter macro now has a minute column for each of the file types, and the outputs from the original epw file match the results from the converted SAM csv file. We could get into the weeds on the instantaneous vs average interpretation of different file types but that seems like a separate issue. 

I think this just needs code review before merging.

mjprilliman:
@cpaulgilman If you have time this needs review before merging.

cpaulgilman (approved):
This fixes the original issues reported in https://github.com/NREL/SAM/issues/365.

PR #462, opened 2020-10-21
PVWatts5_1ts shading

nmerket:
This adds a `shading_percent` parameter to the PVWatts5_1ts model.


nmerket:
@dguittet

dguittet (changes requested):
Thanks for the contribution! This is good to merge. The only thing that's missing is adding this to our unit test for pvwattsv5_1ts here: https://github.com/NREL/ssc/blob/0e23a2b37fb2a7e02027664baa12f398369fca3f/test/ssc_test/cmod_pvwattsv5_test.cpp#L130

Could you please add this variable to the inputs and just check that the changes to the test results are as you expect?

ssc/cmod_pvwattsv5.cpp
```diff
@@ -671,7 +672,7 @@ class cm_pvwattsv5_1ts : public cm_pvwattsv5_base
 		double wspd = as_double("wspd");
 		double alb = as_double("alb");
 
-		double shad_beam = 1.0;
+		double shad_beam = 1.0 - as_double("shaded_percent") / 100.0;
```
nmerket, line 675: I just want to make sure I'm not getting things turned around here. Does this look right?

test/ssc_test/cmod_pvwattsv5_test.cpp
```diff
@@ -180,4 +179,17 @@ TEST_F(CMPvwattsV5Integration_cmod_pvwattsv5, singleTS) {
...
+    ssc_data_set_number(data, "shaded_percent", 50);
+    EXPECT_TRUE(ssc_module_exec(mod, data));
+    ssc_data_get_number(data, "poa", &val);
+    EXPECT_NEAR(val, 140.05, .1);
+    ssc_data_get_number(data, "tcell", &val);
+    EXPECT_NEAR(val, 13.36, .1);
+    ssc_data_get_number(data, "dc", &val);
+    EXPECT_NEAR(val, 106341, 1);
+    ssc_data_get_number(data, "ac", &val);
+    EXPECT_NEAR(val, 100463, 1);
```
nmerket, line 193: Added some testing with shading. It didn't decrease the power output as much as I thought it would, but then again I'm not a solar guy so maybe I don't have as good a sense of what the results _should_ be.

nmerket (commented):
@dguittet, now with testing!

test/ssc_test/cmod_pvwattsv5_test.cpp
dguittet: Thanks! I think this makes sense because the solar data here has very little beam irradiance compared to diffuse. The shaded_percent only gets applied to the beam so overall there's little effect. This and the (1-shaded_percent) is consistent with `shading_factor_calculator`'s `shading:timestep` with default `shading:string_option == -1`.

ssc/cmod_pvwattsv5.cpp
dguittet: Yes!

dguittet (approved):
I'll merge this once Travis is done!

nmerket:
Uh oh. It looks like Travis didn't like it. Any idea why?

PR #467, opened 2020-10-23
Reduce compiler warnings introduce by btm_price_signals branch

brtietz:
Clean up files that had compiler warnings introduced by price signals dispatch. lib_util and lib_time were out of scope for this, maybe we can clean those up after the release.


dguittet (approved):
Thanks!

PR #468, opened 2020-10-23
Apply Net Metering Credits to Future Bills

brtietz:
Previously (based on PR https://github.com/NREL/ssc/pull/382) the "Apply credits to future bills" option just delayed the check by a month. This PR implements @cpaulgilman 's request from that PR, and applies them as bill credits, reducing the bill amount subject to annual minimums.

The implementation leans on some of the Net metering with $ credits code in order to apply the true-up period credits.
 Comparisons between a rate that cuts a check vs a rate that applies the credits to future bills can be made in the attached SAM file: [rollover_tests.zip](https://github.com/NREL/ssc/files/5431680/rollover_tests.zip)

This will close https://github.com/NREL/SAM/issues/16, though the third part of that issue has been moved to its own issue: https://github.com/NREL/SAM/issues/410


cpaulgilman (approved):
@brtietz Thanks for untangling this mess of requests! This handling of net metering end of period rollovers seems to work as I would expect from month to month and across years.

ssc/cmod_utilityrate5.cpp
```diff
@@ -1787,6 +1786,15 @@ class cm_utilityrate5 : public compute_module
 			ssc_number_t dollars_applied = 0;
 			if (enable_nm)
 			{
+                if (rollover_credit && m == net_metering_credit_month)
+                {
+                    ssc_number_t year_end_dollars = monthly_cumulative_excess_energy[m] * as_number("ur_nm_yearend_sell_rate") * rate_esc;
+                    payment[c - 1] -= year_end_dollars;
```
sjanzou, line 1792: Are the hourly values matching with the bill when summed?

ssc/cmod_utilityrate5.cpp
```diff
@@ -1787,6 +1786,15 @@ class cm_utilityrate5 : public compute_module
...
+                    monthly_ec_charges[m] -= year_end_dollars;
+                    excess_dollars_earned[m] += year_end_dollars;
+                    dollars_applied += year_end_dollars;
```
sjanzou, line 1795: Are the monthly, hourly, and bill values matching?

test/ssc_test/cmod_utilityrate5_test.cpp
```diff
@@ -267,7 +267,7 @@ TEST(cmod_utilityrate5_eqns, Test_Residential_net_metering_credits_in_may_with_r
     EXPECT_NEAR(0, may_dollars, 0.1);
 
     float june_dollars = excess_dollars[credit_month + 1];
-    EXPECT_NEAR(50.28, june_dollars, 0.1);
+    EXPECT_NEAR(11.37, june_dollars, 0.1);
```
sjanzou, line 270: Does the change show the difference between the true up and the rollover when happening mid-year?

sjanzou (commented):
Looks good and thanks for digging through this!

ssc/cmod_utilityrate5.cpp
brtietz: Yes - for the file I attached the annual bill in the summary is $251 (rounded to the nearest dollar), and the sum of the hourly values is $251.19.

ssc/cmod_utilityrate5.cpp
brtietz: Yes.

test/ssc_test/cmod_utilityrate5_test.cpp
brtietz: The previous value here was for the "cut a check a month late" code, so the old value was the full bill for June. The new code applies the credit to the bill, so the June value is reduced. To compare with the "cut a check in the true up month" rate, see Test_Residential_net_metering_credits_in_may

PR #470, opened 2020-10-27
Solarpos spa tamb

mjprilliman:
-Update SPA implementation to use ambient temperature inputs in place of dry_temperature optional input where appropriate, as dry bulb temperature is used as the ambient temperature in most cases. 
-Update function documentation to reflect use of ambient temperature as an input
-Remove missed commented messages that are handled by git history instead.


shared/lib_irradproc.h
```diff
@@ -705,7 +705,7 @@ void calculate_eot_and_sun_rise_transit_set(double jme, double tz, double alpha,
 * \param[in] dut1 fractional second difference between UTC and UT which is used to adjust UTC for earth's irregular rotation rate (http://maia.usno.navy.mil/ser7/ser7.dat) (-1 to 1 second)
 * \param[in] alt altitude in meters
 * \param[in] pressure pressure in millibars
-* \param[in] temp temperature in degrees C
+* \param[in] temp ambient temperature (dry-bulb temperature) in degrees C
```
dguittet, line 708: Thanks for adding the additional comment! However, most users won't find this straight-away. A more user-friendly place to also add this to would be the labels in the var_info tables for pvwatts, irrad_proc, etc.

dguittet (approved):
Looks great! Think it's good to merge after my one comment.

PR #471, opened 2020-10-27
Inverter Temperature Derate based on Pmax ratio

### LINKED ISSUE #442: Inverter Temperature Derate of Max Power
**Describe the bug**
 - The inverter temperature derate curves apply derates to efficiency as a function of temperature. However, a derate applied to power is consistent with real world operation of many inverters.
 - In the GUI, there's y-axis unit confusion between ratio and percent.
 - Default in GUI should be more informative, showing a few Vdcs for the default inverter with piecewise functions.

**Expected behavior**
The y-axis should be the ratio of the max output power to the rated power. Once the maximum output is limited by a higher temperature, the existing inverter model will produce lower efficiencies across all input DC power. Example:
 
image

**Additional context**
Lower efficiency at high temperatures could be a real physical behavior, but likely isn't as important to capture as the power limiting (or current) built into many inverters. 

Here are a few other power vs temp curves for reference:
https://www.ge.com/renewableenergy/sites/default/files/related_documents/lv5plus-solar-inverter-datasheet.pdf, page 2.
https://files.sma.de/downloads/Temp-Derating-TI-en-15.pdf, page 2 for temp-only, page 3 for temp and Vdc.
https://solar.huawei.com/en/download?p=%2F-%2Fmedia%2FSolar%2Fattachment%2Fpdf%2Fla%2Fservice%2Fcommercial%2Fproduct%2Fdatasheet%2FSUN2000-60KTL-M0_Deratingcurve.pdf, page 2

mjprilliman:
-Changed inverter derate table to be based on ratio of max output power to rated power rather than efficiency
-Update Y-axis label to reflect the ratio rather than the percentage
-Update default inverter temperature derate table to reflect more realistic examples (resource provided by Will Hobbs from Southern Company: https://files.sma.de/downloads/Temp-Derating-TI-en-15.pdf)
-Closes #442


dguittet:
See also https://github.com/NREL/SAM/pull/412

shared/lib_shared_inverter.cpp
```diff
@@ -45,162 +45,161 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
+
+    // If slope is positive, set to zero with no derating
+    if (slopeInterpolated >= 0) return;
+    if (slopeInterpolated < -1) slopeInterpolated = -1;
+
+    // Power in units of W, ratio = max output / rated output
+    ratio += deltaT * slopeInterpolated;
+    if (ratio < 0) ratio = 0.;
+    loss = p_rated - (p_rated * ratio);
+    p_rated = p_rated * ratio;
```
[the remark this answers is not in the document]
dguittet: I think this on the right track, but the derate is happening to the max power of the inverter which means that the max DC power that the inverter can now handle at this time step is the reduced max DC power input. We're not going to worry about the efficiency changes given a smaller max DC power, just potentially increased inverter clipping.

So this function has to calculate the max power of the inverter, which is has to be accessed from the inverter classes:
```
	sandia_inverter_t * m_sandiaInverter;
	partload_inverter_t * m_partloadInverter;
	ond_inverter * m_ondInverter;
```

and usually called `Pdco`. 

I'd suggest

1. A new function in SharedInverter called `getInverterDCMaxPower` that gets the correct value depending on the inverter type (similar to how it's done in `calculateACPower`).

2. Change this function to use the DC input power.

3. Call getInverterDCMaxPower from here and apply a temp derate to that value if applicable.

shared/lib_shared_inverter.cpp
```diff
@@ -45,162 +45,161 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
-	if (slopeInterpolated >= 0) return;
-	if (slopeInterpolated < -1) slopeInterpolated = -1;
-
-	// Power in units of W, eff as ratio
-	double pDC = pAC/eff;
-	eff += deltaT* slopeInterpolated;
-	if (eff < 0) eff = 0.;
-	loss = pAC - (pDC * eff);
-	pAC = pDC * eff;
+    if (ratio == 0. || p_rated == 0.) return;
```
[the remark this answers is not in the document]
dguittet: ratio makes sense. pAC however is the calculated AC power of the inverter. the derate should be happening on the max dc power, so this will have to be renamed.

shared/lib_shared_inverter.cpp
```diff
@@ -211,66 +210,68 @@ void SharedInverter::calculateACPower(const double powerDC_kW_in, const double D
...
-	//Convert units to kW and scale to total array for all other outputs
-	convertOutputsToKWandScale(tempLoss, powerAC_Watts);
+    double power_ratio = 1.0;
+    if (m_tempEnabled) {
+        //use average of the DC voltages to pick which temp curve to use- a weighted average might be better but we don't have that information here
+        double avgDCVoltage = 0;
+        for (size_t i = 0; i < DCStringVoltage.size(); i++)
+            avgDCVoltage += DCStringVoltage[i];
+        avgDCVoltage /= DCStringVoltage.size();
+        calculateTempDerate(avgDCVoltage, tempC, powerAC_Watts, power_ratio, tempLoss);
```
dguittet, line 259: tempLoss will now be in DC. So any comments and the var_info table label should specify that

shared/lib_shared_inverter.cpp
```diff
@@ -211,66 +210,68 @@ void SharedInverter::calculateACPower(const double powerDC_kW_in, const double D
...
-	powerDC_kW = powerDC_Watts * util::watt_to_kilowatt;
-	convertOutputsToKWandScale(tempLoss, powerAC_Watts);
-
-	// In event shared inverter is charging a battery only, need to re-convert to negative power
-	if (negativePower) {
-		powerAC_kW = -1.0 * fabs(powerAC_kW);
-	}
+    double power_ratio = 1.0;
+    if (m_tempEnabled) {
+        calculateTempDerate(DCStringVoltage, tempC, powerAC_Watts, power_ratio, tempLoss);
```
dguittet, line 219: Since this code is now working to potentially clip the input DC power, this should be done before the calls to `acpower`.

test/shared_test/lib_shared_inverter_test.cpp
```diff
@@ -147,49 +147,49 @@ TEST_F(sharedInverterTest_lib_shared_inverter, tempDerateTest_lib_shared_inverte
...
 	EXPECT_NEAR(pAC, 100, e) << "case 9";
 
 	// 300<V and 40 < T, interpolated startT is 40. and slope is -0.4
 	V = 400.;
 	T = 41.;
 	reset();
-	inv->calculateTempDerate(V, T, pAC, eff, loss);
+	inv->calculateTempDerate(V, T, pAC, ratio, loss);
 	EXPECT_NEAR(pAC, 60, e) << "case 9";
 }
```
dguittet, line 194: Add a new test around `calculateACPower` similar to `calculateEffForACPower` but testing the temp derating instead.

shared/lib_pv_io_manager.cpp
```diff
@@ -1488,7 +1488,7 @@ void Inverter_IO::setupSharedInverter(compute_module* cm, SharedInverter* a_shar
         thermalDerateCurves.push_back(row);
     }
     int err = sharedInverter->setTempDerateCurves(thermalDerateCurves);
-    if (err > 1) {
-        throw exec_error("pvsamv1", "Inverter temperature derate curve " + util::to_string((int)(-err - 1)) + " is invalid.");
+    if (err > 0) {
+        throw exec_error("pvsamv1", "Inverter temperature derate curve row " + util::to_string((int)(err - 1)) + " is invalid.");
```
mjprilliman, line 1492: @dguittet This correction caused a (correct) test failure in PVYield where the inverter temperature derate curves are not defined for the Inverter Mermoud Lejeune Model. Should this model include the option for temperature derates. How should this be handled?

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,275 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
+    if (ratio == 0. || p_dc_rated == 0.) return;
+
+    double slope = 0.0;
+    double startT = 0.0;
+    double Vdc = 0.0;
+    double slope2 = 0.0;
+    double startT2 = 0.0;
+    double Vdc2 = 0.0;
+
+    double p_dc_max = getInverterDCMaxPower(p_dc_rated);
```
mjprilliman, line 117: Added function to get rated DC power from inverter.

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,275 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
+    if (slopeInterpolated < -1) slopeInterpolated = -1;
+
+    // Power in units of W, ratio = max output / rated output
+    ratio += deltaT * slopeInterpolated;
+    if (ratio < 0) ratio = 0.;
+    double p_dc_limit = p_dc_max * ratio;
+    if(p_dc_rated > p_dc_limit) {
+        loss = p_dc_rated - (p_dc_limit);
+        p_dc_rated = p_dc_limit;
+    }
```
mjprilliman, line 188: Check for clipping due to temperature derate.

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,275 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
+    else {
+        loss = 0;
+    }
+}
+
+double SharedInverter::getInverterDCMaxPower(double p_dc_rated)
```
mjprilliman, line 194: New function

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,275 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
-	acWiringLoss_ond_kW = 0.0;
+    // Power quantities go in and come out in units of W
+    double powerDC_Watts = powerDC_kW_in * util::kilowatt_to_watt;
+    double powerAC_Watts = 0.0;
+    Tdry_C = tempC;
+    StringV = DCStringVoltage;
+    double tempLoss = 0.0;
+    double power_ratio = 1.0;
+    if (m_tempEnabled) {
+        calculateTempDerate(DCStringVoltage, tempC, powerDC_Watts, power_ratio, tempLoss);
```
mjprilliman, line 231: Moved derate to the DC side

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,275 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
+   
+    double power_ratio = 1.0;
+    if (m_tempEnabled) {
+        //use average of the DC voltages to pick which temp curve to use- a weighted average might be better but we don't have that information here
+        double avgDCVoltage = 0;
+        double avgDCPower_Watts = 0;
+        for (size_t i = 0; i < powerDC_Watts_one_inv.size(); i++) {
+            //avgDCVoltage += DCStringVoltage[i];
+            power_ratio = 1.0;
+            calculateTempDerate(DCStringVoltage[i], tempC, powerDC_Watts_one_inv[i], power_ratio, tempLoss[i]);
```
mjprilliman, line 285: Again, moved derate to the DC side (iterative)

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,275 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
+
+    // Scale to total system size
+    // Do not need to scale back up by m_numInverters because scaling them down was a separate vector, powerDC_Watts_one_inv
+    powerDC_kW = 0;
+    double tempLoss_avg = 0;
+    for (size_t i = 0; i < powerDC_Watts_one_inv.size(); i++) {
+        powerDC_kW += powerDC_Watts_one_inv[i] * util::watt_to_kilowatt * m_numInverters;
+        tempLoss_avg += tempLoss[i];
+    }
+    tempLoss_avg /= tempLoss.size();
```
mjprilliman, line 305: The only output that is reported per MPPT input is the DC voltage. Thus, temperature loss is reported as an average

shared/lib_shared_inverter.h
```diff
@@ -28,10 +28,13 @@ class SharedInverter
...
 	std::vector<std::vector<double>> getTempDerateCurves();
 
-	/// Modifies pAc, eff, and loss by calculating derate, using curves interpolated by input V
-	void calculateTempDerate(double V, double tempC, double& pAC, double& eff, double& loss);
+	/// Modifies pDc, power ratio of max output to rated output, and loss by calculating derate, using curves interpolated by input V
+	void calculateTempDerate(double V, double tempC, double& p_dc_rated, double& ratio, double& loss);
 
-	/// Given the combined PV plus battery DC power (kW), voltage and ambient T, compute the AC power (kW) for a single inverter with one MPPT input
+    /// Find the Maximum DC power of the inverter at a given time step to use for clipping comparisons
+    double getInverterDCMaxPower(double p_dc_rated);
```
mjprilliman, line 35: Need to fix tabs to spaces. Documentation for new function

ssc/cmod_pvsamv1.cpp
```diff
@@ -671,7 +671,7 @@ static var_info _cm_vtab_pvsamv1[] = {
...
         { SSC_OUTPUT,        SSC_ARRAY,      "inv_psoloss",                          "Inverter power consumption loss",                      "kW",   "",  "Time Series (Inverter)",       "*",                    "",                              "" },
         { SSC_OUTPUT,        SSC_ARRAY,      "inv_pntloss",                          "Inverter night time loss",                             "kW",   "",  "Time Series (Inverter)",       "*",                    "",                              "" },
-        { SSC_OUTPUT,        SSC_ARRAY,      "inv_tdcloss",                       	 "Inverter thermal derate loss",                         "kW",   "",   "Time Series (Inverter)",      "*",             "",                   "" },
+        { SSC_OUTPUT,        SSC_ARRAY,      "inv_tdcloss",                       	 "Inverter thermal derate DC power loss",                "kW",   "",   "Time Series (Inverter)",      "*",             "",                   "" },
```
mjprilliman, line 674: Updated inputs to reflect DC nature of temperature derate in outputs

mjprilliman (commented):
@dguittet Need input on how to handle PVYield inverter derate

shared/lib_pv_io_manager.cpp
dguittet: I think you can throw an exec_error for when that inverter model is selected and temperature derate curves are provided. That inverter model is only accessible via the SDK and so the UI defaults won't be contradictory to throwing that error.

mjprilliman: I understand that these inputs aren't needed for that particular inverter model but don't we still want to allow the model to continue running even if the defaults for the derate curves are inputs? Adding an else loop to get out of the loop before the setTempDerateCurves works for the tests. Would this cause any other errors?

```
if (inverterType == 0) // cec database
    {
        inv_tdc = cm->as_matrix("inv_tdc_cec_db");
    }
    else if (inverterType == 1) // datasheet data
    {
        inv_tdc = cm->as_matrix("inv_tdc_ds");
    }
    else if (inverterType == 2) // partload curve
    {
        inv_tdc = cm->as_matrix("inv_tdc_plc");
    }
    else if (inverterType == 3) // coefficient generator
    {
        inv_tdc = cm->as_matrix("inv_tdc_cec_cg");
    }
    else
    {
        return;
    }
```

dguittet: The default curves could be inputs to this inverter model if the user first used the code generator feature or loaded the defaults in PySAM and then changed the `inverter_model`. But in that case, rather than just ignoring the temp derate curves, I think it's safer from the user impact perspective to throw an error. Then they can remove the curves. If we just don't run the temperature derate, that could be more confusing as to why their simulation results are inconsistent with their inputs.

test/shared_test/lib_shared_inverter_test.cpp
```diff
@@ -7,193 +7,193 @@
...
+        loss = 0.;
+    }
+    void SetUp() {
         sandia.C0 = -2.06147e-07;
         sandia.C1 = 2.7e-05;
         sandia.C2 = 0.002606;
         sandia.C3 = 0.000501;
         sandia.Paco = 59860;
-        sandia.Pdco = 61130.8;
+        sandia.Pdco = 100;
```
dguittet, line 28: Why is this value now so low? This gives a Dc to Ac efficiency of 59860/100.

mjprilliman: I should've changed the pDC starting value rather than the inverter rating. The sandia.Pdco value was much higher than the pDC regardless of the derate with the new approach. I will correct that

mjprilliman: Changed PDC to 61130.8, updated test results to reflect hypothetical system size.

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,272 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
-
-	Tdry_C = tempC;
+    double P_par, P_lr;
+
+    //need to convert to watts and divide power by m_num_inverters
+    std::vector<double> powerDC_Watts_one_inv;
+    std::vector<double> powerDC_Watts_one_inv_iter;
+    for (size_t i = 0; i < powerDC_kW_in.size(); i++) {
+        powerDC_Watts_one_inv.push_back(powerDC_kW_in[i] * util::kilowatt_to_watt / m_numInverters);
+        //powerDC_Watts_one_inv_iter[i] = powerDC_kW_in[i] * util::kilowatt_to_watt / m_numInverters;
```
dguittet, line 267: is this comment required?

shared/lib_shared_inverter.cpp
```diff
@@ -45,232 +45,272 @@ SharedInverter::SharedInverter(const SharedInverter& orig){
...
+    int size = DCStringVoltage.size();
+    std::vector<double> tempLoss(size, 0);
+
+    double power_ratio = 1.0;
+    if (m_tempEnabled) {
+        //use average of the DC voltages to pick which temp curve to use- a weighted average might be better but we don't have that information here
+        double avgDCVoltage = 0;
+        double avgDCPower_Watts = 0;
+        for (size_t i = 0; i < powerDC_Watts_one_inv.size(); i++) {
+            //avgDCVoltage += DCStringVoltage[i];
```
dguittet, line 280: why comment?

dguittet (approved):
Thanks, looks great!

shared/lib_shared_inverter.cpp
mjprilliman: Removed.

shared/lib_shared_inverter.cpp
mjprilliman: Removed.

PR #473, opened 2020-10-28
Added cross axis slope input to irrad class and SubarrayIO

mjprilliman:
-Add cross axis slope (ground slope) as an input to the irradiance object
-Add cross axis slope object in SubarrayIO for accessing var table inputs in pvsamv1
-Slope is currently set to 0 in all other models through added input to irrad.set_surface() (can be exposed if desired)
-Needs testing

fyi @kanderso-nrel The work to get this input added should live here.


mjprilliman:
We need this input to be accessible for shading calculations for fixed and single-axis trackers, maybe more. The integration is not as simple as what is currently in this PR.

Issue #383, opened 2020-06-02
Generalize battery power flow outputs


dguittet:
In the GUI, there's no standalone battery or pairing of batteries with another technology besides PV. In the SDK & PySAM, battery + wind or just battery is possible. The outputs of the battery compute module such as `pv_to_grid`, `pv_to_batt` should be changed to more general terms such as `generator_to_grid`, `generator_to_batt`.

cpaulgilman:
In 7bcb572a02f1b3d3199431a5701eb20d34e6ca84, I changed the variable labels to refer to 'system' instead of 'pv' but did not change the variable names themselves. I'm not sure what name to use, "generator" might suggest a fuel-powered genset, but "system" might appear to include the battery.

dguittet:
Thanks @cpaulgilman glad to know one half of this is done already.

I am also not sure what's the best nomenclature... I am remembering unresolved naming issues around `gen`. In particular how it feeds into the financial models and whether the power going to the `load` was expected or not depends on if it was BTM or FOM-- from the bug @sjanzou resolved. I think we discussed a separate output from `gen` for the case where the power going to the load has been subtracted called `system_to_grid`.

How about `plant_to_grid`? We could also discuss at the meeting.

dguittet:
Working with @brtietz on generic-battery, we noticed it's very confusing that "System Power Generated" is both the power plant + battery since there isn't a separate output for just the power plant/pv.

I think we need to either make a new output for just the power plant/pv or a new one for when battery is included. @cpaulgilman thoughts on naming?

cpaulgilman:
I like gen_with_storage.

cpaulgilman:
https://github.com/NREL/ssc/commit/ce96913e5c228bb96b702211269a8d0c8efd571d fixes SAM383 not SSC 383.

brtietz:
@dguittet - should I take this issue to solve along with https://github.com/NREL/SAM/issues/370? `gen_with_storage` is a good variable name for the "net energy delivered to the grid" variable we discussed at last week's meeting.

dguittet:
@brtietz Yes, that would be awesome!

brtietz:
Closed with https://github.com/NREL/ssc/commit/e96e00f476f0af3dcefe8c5f78d3157fabe66690

PR #474, opened 2020-10-29
Sam 370 correct grid charge expenses

brtietz:
Pairs with SAM PR: https://github.com/NREL/SAM/pull/417

Covers both https://github.com/NREL/ssc/issues/383 and https://github.com/NREL/SAM/issues/370

Changes include:

- Naming changes in ssc vartable names and C++ variables from PV to system
- Add cash flow calculations for positive and negative revenues with electricity rates, solving SAM 370
- Add new gen_without_battery variable to all battery configurations

Single Owner is the only financial model affected, but additional variable names and gen_without_battery will apply to all battery technologies.

Plan is to have extensive review and polishing (especially of names and variable labels) prior to merge. We also need to decide if we only want to apply the PPA cashflow changes Single Owner or to all front of meter modules.


brtietz:
I expect a test or two to fail due to the changing definition of annual energy with the changes in the utility rate code. Once we get that nailed down, I'll update the tests.

cpaulgilman (approved):
Looks good. I created a new issue to update Send To Excel with Equations: https://github.com/NREL/SAM/issues/424.

sjanzou (approved):
As far as I can tell all the "PV" variables have been changes to "System" and the revenue gen values appear correct. 

Beyond this pull request, we should double check the index bounds of the months, and hours when testing for weekday or weekend and month value and then decrementing from 1-12 to 0-11 where used in schedules.

ssc/cmod_pvsamv1.cpp
```diff
@@ -2188,30 +2190,54 @@ void cm_pvsamv1::exec()
...
+                {
+                    if (PVSystem->p_systemACPower[idx] < transformerRatingkW)
+                        xfmr_ll_no_batt *= batt->outGenWithoutBattery[idx] * batt->outGenWithoutBattery[idx] / transformerRatingkW;
+                    else
+                        xfmr_ll_no_batt *= batt->outGenWithoutBattery[idx];
+                }
+                // total load loss
+                ssc_number_t xfmr_loss_no_batt = xfmr_ll + xfmr_nll; // kWh
+                batt->outGenWithoutBattery[idx] -= xfmr_loss_no_batt / ts_hour;
             }
```
dguittet, line 2230: Looks like this code was duplicated from earlier on in the code where it was used for `xfmr_ll`. Could you make a separate fx for calculating transformer losses?

test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -345,14 +345,14 @@ TEST_F(CMPvsamv1BatteryIntegration_cmod_pvsamv1, PPA_ACBatteryModelIntegration)
...
-    ssc_number_t expectedBatteryDischargeEnergy[3] = { 14663, 284336, 284336 };
+    ssc_number_t expectedEnergy[3] = { 37307927, 37306925, 37308021 };
+    ssc_number_t expectedBatteryChargeEnergy[3] = { 14779, 25275, 14779 }; // No rate model means battery use is low
+    ssc_number_t expectedBatteryDischargeEnergy[3] = { 14663, 24062, 14663 };
 
     ssc_number_t peakKwCharge[3] = { -1040.2, -1051.5, -1051.5 };
     ssc_number_t peakKwDischarge[3] = { 967.5, 969.5, 969.5 };
     ssc_number_t peakCycles[3] = { 1, 1, 1 };
-    ssc_number_t avgCycles[3] = { 0.003, 0.2356, 0.2356 };
+    ssc_number_t avgCycles[3] = { 0.003, 0.014, 0.003 };
```
dguittet, line 355: What led to these large changes?

ssc/cmod_pvsamv1.cpp
brtietz: A function is tricky because of lines 2242 - 2255. With a function, annual_xfmr_nll, annual_xfmr_ll, and annual_xfmr_loss now have to become member variables of cm_pvsamv1, and there needs to be a boolean to tell the function whether or not to write to them (is this the actual loss, or the theoretical loss without the battery?) 

I can do the function if you prefer, but was hesitant to introduce the member variables as a part of this change.

test/ssc_test/cmod_battery_pvsamv1_test.cpp
brtietz: This test has grid charging enabled. The root error for these changes was that the code was enabling retail rates when it shouldn't have been. So the changes to cmod_battery increased the price of charging for this test, which decreased the cycles in years 2 & 3. I think the PPA rate escalation was the reason for the large jump in cycles prior to this change, but now the charging costs escalate at the same rate as the discharging revenue.

ssc/cmod_pvsamv1.cpp
dguittet: I agree with not adding unnecessary member variables. 

Would something like the below work? 

```
double transformerLoss(powerkW, transformerRatingkW, LoadLossFraction, NoLoadLossFraction, ts_hour, &xfmr_ll, $xfmr_nll) {
   if transformerRatingkW == 0 return 0

// calculate xfmr_ll, xfmr_nll
   
     if (powerkW < transformerRatingkW)
                        xfmr_ll *= powerKW * powerKW / transformerRatingkW;
                    else
                        xfmr_ll *= powerKW;
                }
                // total load loss
                return xfmr_ll + xfmr_nll; // kWh
}
```

Is line 2228 supposed to be `ssc_number_t xfmr_loss_no_batt = xfmr_ll_no_batt + xfmr_nll_no_batt;` ?

test/ssc_test/cmod_battery_pvsamv1_test.cpp
dguittet: Ok, thanks!

ssc/cmod_pvsamv1.cpp
brtietz: Good catch on line 2228! I implemented your function suggestion in the latest push, thanks!

dguittet: Great! I think it's good to merge as long once Travis is done!

dguittet (approved):

brtietz:
Closed with https://github.com/NREL/ssc/commit/e96e00f476f0af3dcefe8c5f78d3157fabe66690

Issue #472, opened 2020-10-28
Default battery thermal doesn't work for sub-hourly simulations


brtietz:
**Describe the bug**
When attempting to run a detailed pv-battery simulation with a 30 minute weather file, I get the following error:

`exec fail(battery): Environment temperature input length must equal number of weather file records
Simulation pvsamv1 failed :exec fail(battery): Environment temperature input length must equal number of weather file records
`

This is with "Enter single fixed temperature" on the battery thermal page.

**To Reproduce**
1. Start a PV battery simulation
2. Select a subhourly weather file
3. Run simulation

**Expected behavior**
The default battery parameter (enter single fixed temperature) should be able to handle any weather file

brtietz:
This is new to the 2021 release (develop), in 2020.2.29 r3 this problem does not exist.

dguittet:
Currently in develop, `batt_room_temperature_celsius` must have length == nrec. I think this is unnecessary and could have length of 1 as well.

brtietz:
Closed by https://github.com/NREL/SAM/pull/418

PR #475, opened 2020-10-30
add code to resize temperature array from length 1 to length nrec

brtietz:
Add code to cmod_battery to take a battery temperature array from length 1 and resize it to length nrec.

Note that this will not solve https://github.com/NREL/ssc/issues/472 by itself, since the UI is still providing an array of length nrec. See https://github.com/NREL/SAM/pull/418 for details. I'm open to changing things on the SAM side if desired, but figured this code would simplify PySAM inputs.


dguittet (approved):
Thanks! This will definitely improve usability. Could you just add to the meta label in the var_info to say what lengths are acceptable?

PR #479, opened 2020-11-05
Battery Calendar Model set to None results in Cap % > 100

### LINKED ISSUE #478: Battery Calendar Model set to None results in Cap % > 100
**Describe the bug**
When the calendar model is set to None, as in the case of using a default Lead Acid set of inputs, the capacity percent grows above 100%.

**Expected behavior**
If calendar model is None, then calendar degradation should not occur and relative calendar capacity should be constant at 100%.

**Screenshots**
Over 25 years:
image

mjprilliman:
-Add initialization for relative capacity when Calendar setting is set to None
-Closes #478


dguittet (approved):
Awesome, thanks!

PR #480, opened 2020-11-07
Update 6par Module Model Parameter Generator

dguittet:
Fixed the failing modules from @cpaulgilman's helpful spreadsheet by 

1. Increasing the allowable Light Current parameter, Il, from 15 A to 20 A. This fixed the newer, higher power and higher voltage modules, which all had Il around 18 A. It makes sense that the induced current is greater in these modules.

2. Codified original workaround of switching to "multiSi" technology to get failing modules to solve. This works because the "multiSi" technology uses the "default" set of initial guesses, which were created via regression from an old CEC database.

In the future, we may need to update these regressions for generating a working initial guess with the newest, expanded CEC database.

Included a SAM file with each module's IV curves.

[NewModules.zip](https://github.com/NREL/ssc/files/5504895/NewModules.zip)

### DIFF
--- a/shared/6par_solve.h
+++ b/shared/6par_solve.h
@@ -517,7 +517,7 @@ public:
 		Isc = Isc_save;
 
         // try the "default" initial guesses that multiSi uses
-        if (Type != multiSi and err < 0 ) {
+        if (Type != multiSi && err < 0 ) {
             Type = multiSi;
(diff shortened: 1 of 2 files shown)


cpaulgilman (approved):
Thanks Darice! This worked well for high-current and CIGS modules once I fixed the "and" typo in e4b15dfa.

PR #488, opened 2020-11-13
initialize irrad_proc with defaults for optional inputs & check for NaN

### LINKED ISSUE #487: Solarpos_spa fails when pressure is not available from weather data
When the pressure column is missing from the weather file / data, then wf.pres returns NaN. Within the solar position calculation, this NaN pressure propagates to NaN irradiance.

If pressure is not provided, do not use in spa calculations.

dguittet:
close #487


mjprilliman (approved):
This fixed the simulations for two weather files in SAM-documentation that did not have a pressure column.

PR #483, opened 2020-11-11
Sam 372 reframe credit outputs

brtietz:
Reformat excess generation outputs in accordance with https://github.com/NREL/SAM/issues/372

- All "Excess generation" variables with units other than kWh have been removed or changed
- Net metering and net billing have separate outputs (will be zero for the other)
- Only "credits applied" are shown in $ values
- Buy all / sell all sets all excess generation outputs to zero
- Fixed bug where final month's credits in "net billing with carryover to next month" were not being applied. Now applied during the final month (default December)

One open question: should net billing's excess generation change to sum positive (sub) hourly values? https://github.com/NREL/SAM/issues/372#issuecomment-725020561

Test changes are due to removal of some SSC vartable outputs, and the additional credits due to the "net billing with carryover to next month" bugfix. All rates except "net billing with carryover to next month" should have the same monthly & annual bills as before. No changes to the SAM repository were required.

PR #489, opened 2020-11-17
Remove net metering rollover credits output

brtietz:
Remove output added in https://github.com/NREL/ssc/pull/483. After trying to write the help documentation, @cpaulgilman and I determined it will be easier to direct user so the "electricity rate data by tier and period" variables.


cpaulgilman (approved):
Looks good. Made a couple of edits to variable labels in 91afa78c.

PR #490, opened 2020-11-18
ReOpt: use time_steps_per_hour input

dguittet:
Pairs with https://github.com/NREL/SAM/pull/445 to streamline how the SAM-ReOpt integration passes subhourly data


mjprilliman (approved):
I mistakenly claimed that this feature wasn't working when it was. ReOpt still works with these changes.

PR #491, opened 2020-11-18
Price signals dispatch feedback

brtietz:
Two tweaks to the price signals dispatch code based on feedback from our beta testers:

- Update the dispatch planner to better match the powerflow calculations. This allows DC connected systems to charge from PV even if the load is greater than generation at that timestep
- Return the cycle degradation cost in $/cycle-kWh such that the output and input have the same units.


dguittet (approved):
Changes look ok to me. What was the thinking behind charging all the PV for DC connected battery, instead of looking at the net load / generation?

Issue #505, opened 2021-01-08
cmake install target is broken


dangerski:
**Describe the bug**
The cmake install target is broken in the 250 release

**To Reproduce**
Create a cmake project and include ssc as subdirectory, then in your cmake project add an install directive e.g.

Example cmake file
`cmake_minimum_required(VERSION 3.17)`
`project(solartest VERSION 1.0.0)`
`add_executable(solartest "main.cpp")`
`set(SAM_SKIP_TOOLS ON CACHE BOOL "Skips the sdktool and tcsconsole builds" FORCE)`
`set(SAM_SKIP_TESTS ON CACHE BOOL "Skips building tests" FORCE)`
`add_subdirectory(ssc)` # add ssc as subdirectory
`include_directories(ssc/shared ssc/splinter)`
`target_link_libraries(solartest PUBLIC shared ssc)`
Add an install Directive
`install(TARGETS solartest ssc
        PUBLIC_HEADER DESTINATION include
        ARCHIVE DESTINATION lib
        LIBRARY DESTINATION lib
        RUNTIME DESTINATION bin)`

**Expected behavior**
Expect there to be an install target that installs ssc as a lib or any other targets, e.g.

`cmake --build . --target install`

however the result of trying to install results in

> make: *** No rule to make target 'install'.  Stop.

Because of some macros around jsoncpp it kills the install target with no warning.  If you look at the make file created for your project, no install target is made. 

 - OS: OSX, Ubuntu
 - Version 250

I can submit a pull request to fix it.

PR #506, opened 2021-01-08
Fix for SSC issue #505

dangerski:
Fix SSC issue #505 that breaks the install targets of projects that include this project.  EXCLUDE_FROM_ALL will prevent the jsoncpp install target without breaking other projects.


dguittet (approved):
Works, thank you for for fixing the hack to get around jsoncpp install!

PR #507, opened 2021-01-11
496 update required vars

brtietz:
Update batt_cycle_cost to only be required when batt_cycle_cost_choice==1

Tested by removing batt_cycle_cost from the setup of CMBattery_cmod_battery


dguittet:
Could you take a look at the failing tests?

dguittet (approved):
Thanks!

PR #508, opened 2021-01-11
Fix time series buy rates

brtietz:
The if statement for time series buy rates re-initialized the variables cnt and ts_br, such that the values read out of the vartable never make it to the setup_rate_data function. Thus, the length of rate.m_ec_ts_buy_rate was always 0, and the flat rates were used instead.

Can be tested with the SAM file attached to https://github.com/NREL/SAM/issues/494, closes https://github.com/NREL/SAM/issues/494


cpaulgilman (approved):

sjanzou (commented):
Looks good.

PR #509, opened 2021-01-13
Fixed bug in atmospheric_refraction_correction

dangerski:
Storing a double in an int and getting zero.


mikofski:
Surprising that tests still pass. Is there a unit test that should capture this?

dangerski:
There was a comment about investigating some potential rounding issues in some tests. May be related but not sure.

mjprilliman (approved):
The unit tests did not contain an example where the difference in the SUN_RADIUS value affected whether the atmospheric refraction correction was performed or not. A unit test was added that checks for this behavior (i.e. the difference in int and double storage affects the result of the test). 

Thanks for the catch!

PR #510, opened 2021-01-13
Price signals dispatch improved demand

brtietz:
Incorporates several improvements to price signals dispatch to better handle demand charges.

Previously, the price signals dispatch algorithm would ignore peaks on the last day of each month in order to dispatch at midnight, due to a perceived high price for each month's initial load. The forecast function was using net load to forecast demand charges, so the forecast load was lower than typical night time loads. Increase the prediction to average gross load, so that the algorithm waits for higher peaks.

Additionally, anticipate demand charges in future hours by planning additional dispatch in hours with higher net grid use than the highest cost hour. Previously, the 2nd hour of high power would show a much lower cost than the first, leading the algorithm to plan a smaller dispatch, but still seeing a high cost when the cost of the dispatch plan is computed. This approach compensates by looking at the power demand, and planning additional dispatch if the power required is equal to or greater than the highest cost period.

Finally, refactor the cost forecast function to ensure energy charges are handled consistently between 1 hour and multi hour forecasts.

Increases NPV 3.7% for a shopping center load profile.

### DIFF
--- a/shared/lib_utility_rate.cpp
+++ b/shared/lib_utility_rate.cpp
@@ -198,10 +198,7 @@ double UtilityRateForecast::forecastCost(std::vector<double>& predicted_loads, s
 	if (crossing_month)
 	{
         initializeMonth(month_at_end, year_at_end);
-        if (n > 1)
-        {
-            previousDemandCharge += rate->get_demand_charge(month_at_end, year_at_end);
-        }
+        previousDemandCharge += rate->get_demand_charge(month_at_end, year_at_end);
 	}
 
     double newEnergyCharge = 0;
@@ -248,9 +245,20 @@ double UtilityRateForecast::forecastCost(std::vector<double>& predicted_loads, s
 		}
 	}
 
-	// Compute new peak cost - may need to run two months
-	double newDemandCharge = rate->get_demand_charge(month, year);
-	if (crossing_month && n > 1)
+    // Compute new peak cost - may need to run two months
+    double newDemandCharge = rate->get_demand_charge(month, year);
+    // If forecast length is 1, restartMonth won't be triggered on the next forecast. Trigger it now
+    if (crossing_month && n == 1)
+    {
+        if (rate->enable_nm)
+        {
+            newEnergyCharge += getEnergyChargeNetMetering(month, current_composite_buy_rates, current_composite_sell_rates);
+        }
+        restartMonth(month, month_at_end, year_at_end);
+        copyTOUForecast();
+    }
+
+	if (crossing_month)
 	{
 		newDemandCharge += rate->get_demand_charge(month_at_end, year_at_end);
         if (rate->enable_nm)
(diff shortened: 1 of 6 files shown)


test/shared_test/lib_utility_rate_test.cpp
```diff
@@ -183,8 +183,33 @@ TEST(lib_utility_rate_test, test_demand_charges_crossing_months)
 	int hour_of_year = 742; // 10 pm on Jan 31st
 	double cost = rate_forecast.forecastCost(forecast, 0, hour_of_year, 0);
 
-	// Total cost for the months would be $1511.25, but this subtracts off the peaks predicted by average load ($3.12)
-	ASSERT_NEAR(1508.11, cost, 0.02);
+	// Total cost for the months would be $1511.25, but this subtracts off the peaks predicted by average load ($1500)
+	ASSERT_NEAR(11.25, cost, 0.02);
```
dguittet, line 187: do you mean peaks predicted by gross load instead of "peaks predicted by average load"? Seems like with the forecast baseline using the gross peak instead of the net peak for demand, the difference of 11.25 here is due primarily to energy charges freshly incurred

shared/lib_utility_rate.cpp
```diff
@@ -143,7 +143,7 @@ UtilityRateForecast::UtilityRateForecast(rate_data* util_rate, size_t stepsPerHo
 	rate = std::shared_ptr<rate_data>(new rate_data(*util_rate));
 	m_monthly_load_forecast = monthly_load_forecast;
 	m_monthly_gen_forecast = monthly_gen_forecast;
-	m_monthly_peak_forecast = monthly_peak_forecast;
+	m_monthly_avg_load_forecast = monthly_avg_load_forecast;
```
dguittet, line 146: While I see how going from max net peak to average gross load for the forecast's baseline assumption for demand makes sense, does the average gross load actually work better than max gross peak?

Also, in all the tests the input to the constructor for the forecast uses the name `monthly_peak_forecast` instead of `monthly_avg_load_forecast`.

Depending on the shape of the load, the average gross load can be very different from the max gross peak (imagine a single peak vs a uniform load).

test/shared_test/lib_utility_rate_test.cpp
brtietz: I think the most accurate statement would be "average gross load" - I'll update the comment. Yes, the $11.25 is due to energy charges.

shared/lib_utility_rate.cpp
brtietz: The m_monthly_peak_forecast was never used, and I neglected to remove it in my final code cleanup. I went through and renamed the variables in the code for this update, I'll rename the tests shortly. The previous code was using average net.

The purpose of that line is to prevent the first hour(s) of each month from looking like a peak if they are below the average load. I'm not sure what the algorithm would do with max gross or max net. I could imagine setting this variable to (max net - battery power) but that means the price forecast would need to be aware of the battery power, which seems convoluted. Additionally, that value could end up being lower than the typical nightly load - the case I was working with the max net peak was 625 kW, the battery power was 200 kW, and the typical nightly load was 440 kW, so there would still be some degree of predicted demand charge on the first hour of each month with that approach. Average gross load was ~490 kW, and average net was ~350 kW.

dguittet: Okay it sounds like the baseline you're looking for in this case is the some lower estimate of demand peak after battery dispatch-- therefore you can compare how much different dispatch plans __increase__ the demand peak. In my mind, I was thinking of the baseline being the original demand peak, whereby you would compare how much different dispatch plans __decrease__ the demand peak. IIRC (but haven't looked at this in some time) that's similar to what the peak shaving algorithm does in sorting the peaks by highest existing peak.

dguittet (approved):

PR #512, opened 2021-01-14
Unhandled Divide by Zero Error

### LINKED ISSUE #511: Unhandled Divide by Zero Error
**Describe the bug**
PV models with tilt = 0 will result in dividing by zero in the sssky_diffuse_table since tand_stilt and tind_stilt are 0. https://github.com/NREL/ssc/blob/develop/shared/lib_pvshade.cpp#L305

On most tested platforms, this produces a NaN will gets caught in line 309, however on Linux Debug, this was throwing an Arithmetic error. 

**Expected behavior**
The NaN should be generated explicitly and then caught.

dguittet:
Fixes unhandled divide by zero error when pv tilt = 0 for some builds

close #511

### DIFF
--- a/shared/lib_pvshade.cpp
+++ b/shared/lib_pvshade.cpp
@@ -302,7 +302,10 @@ double sssky_diffuse_table::compute(double surface_tilt) {
     double Asky_shade[1000];
     for (int n = 0; n < 1000; n++)
     {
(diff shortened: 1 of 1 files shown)


mjprilliman (approved):
Was not able to test in Linux but the changes make sense to me.

PR #514, opened 2021-01-16
Batt cutoff voltage

mjprilliman:
-Modify battery compute modules to consider slightly larger Qmax when calculating max discharge current, dynamic voltage model parameters
-Update tests, add test with voltage cutoff to show impact on results


mjprilliman:
@dguittet @brtietz It would be helpful if you could at least skim this before we meet tomorrow just to have an idea of where things stand. Also for the same branch name in SAM there is a PR.

ssc/cmod_battery.cpp
```diff
@@ -91,11 +91,14 @@ var_info vtab_battery_inputs[] = {
         { SSC_INPUT,        SSC_NUMBER,      "batt_Vnom",                                  "Cell voltage at end of nominal zone",                     "V",       "",                     "BatteryCell",       "",                           "",                              "" },
         { SSC_INPUT,        SSC_NUMBER,      "batt_Vnom_default",                          "Default nominal cell voltage",                            "V",       "",                     "BatteryCell",       "",                           "",                              "" },
         { SSC_INPUT,        SSC_NUMBER,      "batt_Qfull",                                 "Fully charged cell capacity",                             "Ah",      "",                     "BatteryCell",       "",                           "",                              "" },
+        { SSC_INPUT,        SSC_NUMBER,      "batt_Qfull_cutoff",                                 "Fully charged cell capacity",                             "Ah",      "",                     "BatteryCell",       "",                           "",                              "" },
```
brtietz, line 94: Can you change the description to explain the difference between Qfull and Qfull_cutoff?

ssc/cmod_battery.cpp
```diff
@@ -91,11 +91,14 @@ var_info vtab_battery_inputs[] = {
...
         { SSC_INPUT,        SSC_NUMBER,      "batt_Qnom",                                  "Cell capacity at end of nominal zone",                    "Ah",      "",                     "BatteryCell",       "",                           "",                              "" },
         { SSC_INPUT,        SSC_NUMBER,      "batt_C_rate",                                "Rate at which voltage vs. capacity curve input",          "",        "",                     "BatteryCell",       "",                           "",                              "" },
         { SSC_INPUT,        SSC_NUMBER,      "batt_resistance",                            "Internal resistance",                                     "Ohm",     "",                     "BatteryCell",       "",                           "",                              "" },
+        { SSC_INPUT,        SSC_NUMBER,      "batt_voltage_cutoff",                        "Cutoff voltage for battery rated capacity",               "V",       "",                     "BatteryCell",       "",                           "",                              "" },
```
brtietz, line 101: Should this be "?=0" to maintain backwards compatibility with previous versions?

ssc/cmod_battery.cpp
brtietz: We should talk about whether this is computed in SSC or SAM. It might be easier to maintain consistency between batt stateful and cmod battery if SSC does the computation.

brtietz (changes requested):

shared/lib_battery_voltage.cpp
```diff
@@ -332,7 +334,12 @@ void voltage_dynamic_t::parameter_compute() {
     _K = ((params->dynamic.Vfull - params->dynamic.Vnom + _A * (std::exp(-_B0 * params->dynamic.Qnom) - 1)) *
           (params->dynamic.Qfull - params->dynamic.Qnom)) / (params->dynamic.Qnom); // [V] - polarization voltage
     _E0 = params->dynamic.Vfull + _K + params->resistance * I - _A;
-
+    if (params->dynamic.Vcut != 0) {
+        double C = (-1 * params->dynamic.Vcut + _E0 - params->resistance * I + _A * (std::exp(-_B0 * params->dynamic.Qfull))) / _K;
+        double x = params->dynamic.Qfull / (C - 1);
+        //params->dynamic.Qfull += x;
+        params->dynamic.Qfull_mod = params->dynamic.Qfull_mod + x;
```
brtietz, line 341: What happens if this is called multiple times? This happens in the copy constructor. Will Qfull_mod increase over the simulation?

shared/lib_battery_voltage.cpp
```diff
@@ -348,6 +355,12 @@ void voltage_dynamic_t::set_initial_SOC(double init_soc) {
 // everything in here is on a per-cell basis
 double voltage_dynamic_t::voltage_model_tremblay_hybrid(double Q_cell, double I, double q0_cell) {
     double it = Q_cell - q0_cell;
+    /*
+    double C = (-1 * params->dynamic.Vcut + _E0 - params->resistance * I + _A * exp(-_B0 * Q_cell)) / _K;
+    double x = Q_cell / (C - 1);
+    double Q_cell_mod = Q_cell + x;
+    //double E = _E0 - _K * (Q_cell / (Q_cell - it)) + _A * exp(-_B0 * it);
+    */
     double E = _E0 - _K * (Q_cell / (Q_cell - it)) + _A * exp(-_B0 * it);
```
brtietz, line 364: Should Qfull_mod be used when calculating the voltage?

brtietz (changes requested):
I'm not seeing the behavior I expect at this time. The GUI looks good in the SAM PR, but the voltage is too low at low SOC in the attached SAM file. I think Qfull_mod needs to be used more broadly in the voltage calculations in SSC.

[voltage_test_single_series.zip](https://github.com/NREL/ssc/files/5858535/voltage_test_single_series.zip)

shared/lib_battery_voltage.cpp
```diff
@@ -342,13 +350,30 @@ void voltage_dynamic_t::parameter_compute() {
...
+    else {
+        Q_cell_mod = Q_cell;
+        params->dynamic.Qfull_mod = 0;
+        
+    }
+    //->dynamic.Qfull_mod = Q_cell_mod;
+    //double E = _E0 - _K * (Q_cell / (Q_cell - it)) + _A * exp(-_B0 * it);
+    
+    double E = _E0 - _K * (Q_cell_mod / (Q_cell_mod - it)) + _A * exp(-_B0 * it);
+    //double E = _E0 - _K * (params->dynamic.Qfull_mod / (params->dynamic.Qfull_mod - it)) + _A * exp(-_B0 * it);
```
mjprilliman, line 376: @brtietz @dguittet Ended up putting the modification to Qfull here to account for the potential change in capacity due to the lifetime and thermal models. If Qfull_mod is calculated in parameter_compute and not adjusted for changes in maximum battery capacity, the maximum possible discharge will be greater than what is actual possible. Ran into this issue when existing tests with 0 cutoff voltage were failing due to a difference between the qmax that was originally passed and the non-changing Qfull_mod. Does this track?

test/shared_test/lib_battery_voltage_test.cpp
```diff
@@ -291,34 +295,155 @@ TEST_F(voltage_dynamic_lib_battery_voltage_test, calculateMaxDischargeSubMinute)
...
+    double dt_hour = 0.5;
+    CreateModel(dt_hour);
+
+    // start at half SOC
+    double max_current;
+    double power = model2->calculate_max_discharge_w(cap2->q0(), cap2->qmax(), 0, &max_current);
+    EXPECT_NEAR(power, 4669, 1);        // current ~8
+    double max_current_calc = model2->calculate_current_for_target_w(power, cap2->q0(), cap2->qmax(), 0);
+    EXPECT_NEAR(max_current_calc, max_current, 2);
+    // Does not empty battery for highest power
```
mjprilliman, line 384: @brtietz @dguittet Running into some issues with calculate_current_for_target_w when creating new tests with non-zero voltage cutoffs. The solver this function uses is not aware of the change to Qfull, and my attempts to modify it have not improved the accuracy as of yet between the current results (notice the 2 amp difference between max_current_calc and max_current here).

shared/lib_battery_voltage.cpp
brtietz: Yes, Q_cell_mod should be adjusted for thermal and lifetime changes, which means it should be recomputed on each call to this function.

test/shared_test/lib_battery_voltage_test.cpp
brtietz: I think this function should be aware of Q_full_mod and/or V_cut. Is it possible to store them in voltage_params and pull them out again inside `calculate_current_for_target_w`?

Happy to meet to discuss, send me an invite if you think that's helpful (please avoid Wednesday if possible).

shared/lib_battery_voltage.cpp
```diff
@@ -275,8 +275,8 @@ void voltage_dynamic_t::initialize() {
 }
 
 voltage_dynamic_t::voltage_dynamic_t(int num_cells_series, int num_strings, double voltage, double Vfull,
-                                     double Vexp, double Vnom, double Qfull, double Qexp, double Qnom,
-                                     double C_rate, double R, double dt_hr) :
+                                     double Vexp, double Vnom, double Vcut, double Qfull, double Qexp, double Qnom,
+                                     double C_rate, double R, double dt_hr ) :
```
dguittet, line 279: I would put the Vcut after Qnom since it's not part of the exp, nom, full pairs between V and Q.

shared/lib_battery_voltage.cpp
dguittet: Looks right to recalculate it per call given how Q_cell changes. I think that mean the `Q_full_mod` should not be within the `params` struct, since those are values that are consistent throughout the simulation. It should be within the `state` struct instead

shared/lib_battery_voltage.cpp
```diff
@@ -379,37 +405,49 @@ double voltage_dynamic_t::calculate_max_charge_w(double q, double qmax, double,
...
         if (p > max_p) {
             max_p = p;
             max_I = current;
+            max_V = vol;
+            if (params->dynamic.Vcut != 0) {
+                C = (-1 * params->dynamic.Vcut + _E0 - params->resistance * q - current * params->dt_hr + _A * exp(-_B0 * qmax)) / _K;
+                x = qmax / (C - 1);
+                params->dynamic.Qfull_mod = x;
+                 
+            }
```
dguittet, line 431: Isn't this now moved into `voltage_model_tremblay_hybrid`?

This function should not make any changes to `params` or `state`

ssc/cmod_battery_stateful.cpp
```diff
@@ -389,22 +391,25 @@ std::shared_ptr<battery_params> create_battery_params(var_table *vt, double dt_h
...
     m_vartab = vt;
     try {
         if (!compute_module::verify("precheck input", SSC_INPUT))
             throw exec_error("battery_stateful", log(0)->text);
         dt_hr = as_number("dt_hr");
         control_mode = as_integer("control_mode");
-        params = create_battery_params(m_vartab, dt_hr);
+        params = create_battery_params(m_vartab, dt_hour);
+        if ((params->voltage->dynamic.Vcut > params->voltage->dynamic.Vnom) || (params->voltage->dynamic.Vnom > params->voltage->dynamic.Vexp) || (params->voltage->dynamic.Vexp > params->voltage->dynamic.Vfull)) {
+            throw exec_error("battery_stateful", "Vfull must be greater than Vexp, which must be greater than Vnom, which must be greater than Vcut");
```
dguittet, line 411: Can you put these error checks and the other voltage checks from cmod_battery into the voltage_dynamic_t constructor? See `initialize`

shared/lib_battery_voltage.cpp
dguittet: However, `voltage_model_tremblay_hybrid` should not update the `state` since this function is used in many "prospective" functions such as all the calculate_x functions. These prospective functions are used to query what is a possible next state of the battery without actually moving the battery state to any of them. The only function that can "step forward" the battery is `updateVoltage`.

```
    double Vfull = 4.1;
    double Vexp = 4.05;
    double Vnom = 3.4;
    double Qfull = 2.25;
    double Qexp = 0.04;
    double Qnom = 2.0;
    double C_rate = 0.2;
```

This gives us 3 points on the V vs Q curve: (V_full, 0), (V_exp, Q_full - Q_exp), (V_nom, Q_full - Q_nom).

Now, however, we have 4 different points: (V_full, 0), (V_exp, Q_full - Q_exp), (V_nom, Q_full - Q_nom), (V_cutoff, Q_full)

Once the parameters have been generated, we calculate an additional "parameter" Q_full_mod.

When the Q_full is decreased due to thermal or lifetime, the point (V_cutoff, Q_full) may no longer hold as we'll reach the same V_cutoff but with less Q_full. And so `Q_full_mod` needs to be updated whenever the battery state is changed.

This update of Q_full and Q_full_mod can occur just in `updateVoltage`, before calling `voltage_model_tremblay_hybrid`. The latter should take Q_full_mod as given. All the other calculate_x functions should do the same.

For `calculate_max_discharge_w` the reason we have to do the while loop was because the highest current (I = q) doesn't necessarily produce the highest discharge power. So the loop iterated from I = 0 to q0 * dt, and I think that's still the same now since q0 <= Q_full < Q_full_mod, but the `voltage_model_tremblay_hybrid` will be using Q_full_mod.

For `calculate_current_for_target_w` the `solve_current_for_discharge_power` and `solve_current_for_charge_power` may need some way to enforce the V_cutoff if it's not matching up with `calculate_max_discharge_w`. I'll have to think a bit more later.

ssc/cmod_battery_stateful.cpp
mjprilliman: I added Vcut to the checks in the constructor but there was already one there. Do you mean instead of the checks in cmod_battery and battery_stateful or in addition to?

test/shared_test/lib_battery_test.h
```diff
@@ -257,8 +262,8 @@ class lib_battery_test : public ::testing::Test
         dtHour = 1.0;
 
         capacityModel = new capacity_lithium_ion_t(q, SOC_init, SOC_max, SOC_min, dtHour);
-        voltageModel = new voltage_dynamic_t(n_series, n_strings, Vnom_default, Vfull, Vexp, Vnom, Qfull, Qexp, Qnom,
-                                             C_rate, resistance, dtHour);
+        voltageModel = new voltage_dynamic_t(n_series, n_strings, Vnom_default, Vfull, Vexp, Vnom, Qfull, Qexp, Qnom, Vcut, 
+                                             C_rate, resistance, dtHour );
```
mjprilliman, line 266: With the code changes there are always going to be issues when the q value used to initialize the capacity (q = 1000) is not tied to the cell capacity and number of strings (Qfull = 2.25, n_strings = 9).

test/ssc_test/cmod_battery_stateful_test.h
```diff
@@ -17,7 +17,7 @@ class CMBatteryStatefulIntegration_cmod_battery_stateful : public ::testing::Tes
...
....000, "maximum_SOC": 95.000, "minimum_SOC": 5.000, "dt_hr": 1.000, "leadacid_tn": 0.000, "leadacid_qn": 0.000, "leadacid_q10": 0.000, "leadacid_q20": 0.000, "voltage_choice": 0, "Vnom_default": 3.600, "resistance": 0.000, "Vfull": 4.100, "Vexp": 4.050, "Vnom": 3.400, "Vcut": 0.000, "Qfull": 2.250, "Qexp": 0.040, "Qnom": 2.000, "C_rate": 0.200, "mass": 507.000, "surface_area": 2.018, "Cp": 1004.000, "h": 20.000, "cap_vs_temp": [ [ -10, 60 ], [ 0, 80 ], [ 25, 1E+2 ], [ 40, 1E+2 ] ], "option": 1, "T_room_init": 20, "cycling_matrix": [ [ 20, 0, 1E+2 ], [ 20, 5E+3, 80 ], [ 20, 1E+4, 60 ], [ 80, 0, 1E+2 ], [ 80, 1E+3, 80 ], [ 80, 2E+3, 60 ] ], "calendar_choice": 1, "calendar_q0": 1.020, "calendar_a": 0.003, "calendar_b": -7280.000, "calendar_c": 930.000, "calendar_matrix": [ [ -3.1E+231 ] ], "loss_choice": 0, "monthly_charge_loss": [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 ], "monthly_discharge_loss": [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 ], "monthly_idle_loss": [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 ], "schedule_loss": [], "replacement_option": 0, "replacement_capacity": 0.000, "replacement_schedule": [], "replacement_schedule_percent": [], "analysis_period": 1, "load_escalation": [0]})";
```
mjprilliman, line 20: Not sure why where the issue is in battery_stateful (AdaptiveTimestep and HourlyvsSubHourly failing). @dguittet @brtietz I could use some guidance on the AdaptiveTimestep tests.

shared/lib_battery_voltage.cpp
mjprilliman: Removed.

shared/lib_battery_voltage.cpp
```diff
@@ -379,37 +405,49 @@ double voltage_dynamic_t::calculate_max_charge_w(double q, double qmax, double,
...
 using namespace std::placeholders;
 
-double voltage_dynamic_t::calculate_max_discharge_w(double q, double qmax, double, double *max_current) {
+double voltage_dynamic_t::calculate_max_discharge_w(double q, double qmax, double kelvin, double *max_current) {
     q /= params->num_strings;
     qmax /= params->num_strings;
 
-    double current = 0., vol = 0;
+    double current = 0.;
+    double vol = params->dynamic.Vcut + 0.012;
```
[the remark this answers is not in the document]
mjprilliman: Leftover from a misunderstanding of how the while loop was operating. Removed.

shared/lib_battery_voltage.cpp
```diff
@@ -367,9 +392,10 @@ void voltage_dynamic_t::updateVoltage(double q, double qmax, double I, const dou
     state->cell_voltage = fmax(voltage_model_tremblay_hybrid(qmax, I, q), 0);
 }
 
-double voltage_dynamic_t::calculate_max_charge_w(double q, double qmax, double, double *max_current) {
+double voltage_dynamic_t::calculate_max_charge_w(double q, double qmax, double kelvin, double *max_current) {
```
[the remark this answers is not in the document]
mjprilliman: Corrected.

shared/lib_battery_voltage.cpp
mjprilliman: Moved Vcut after Qnom, fixed in all voltage_dynamic_t calls.

shared/lib_battery_voltage.cpp
```diff
@@ -440,8 +459,9 @@ void voltage_dynamic_t::solve_current_for_charge_power(const double *x, double *
 
 void voltage_dynamic_t::solve_current_for_discharge_power(const double *x, double *f) {
     double I = x[0];
-    double V = _E0 - _K * solver_Q / (solver_q - I * params->dt_hr) +
-               _A * exp(-_B0 * (solver_Q - (solver_q - I * params->dt_hr))) - params->resistance * I;
+    //double V = _E0 - _K * solver_Q / (solver_q - I * params->dt_hr) +
+    //         _A * exp(-_B0 * (solver_Q - (solver_q - I * params->dt_hr))) - params->resistance * I;
+    double V = _E0 - _K * state->Q_full_mod / (state->Q_full_mod - (solver_Q - (solver_q - I * params->dt_hr))) + _A * exp(-_B0 * (solver_Q - (solver_q - I * params->dt_hr))) - params->resistance * I;
```
brtietz, line 464: What is the relationship between solver_Q and Q_full_mod? Why was it changed here but not in solve_current_for_charge_power?

My hunch is this line is the source of the testing woes, but I haven't studied it enough to suggest a solution.

ssc/cmod_battery_stateful.cpp
brtietz: Instead of - if the checks are in voltage_dynamic_t both cmods will run them, rather than having to have the code in two places.

test/input_cases/generic_common_data.h
```diff
@@ -79,6 +79,8 @@ void generic_singleowner_battery_60min(ssc_data_t &data)
 	ssc_data_set_number( data, "batt_Vfull", 3.5999999046325684 );
 	ssc_data_set_number( data, "batt_Vexp", 3.4000000953674316 );
 	ssc_data_set_number( data, "batt_Vnom", 3.2999999523162842 );
+    //ssc_data_set_number(data, "batt_Vcut", 0.66 * 3.5999999046325684);
```
brtietz, line 82: Is this commented out line still needed?

shared/lib_battery_voltage.cpp
```diff
@@ -428,20 +452,20 @@ double voltage_dynamic_t::calculate_current_for_target_w(double P_watts, double
...
     newton<double, std::function<void(const double *, double *)>, 1>(x, resid, check, f,
                                                                      100, 1e-6, 1e-6, 0.7);
+
     return x[0] * params->num_strings * direction;
 }
 
 void voltage_dynamic_t::solve_current_for_charge_power(const double *x, double *f) {
     double I = x[0];
-    double V = _E0 - _K * solver_Q / (solver_q + I * params->dt_hr) +
+    double V = _E0 - _K * solver_Q_mod / (solver_Q_mod - (solver_Q - (solver_q + I * params->dt_hr))) +
```
brtietz, line 461: I'm mostly curious, why isn't this `_K * solver_Q_mod / (solver_q + I * params->dt_hr)`? Additionally, why is solver_Q in the exponent unchanged?

test/shared_test/lib_battery_dispatch_automatic_btm_test.cpp
```diff
@@ -300,7 +300,7 @@ TEST_F(AutoBTMTest_lib_battery_dispatch, TestBasicForecast) {
     batteryPower->connectionMode = ChargeController::AC_CONNECTED;
 
     // Discharge first 4 hours to avoid peak demand charges. Charge while there's solar, then discharge again at 6 pm since this is a high TOU rate
-    std::vector<double> expectedPower = { 50, 50, 50, 34.79, 0, 0, 0, -47.84, -47.84, -47.84, -47.84, -47.84, -47.84, -47.84, -47.84, -47.91, -10.538, 0, 50, 50, 50,
+    std::vector<double> expectedPower = { 50, 50, 50, 37.75, 0, 0, 0, -47.87, -47.97, -47.90, -47.90, -47.90, -47.90, -47.90, -47.90, -47.90, -11.87, 0, 50, 50, 50,
```
brtietz, line 303: Increased power due to the higher voltage at low SOC makes total sense! This is great!

dguittet (approved):

PR #518, opened 2021-01-25
fix battery capacity in lib_battery_dispatch_automatic_btm_test.h

dguittet:
Battery capacity was set to a separate variable that made the capacity and voltage model have different capacities.

Modify capacity model to use the same value, adjust the expected values.


mjprilliman (approved):

mjprilliman:
@dguittet looks like the same line was copied into lib_battery_dispatch_manual_test.h (affects ManualTest_lib_battery_dispatch tests).

dguittet:
Thanks! Just fixed those tests also.

mjprilliman (approved):

PR #522, opened 2021-01-28
Set initial SOC for voltage with battery capacity rather than cell

### LINKED ISSUE #520: Dynamic voltage model set_initial_SOC based on cell capacity rather than system capacity
**Describe the bug**
The initial voltage calculation for the battery is based on the cell capacity (Qfull) rather than the system capacity (Qfull * n_strings)

image

image

mjprilliman:
-Change set_initial_SOC for dynamic voltage model to use battery capacity rather than cell capacity, as updateVoltage() divides by the number of strings to get to cell capacity. 
-Closes #520

### DIFF
--- a/shared/lib_battery_voltage.cpp
+++ b/shared/lib_battery_voltage.cpp
@@ -342,7 +342,7 @@ void voltage_dynamic_t::parameter_compute() {
 }
 
 void voltage_dynamic_t::set_initial_SOC(double init_soc) {
-    updateVoltage(init_soc * 0.01 * params->dynamic.Qfull, params->dynamic.Qfull, 0, 25, params->dt_hr);
+    updateVoltage(init_soc * 0.01 * params->dynamic.Qfull * params->num_strings, params->dynamic.Qfull * params->num_strings, 0, 25, params->dt_hr);
 }
 
 // everything in here is on a per-cell basis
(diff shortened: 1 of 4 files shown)


shared/lib_battery_voltage.cpp
```diff
@@ -342,7 +342,7 @@ void voltage_dynamic_t::parameter_compute() {
 }
 
 void voltage_dynamic_t::set_initial_SOC(double init_soc) {
-    updateVoltage(init_soc * 0.01 * params->dynamic.Qfull, params->dynamic.Qfull, 0, 25, params->dt_hr);
+    updateVoltage(init_soc * 0.01 * params->dynamic.Qfull * params->num_strings, params->dynamic.Qfull * params->num_strings, 0, 25, params->dt_hr);
```
brtietz, line 345: Please add comments to the header file with the final units convention.

I did a "find all references" check for this function - should lib_battery_test be updated as well?

brtietz (changes requested):

shared/lib_battery_voltage.cpp
mjprilliman: See latest commit to see if comments are sufficient. 

On the second point in lib_battery_test the inputs to updateVoltage come from the capacity model functions that return correct system capacity inputs (confirmed in debugger).

brtietz (approved):
Thanks for adding the comment!

PR #524, opened 2021-01-28
Adaptive battery timesteps

### LINKED ISSUE #523: Adaptive battery timesteps
The current `battery_t` uses the timestep variable `dt_hr` as a fixed parameter that cannot change within a simulation. 

Allow the `battery_t` timestep to be changed from the interface, and make sure all the component models are updated with the new timestep.

Alternatives are to construct a new battery with the same state and parameters save `dt_hr`. This will be more computationally inefficient.

This is required for the battery model to be used inside EnergyPlus' power & heat convergence loops which use adaptive timesteps.

dguittet:
Add ChangeTimestep function to `battery_t`

Rename all `dt_hour` vars to `dt_hr`

Close #523

### DIFF
--- a/shared/lib_battery.cpp
+++ b/shared/lib_battery.cpp
@@ -48,7 +48,7 @@ void thermal_t::initialize() {
     state->T_batt_prev = state->T_room;
     state->heat_dissipated = 0;
     state->q_relative_thermal = 100;
-    dt_sec = params->dt_hour * 3600;
+    dt_sec = params->dt_hr * 3600;
 }
 
 thermal_t::thermal_t(double dt_hour, double mass, double surface_area, double R, double Cp, double h,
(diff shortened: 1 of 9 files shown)


brtietz:
The interface with battery_t looks good, but I was expecting a call to "ChangeTimestep" somewhere in battery stateful. I'd prefer to collect all of the changes to dt_hr in that function for futureproofing. I'm also missing where `params->capacity->dt_hr` is adjusted in `create_battery_params`

dguittet:
Good point that I should add this to battery stateful so it's useable externally.

dguittet:
Good catch on the missing `dt_hr` initialization

dguittet:
@brtietz ready for re-review

brtietz (approved):
Looks great, thanks!

PR #525, opened 2021-01-28
Sam492 fix doublecounted credits

brtietz:
When debugging https://github.com/NREL/SAM/issues/492 Paul noticed some values not adding up. As it turns out, the previous code credited the net billing "true-up" month twice if the energy charges were negative, since monthly_cumulative_excess_dollars[m] was applied to both months. Correct for this, and also allow dollars_applied to be greater than monthly_ec_charges_gross to represent the cash out of the credits at the end of the 12 month period.

@cpaulgilman I wasn't able to get the net billing credit to match the true-up check exactly, since the net billing credit is also applied to that month's energy charges:

image

However, `net_billing_credit[m] - energy_charges_before_credits[m]` will supply the value that the macro needs.


cpaulgilman:
Thanks Brian! I need to redo some of the Value of RE macro calculations for [SAM/NREL#492](https://github.com/NREL/SAM/issues/492) to emphasize when dollar credits are being applied to the energy charge portion of the bill rather than to the entire bill. I will do that before reviewing this in case something else comes up in that process.

Do you plan to address the net metering $ credits issue (put true-up amount in "net metering credit column") on this branch or a separate one?

Thanks,
Paul.

ssc/cmod_utilityrate5.cpp
```diff
@@ -1798,8 +1796,15 @@ class cm_utilityrate5 : public compute_module
 						if (excess_monthly_dollars || rollover_credit)
 						{
 //							monthly_cumulative_excess_dollars[m] = -monthly_bill[m];
-							dollars_applied += monthly_cumulative_excess_dollars[m] + monthly_ec_charges[m];
-							monthly_cumulative_excess_dollars[m] = -monthly_ec_charges[m];
+                            if (m != net_metering_credit_month)
```
sjanzou, line 1799: Is this balancing out in the utility bill reported by SAM? Do you have a reference SAM project file for testing? Do you check for negative cumulative excess dollars for each month? I think we need both an ssc test and a SAM project file for validation using rates with dollar rollovers, no excess dollars, and some months with and some without excess dollars to fully validate.

ssc/cmod_utilityrate5.cpp
```diff
@@ -2203,15 +2209,18 @@ class cm_utilityrate5 : public compute_module
 					dollars_applied += monthly_cumulative_excess_dollars[m - 1];
 				}
 
-				if (monthly_ec_charges[m] < 0)
-				{
-					payment[c - 1] -= monthly_cumulative_excess_dollars[m] + monthly_ec_charges[m];
-					dollars_applied += monthly_cumulative_excess_dollars[m] + monthly_ec_charges[m];
-					monthly_cumulative_excess_dollars[m] = -monthly_ec_charges[m];
+                if (monthly_ec_charges[m] < 0)
```
sjanzou, line 2212: We need the ssc test and SAM to show the hourly and subhourly payments matching with the monthly roll up values and ultimately the annual energy savings value.

test/ssc_test/cmod_utilityrate5_test.cpp
```diff
@@ -430,7 +430,7 @@ TEST(cmod_utilityrate5_eqns, Test_Residential_TOU_Rates_net_billing_carryover_ap
 
     ssc_number_t cost_with_system;
     ssc_data_get_number(data, "elec_cost_with_system_year1", &cost_with_system);
-    EXPECT_NEAR(-173.7, cost_with_system, 0.1);
+    EXPECT_NEAR(-206.61, cost_with_system, 0.1);
```
sjanzou, line 434: Can we add a test folder in SAM with a corresponding rate and values or maybe easier is to use a rate in a SAM project file and then build corresponding ssc tests with that with an Excel workbook as final validation... what do you think?

sjanzou (changes requested):
I think we need additional tests in ssc and corrsponding project files and tests in SAM with and Excel workbook to keep these issues from popping back up...

ssc/cmod_utilityrate5.cpp
brtietz: Yes, I'm kicking myself for not doing more validation on this before the release.

Here's the SAM file for the SSC tests that were affected by this PR: 
[net_billing_monthly_variation.zip](https://github.com/NREL/ssc/files/5895343/net_billing_monthly_variation.zip)

This is the file that Paul sent me that caught the net metering with $ credits bug initially - we expect net metering credits to align with any negative bills, which they do after this PR: 
[net-metering-$-credits-test-sam-2020-11-29.zip](https://github.com/NREL/ssc/files/5895358/net-metering-.-credits-test-sam-2020-11-29.zip)

And for net billing: 
[monthly-billing-results-tests-2021-11-29.zip](https://github.com/NREL/ssc/files/5895367/monthly-billing-results-tests-2021-11-29.zip)

I'll post Excel files when I have them. Once I've used Excel to validate the existing tests (and find any other bugs), we can determine whether the existing load profiles hit all of the relevant cases or whether we need additional load profiles to fully validate.

brtietz: This file: 
[net_metering_monthly_variation_match_excel.zip](https://github.com/NREL/ssc/files/5895625/net_metering_monthly_variation_match_excel.zip)

matches https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/SAM%202016.3.14%20Test%20Files%20with%20Spreadsheets/Tiered%20TOU%20single%20meter%20rollover%20%24.xlsx, with that caveat that "Net Metering Credits" now corresponds to "net metering credits applied in month m"

brtietz: Ok, it looks like I introduced the bug here: https://github.com/NREL/ssc/commit/f94e50429a1bcb330a7a33af6679f930c7fbb2cf#diff-b252d497c315d490c9e716fbc1355f21e42f51db78299f43e7935b068bb03716

I wrote an Excel file for the December case of net_billing_monthly_variation.zip here: 
[2020.11.29_net_billing_carryover_dec.xlsx](https://github.com/NREL/ssc/files/5896471/2020.11.29_net_billing_carryover_dec.xlsx)

I'll get the December case fixed up, then see how that affects the other cases - it might be additional columns on the above Excel can represent the true-up month switches.

For what it's worth, that number matches the number in "Electricity bill with system ($/yr)" in 2020.2.29, but not "Electricity bill with system ($)" - so there might be another bug lurking somewhere.

test/ssc_test/cmod_utilityrate5_test.cpp
brtietz: Matching SAM case: https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/net_billing_monthly_variation.sam

Excel workbook: https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/SAM%202020.11.29%20Rollover%20Month%20Tests/2020.11.29_net_billing_carryover.xlsx

brtietz:
So I've solved two bugs: the credits bug introduced in https://github.com/NREL/ssc/commit/f94e50429a1bcb330a7a33af6679f930c7fbb2cf#diff-b252d497c315d490c9e716fbc1355f21e42f51db78299f43e7935b068bb03716, and a mismatch between utility_bill_w_sys_ym and elec_cost_with_system. As of 2020.11.29 (and 2020.2.29) these values do not match for net billing w/ carryover:

image

After the changes in the latest push they match:

image

The values in electricity bill with system now also match the sum of the monthly bills, where they didn't before!!

I'll do another pass Monday adding variables to the tests to lock that down, as well as looking for additional cases that Steve recommended. In particular, I think developing unit tests around net_metering_monthly_variation_match_excel will be valuable.

@cpaulgilman Note that the above means that the True-Up credits appear cleanly in the credit month for net billing, so you should just be able to use `net_billing_credit[m]` now.

test/ssc_test/cmod_utilityrate5_test.cpp
```diff
@@ -455,10 +458,10 @@ TEST(cmod_utilityrate5_eqns, Test_Residential_TOU_Rates_net_billing_carryover_ja
     ssc_data_set_number(data, "ur_nm_credit_month", 0); // January
 
     int analysis_period = 25;
-    ssc_data_set_number(data, "system_use_lifetime_output", 1);
+    ssc_data_set_number(data, "system_use_lifetime_output", 0);
     ssc_data_set_number(data, "analysis_period", analysis_period);
     set_array(data, "load", load_profile_path, 8760);
-    set_array(data, "gen", gen_path, 8760 * analysis_period);
+    set_array(data, "gen", one_year_gen_path, 8760);
```
brtietz, line 464: This needed to change to a non-lifetime simulation so that it was easier to match with Excel

ssc/cmod_utilityrate5.cpp
sjanzou: @brtietz, thanks for adding all of these validation tests and supporting documentation. This has been sorely missing from way back before you got here. If anyone is to be kicking themself, it is me!

brtietz:
@cpaulgilman @sjanzou I just pushed some additional checks in existing tests and an additional case for net billing with carryover. This should be ready for a second review.

sjanzou (approved):

sjanzou (approved):
Thanks for the additional tests - looks good for patch 1

brtietz:
SAM file that I demoed during the meeting: https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/net_billing_monthly_variation.sam

cpaulgilman (approved):

cpaulgilman:
Added SAM and spreadsheet files for testing of monthly bill calculations to private SAM-documentation repo:

https://github.com/NREL/SAM-documentation/tree/master/Unit%20Testing/Utility%20Rates/Monthly%20Bill%20Test%20Files%20with%20Spreadsheets

PR #528, opened 2021-02-03
Li-Ion NMC battery lifetime

### LINKED ISSUE #521: Integrate NREL NMC lifetime model
Add lifetime model for Kokam NMC cells from Kandler Smith 2017 into `battery_t`.

Create new lifetime_t base class, of which the old cycle-calendar-separated lifetime model is derived.

Create new lifetime_nmc_t derived class, which uses combined cycling and calendar effects.

Add this to upcoming patch so that it can be sync'd in E+ for their new release.

@rchintala13, @nmerket

nmerket:
@dguittet Here's the new battery lifetime model @rchintala13 wrote, and all the lifetime refactor work you did. 

Changes:
- Added lithium ion nmc capacity degradation model (Q_li, and Q_neg)
- Added lithium ion nmc states to allow for complex cycling in computing charge degradation. The complex cycling method for computing capacity degradation is based on one of the NREL's teams work and published in the applied energies paper “Analysis of degradation in residential battery energy storage systems for rate-based use-cases.” They use the same Kandler Smith model. 
- Updated thermal_t class to incorporate Kandler Smith's NMC model to update capacity as a function of temperature (analytical model). 
- Ran tests to check for state initialization, and accuracy of functions. 

close #521


shared/lib_battery_lifetime_nmc.h
```diff
@@ -0,0 +1,115 @@
...
+#include "lib_util.h"
+#include "lib_battery_lifetime_calendar_cycle.h"
+
+struct lifetime_nmc_state {
+    double q_relative_li;                // %, SEI degradation
+    double q_relative_neg;               // %, cycle degradation
+    double dq_relative_li_old;
+    double dq_relative_neg_old;
+    double DOD_max;
+    int n_cycles_prev_day;
```
dguittet, line 36: Removed the float version of calendar age to use single double variable

shared/lib_battery_lifetime_nmc.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+void lifetime_nmc_t::runLifetimeModels(size_t lifetimeIndex, bool charge_changed, double prev_DOD, double DOD,
+                                       double T_battery) {
+    double q_last = state->q_relative;
+
+    // update day age of battery
+    size_t day_age_of_battery_old = (size_t)(state->day_age_of_battery);
+    state->day_age_of_battery += params->dt_hr / (double)util::hours_per_day;
+    auto ts_per_day = (size_t)(util::hours_per_day / params->dt_hr);
```
dguittet, line 153: Using the lifetime index to calculate when a full day has passed, this is a bit more rigorous vs converting between a floating point and an int or size_t

dguittet (changes requested):
This looks really great, nice work! 

I made a lot of minor modifications to the code to align better with how SAM works and its coding style and standards, such as renaming variables, functions, cleaning things up, comments etcs.

I think if you're okay with the changes I've made, we're pretty close. The major thing I'd like to see is more tests and with more detail. I've added review comments throughout the tests to specify what I'm thinking, as well as an example in `NoCyclingCapacityTest`. Meanwhile, I'm going to request an additional review from the SAM team, and continue to look through this.

Thanks!

shared/lib_battery_lifetime.h
```diff
@@ -25,234 +24,78 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    double average_range;
+    double day_age_of_battery;
 
-    std::shared_ptr<cycle_state> cycle;
+    // CALCYC model state
     std::shared_ptr<calendar_state> calendar;
+    std::shared_ptr<cycle_state> cycle;
+
+    // NREL NMC model state
+    std::shared_ptr<lifetime_nmc_state> nmc_state;
```
dguittet, line 63: Do you mind renaming this so it doesn't have `state` in the name? I think that'll make it more consistent with the naming in the other state structs.

shared/lib_battery_lifetime_nmc.h
```diff
@@ -0,0 +1,115 @@
...
+
+    double b1_dt;
+    double b2_dt;
+    double b3_dt;
+    double c2_dt;
+
+    friend std::ostream& operator<<(std::ostream& os, const lifetime_nmc_state& p);
+};
+
+class lifetime_nmc_t : public lifetime_t {
```
dguittet, line 49: Could you please add a reference to where this analytical model is from in this class description?

dguittet:
Looks like `cmod_battery_stateful` wasn't updated to use the new logic for the new thermal analytical calculation. I'm happy to add that in later today, unless @rchintala13 would like to?

rchintala13:
Thank you so much for the review @dguittet.   I will go through all the comments and start working on incorporating the recommended changes. I hope to finish them by Wednesday EOD.

shared/lib_battery_lifetime_nmc.cpp
```diff
@@ -0,0 +1,216 @@
...
+    state->nmc_state->b2_dt += b2_dt_el;
+    state->nmc_state->b3_dt += b3_dt_el;
+
+    //computations for q_neg
+    double c2_dt_el = c2_ref * exp(-(Ea_c_2 / Rug) * (1. / T_battery - 1. / T_ref))
+        * pow(0.01 * DOD, beta_c2);
+    state->nmc_state->c2_dt += c2_dt_el;
+
+    //Run capacity degradation model after every 24 hours
+    if (lifetimeIndex % ts_per_day == 23) {
```
brtietz, line 190: Has this been tested subhourly? My hunch is this could run 4 times per day with 15 minute code.

shared/lib_battery_lifetime_nmc.cpp
```diff
@@ -0,0 +1,216 @@
...
+//        printf("%zu, %f, %zu, %f, %f, %f\n", lifetimeIndex, state->day_age_of_battery, (size_t)(day_age_of_battery_old), state->nmc_state->q_relative_li, state->nmc_state->q_relative_neg, state->q_relative);
+    }
+//    else
+//        printf("%zu, %f, %zu, %zu\n", lifetimeIndex, state->day_age_of_battery, (size_t)state->day_age_of_battery, day_age_of_battery_old);
+
+    state->q_relative = fmin(state->q_relative, q_last);
+}
+
+double lifetime_nmc_t::estimateCycleDamage() {
+    return 0;
```
brtietz, line 204: This is going to break a number of the dispatch algorithms. Is there a way to easily estimate this by tracking recent or average damage?

shared/lib_battery_lifetime_nmc.h
```diff
@@ -0,0 +1,115 @@
...
+    /// Capacity degradation due to SEI
+    double runQli();
+
+    /// Capacity degradation due to cycles
+    double runQneg(double T_battery, double SOC);
+
+    void initialize();
+
+    /// NMC Kandler Smith parameters
+    double Ea_d0_1 = 4126.0;
```
brtietz, line 87: Which of these can safely be hardcoded and which need to be tied to SSC inputs? V_ref and Rug are especially suspicious given what a user can change in SSC.

ssc/cmod_battery.cpp
```diff
@@ -812,22 +816,43 @@ battstor::battstor(var_table& vt, bool setup_model, size_t nrec, double dt_hr, c
...
+            lifetime_model = new lifetime_calendar_cycle_t(batt_vars->batt_lifetime_matrix, dt_hr, batt_vars->batt_calendar_lifetime_matrix);
+        }
+        else {
+            lifetime_model = new lifetime_calendar_cycle_t(batt_vars->batt_lifetime_matrix, dt_hr);
+        }
     }
-    else if (batt_vars->batt_calendar_choice == lifetime_params::CALENDAR_CHOICE::TABLE) {
-        lifetime_model = new lifetime_t(batt_vars->batt_lifetime_matrix, dt_hr, batt_vars->batt_calendar_lifetime_matrix);
+    else if (batt_vars->batt_life_model == 1) {
+        lifetime_model = new lifetime_nmc_t(dt_hr);
```
brtietz, line 832: Do we need to throw exceptions based on the chosen voltage model if the user chooses this lifetime model? Should it be possible to run this lifetime model with a lead acid battery? (it looks like it currently is possible, at least from PySAM)

test/shared_test/lib_battery_test.cpp
```diff
@@ -708,3 +734,18 @@ TEST_F(lib_battery_test, ReplaceByCapacityTest){
...
+    auto capacityModelNMC = new capacity_lithium_ion_t(q, SOC_init, SOC_max, SOC_min, dtHour);
+    auto voltageModelNMC = new voltage_dynamic_t(n_series, n_strings, Vnom_default, Vfull, Vexp, Vnom, Qfull, Qexp, Qnom,
+                                         C_rate, resistance, dtHour);
+    auto lossModelNMC = new losses_t(monthlyLosses, monthlyLosses, monthlyLosses);
+
+    auto batteryNMC = std::unique_ptr<battery_t>(new battery_t(dtHour, chemistry, capacityModelNMC, voltageModelNMC, lifetimeModelNMC, thermalModelNMC, lossModelNMC));
+    double I = Qfull * n_strings * 2;
+
+    batteryNMC->run(0, I);
+
```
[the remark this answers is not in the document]
brtietz: Does this test assert anything at the moment or just log?

test/shared_test/lib_battery_lifetime_test.cpp
```diff
@@ -338,15 +337,91 @@ TEST_F(lib_battery_lifetime_test, runCycleLifetimeTestWithRestPeriod) {
...
+    EXPECT_NEAR(s.cycle->q_relative_cycle, 99.96, tol);
+    EXPECT_NEAR(s.cycle->rainflow_Xlt, 90, tol);
+    EXPECT_NEAR(s.cycle->rainflow_Ylt, 90, tol);
+    EXPECT_NEAR(s.cycle->rainflow_jlt, 2, tol);
     EXPECT_NEAR(s.range, 90, tol);
     EXPECT_NEAR(s.average_range, 90, tol);
     EXPECT_NEAR(s.n_cycles, 2, tol);
 }
 
+TEST_F(lib_battery_lifetime_nmc_test, updateCapacityTest) {
```
brtietz, line 351: Thanks for these tests! It looks like these cover two out of 11 cells in the paper. Can you add the following?
- Cycling at different temperatures
- Storage at different temperatures
- Different C rates

brtietz (changes requested):
It's exciting to see that this is so close! I added some comments about the longer term plans for integrating with SSC and SAM, let me know if you have questions on any comments.

shared/lib_battery_lifetime_nmc.h
dguittet: I'm not sure it's possible to make a connection between the voltage assumptions in this life model with the ones in SAM's voltage model. The V_ref parameter as chosen will determine how the other parameters have been fit. I'm not sure adjusting some of these basic assumptions will make sense / extrapolate well. Instead, treating it as a black-box where the inputs are battery operational parameters (SOC, cycle count, T, age) seems a safer bet.

shared/lib_battery_lifetime_nmc.cpp
```diff
@@ -0,0 +1,216 @@
...
+        Uneg = ((0.0859 - 0.2420) / 0.9) * (SOC - 0.1) + 0.2420;
+    return Uneg;
+}
+
+double lifetime_nmc_t::calculate_Voc(double SOC) {
+    double Voc;
+    if (SOC <= 0.1)
+        Voc = ((0.4679) / 0.1) * SOC + 3;
+    else if (SOC <= 0.6)
+        Voc = ((3.747 - 3.4679) / 0.5) * (SOC - 0.1) + 3.4679;
```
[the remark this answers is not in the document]
dguittet: I think we need to separate this model's internal voltage model from SAM's by treating this life model as a black-box.

ssc/cmod_battery.cpp
dguittet: Good point, added an error check

shared/lib_battery_lifetime_nmc.cpp
dguittet: Yes, good catch. I've fixed this to be `lifetimeIndex % ts_per_day == ts_per_day - 1`. Adding in subhourly tests

test/shared_test/lib_battery_lifetime_test.cpp
dguittet: Added cycling and storage tests with different temperatures and C rates to lib_battery_lifetime_test.cpp-- the results are now tracking well with the static model from the paper.

shared/lib_battery_lifetime_nmc.cpp
dguittet: Updated to return the derivative of the cycle loss term assuming the average DOD and reference temperature

dguittet:
In this implementation, we derive the instantaneous derivative of the global model and then every day (day is the time unit for the model), we integrate over that discrete timestep by determining the values for each parameter at that timestep. The parameter's coefficients are found by taking the average across the timesteps that make a day.

A weakness to this implementation of the dynamic / "complex cycling method" model is that for cycles that last longer than a day, the coefficients from the days prior to the completion of the cycle are lost. For a two-day cycle, the loss term is computed on the second day when it completes, and the parameter's coefficients are the average of the second day only, ignoring those of the first day.

A more rigorous implementation would choose the integration steps separately for the calendar and the cycle model. For the calendar model, the integration should be performed every day; for the cycle model, the integration should be performed with the completion of each cycle.

PR #530, opened 2021-02-05
Lcos cost inputs

mjprilliman:
-Added LCOS calculation to financial models
-To be used for models with battery, potentially CSP technologies
-Current changes causing random crashes in UI, haven't yet traced it to particular change


shared/lib_battery.h
```diff
@@ -314,9 +314,6 @@ class battery_t {
     // Returns the % replacement if on a capacity schedule. Returns 0 for "none" or "calendar"
     double getReplacementPercent();
 
-    // Change the timestep of the battery and its component models
-    void ChangeTimestep(double dt_hr);
-
```
dguittet, line 319: So this might be a merge conflict, as this shouldn't be getting deleted. I'm noticing some other changes spread throughout such as `dt_hr` being changed back to `dt_hour` in battery state structs. And in `cmod_stateful` as well.

dguittet:
@mjprilliman Exciting! Do you mind adding some references for LCOS and its calculation here? Also, if there's a place to make sense to add that reference as a comment in the code, that would be really helpful for future maintenance as well.

shared/lib_battery.h
mjprilliman: Not sure how I managed to do that but thanks for the catch. Fixed those potential merge conflicts and pushed again.

mjprilliman:
> @mjprilliman Exciting! Do you mind adding some references for LCOS and its calculation here? Also, if there's a place to make sense to add that reference as a comment in the code, that would be really helpful for future maintenance as well.

https://www.cell.com/joule/pdfExtended/S2542-4351(18)30583-X

That is the main reference we have been going off of. I will be sure to include references in the code as we go.

cpaulgilman (approved):
Should the variable nameplate1 be named something like nameplate_battery or battery_nameplate for clarity?

ssc/cmod_host_developer.cpp
```diff
@@ -2715,6 +2755,127 @@ class cm_host_developer : public compute_module
...
+    //Equation based on "Projecting the Future Levelized Cost of Electricity Storage Technologies", https://www.cell.com/joule/pdfExtended/S2542-4351(18)30583-X
+    if (is_assigned("battery_total_cost_lcos") && as_double("battery_total_cost_lcos") != 0) {
+        for (int y = 0; y <= nyears; y++) {
+            cf_lcos.at(0, y) = cf.at(CF_battery_replacement_cost, y);
+            cf_lcos.at(1, y) = cf.at(CF_battery_replacement_cost_schedule, y);
+        }
+        int grid_charging_cost_version = 0;
+        lcos_calc(this, cf_lcos, nyears, nom_discount_rate, inflation_rate, lcoe_real, cost_prefinancing, disc_real, grid_charging_cost_version, 0);
+        /*
+        double lcos_investment_cost = as_double("battery_total_cost_lcos"); //does not include replacement costs
```
brtietz, line 2769: Can this commented out code be removed now that lcos calc is in a function?

ssc/cmod_utilityrate5.cpp
```diff
@@ -1775,24 +1828,194 @@ class cm_utilityrate5 : public compute_module
...
+                            else if (ur_ec_hourly_acc_period == 2)
+                                cumulative_energy = daily_surplus_energy;
+
+
+                            // cumulative energy used to determine tier for credit of entire surplus amount
+                            ssc_number_t credit_amt = 0;
+
+                            ssc_number_t tier_credit = 0.0, sr = 0.0, tier_energy = 0.0;
+                            // time step sell rates
+                            if (as_boolean("ur_en_ts_sell_rate")) {
```
brtietz, line 1914: This should always be false. ur_calc only runs for metering options 0 and 1, and the GUI disables time series rates for those options.

ssc/cmod_utilityrate5.cpp
```diff
@@ -1775,24 +1828,194 @@ class cm_utilityrate5 : public compute_module
...
+                                if (c < rate.m_ec_ts_sell_rate.size()) {
+                                    tier_energy = energy_surplus;
+                                    sr = rate.m_ec_ts_sell_rate[c];
+                                    tier_credit = tier_energy * sr * rate_esc;
+                                    //curr_month.ec_energy_surplus.at(row, surplus_tier) += (ssc_number_t)tier_energy;
+                                }
+                            }
+
+                            // Fall back to TOU rates if m_ec_ts_sell_rate.size() is too small
+                            if (tier_credit == 0) {
```
brtietz, line 1924: Given that the above is expected to always be false - this will always be true.

ssc/cmod_utilityrate5.cpp
```diff
@@ -1775,24 +1828,194 @@ class cm_utilityrate5 : public compute_module
...
+                                else {
+                                    tier_energy = energy_deficit;
+                                    br = curr_month.ec_tou_br.at(row, deficit_tier);
+                                    //br_bin_it = br;
+                                   
+                                    
+                                    
+                                }
+                            }
+                            if (as_boolean("ur_en_ts_buy_rate")) {
```
brtietz, line 2011: Same as sell rates: I'd expect this is always false.

ssc/cmod_utilityrate5.cpp
```diff
@@ -1775,24 +1828,194 @@ class cm_utilityrate5 : public compute_module
...
+                            // base period charge on units specified
+                            double cumulative_deficit = -e_in[c];
+                            if (ur_ec_hourly_acc_period == 1)
+                                cumulative_deficit = monthly_deficit_energy;
+                            else if (ur_ec_hourly_acc_period == 2)
+                                cumulative_deficit = daily_deficit_energy;
+
+                            ssc_number_t tier_charge = 0.0, br = 0.0, tier_energy = 0.0;
+                            // time step sell rates
+                            if (as_boolean("ur_en_ts_buy_rate")) {
```
brtietz, line 1970: Same as sell rates: I'd expect this is always false.

ssc/cmod_utilityrate5.cpp
```diff
@@ -586,6 +596,22 @@ class cm_utilityrate5 : public compute_module
...
+            monthly_excess_dollars_earned_batt(12),
+            monthly_excess_kwhs_earned_batt(12),
+            monthly_net_billing_credits_batt(12),
+            monthly_ec_rates_batt(12),
+            monthly_salespurchases_batt(12),
+            monthly_load_batt(12), monthly_system_generation_batt(12), monthly_elec_to_grid_batt(12),
+            monthly_elec_needed_from_grid_batt(12),
+            monthly_cumulative_excess_energy_batt(12), monthly_cumulative_excess_dollars_batt(12), monthly_bill_batt(12), monthly_peak_batt(12), monthly_test_batt(12),
+            monthly_two_meter_sales_batt(12),
+            monthly_true_up_credits_batt(12); // Realistically only one true-up month will be non-zero, but track them all for monthly outputs
```
brtietz, line 613: I'm confused by these new inputs. It looks like they aren't actually used?

Additionally a lot of the computations inside ur_calc don't seem to make it outside of that function. 

It looks like the only thing that is happening here is the computation of e_batt_cy and p_batt_cy, is the rest of the code needed?

mjprilliman: Originally I thought there would need to be changes to utility rate code to calculate the grid charging costs. After conversations I ended up not needing much or any of these code changes. I tried to remove changes but I see I missed a lot. I'm looking through it but I'm fairly confident this can be reverted to develop

PR #534, opened 2021-02-11
Moved dynamic voltage model error checking

### LINKED ISSUE #517: Voltage model lacking error checking message
**Describe the bug**
Running the battery model with impossible voltage model inputs results in an unhandled unknown exception (non-helpful error message). Error checking is needed to explain that to run the voltage model the voltage inputs must satisfy the following: Vfull > Vexp > Vnom

1. Open a PV-Battery, Residential case
2. Go to Battery Cell and System -> Battery Voltage
3. Change Exponential zone cell voltage to a value > Fully charged cell voltage
4. Simulate

image

image

mjprilliman:
-Better fix to #517 
-Do we need a catch in battery_stateful as well? If so where?
-Closes #517


dguittet (approved):
Thanks!

PR #536, opened 2021-02-12
Sam 275 fix nested kwh per kw tiers

brtietz:
Fix SAM bug https://github.com/NREL/SAM/issues/275. Track the tiers required by the peak demand using some new vectors, then overwrite all of the bounds in kWh limits for the subsequent code. Unit test confirms that the behavior matches expected in Excel: https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/block_step/GPC_PLL_Tiered_Bill_Calc_Example_v3_btm_tests.xlsx

Note that we only get the expected utility bill without system in the net metering cases. Anything that uses ur_calc_timestep will have a subtle difference in tier allocation. I've decided to make fixing that a separate branch: https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/block_step/GPC_PLL_Tiered_Bill_Calc_Example_v3_btm_tests.xlsx


shared/lib_utility_rate_equations.cpp
```diff
@@ -173,65 +173,102 @@ void rate_data::init(int num_rec_yearly) {
...
 void rate_data::init_energy_rates(bool gen_only) {
 	// calculate the monthly net energy per tier and period based on units
 	for (int m = 0; m < (int)m_month.size(); m++)
 	{
 		// check for kWh/kW
-        size_t start_tier = 0;
-        size_t end_tier = (int)m_month[m].ec_tou_ub.ncols() - 1;
         size_t num_periods = (int)m_month[m].ec_tou_ub.nrows();
-        size_t num_tiers = end_tier - start_tier + 1;
+        size_t num_tiers = (int)m_month[m].ec_tou_ub.ncols(); // Likely to be overwritten later
```
sjanzou, line 186: Is this finalized? Have the TOU upper bounds stored in columns or rows now?

shared/lib_utility_rate_equations.cpp
```diff
@@ -173,65 +173,102 @@ void rate_data::init(int num_rec_yearly) {
...
-			// 3. resize use and chart based on number of tiers in kWh/kW section
+			// 2. fill in kWh tiers as needed, up to kWh/kW * flat peak
+            // 3. repeat until maximum tier is reached
 			// 4. assumption is that all periods in same month have same tier breakdown
 			// 5. assumption is that tier numbering is correct for the kWh/kW breakdown
 			// That is, first tier must be kWh/kW
+            // See example at: https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/block_step/GPC_PLL_Tiered_Bill_Calc_Example_v3_btm_tests.xlsx
 			if ((m_month[m].ec_tou_units.ncols() > 0 && m_month[m].ec_tou_units.nrows() > 0)
-				&& ((m_month[m].ec_tou_units.at(0, 0) == 1) || (m_month[m].ec_tou_units.at(0, 0) == 3)))
+				&& check_for_kwh_per_kw_rate(m_month[m].ec_tou_units.at(0, 0)))
```
sjanzou, line 200: Thanks for breaking this out to work with the new URDB structure that can have two rows of kWh/kW blocks adjacent to one another now.

shared/lib_utility_rate_equations.cpp
```diff
@@ -242,11 +279,11 @@ void rate_data::init_energy_rates(bool gen_only) {
 				{
 					for (size_t tier = 0; tier < num_tiers; tier++)
 					{
-						br.at(period, tier) = m_month[m].ec_tou_br_init.at(period, start_tier + tier);
-						sr.at(period, tier) = m_month[m].ec_tou_sr_init.at(period, start_tier + tier);
-						ub.at(period, tier) = m_month[m].ec_tou_ub_init.at(period, start_tier + tier);
+						br.at(period, tier) = m_month[m].ec_tou_br_init.at(period, tier_numbers[tier]);
```
sjanzou, line 282: Great - another level of abstraction to handle the irregular numbering in the URDB!

test/shared_test/lib_utility_rate_equations_test.cpp
```diff
@@ -164,3 +164,216 @@ TEST(lib_utility_rate_equations_test, test_seasonal_demand_charges)
     // Same profile provides demand charges in June
     ASSERT_NEAR(10.0, data.get_demand_charge(5, 0), 0.1);
 }
+
+// Excel version of this test: https://github.com/NREL/SAM-documentation/blob/master/Unit%20Testing/Utility%20Rates/block_step/GPC_PLL_Tiered_Bill_Calc_Example_v3_btm_tests.xlsx
```
sjanzou, line 168: Thanks for this! Another DocHub entry!

sjanzou (approved):
Do you feel confident with the number of tests and block tiers in the URDB?
Thanks for tackling this!

shared/lib_utility_rate_equations.cpp
brtietz: Yes, the convention in rate_data::setup_energy_rates is that periods are rows and tiers are columns. I'll add some comments in the header file to reflect this.

PR #537, opened 2021-02-17
Enable leap day to be included in PV simulations using weather data

### LINKED ISSUE #494: Leap year in non-annual PV simulations
**Describe the bug**
Given the new functionality of non-annual simulations, the PV models use time stamp information directly. In particular, the hour of the year is calculated for a given timestamp. However, this is currently throwing an error when the data includes Feb 29 in the function `hour_of_day`:

https://github.com/NREL/ssc/blob/596403ea2a2aca2913bf491f1537ad92c90ee905/shared/lib_util.cpp#L911

**Expected behavior**
Would like to be able to calculate a correct hour of year regardless of when the leap day appears

@MRossol

janinefreeman:
This pull request enables February 29 to be included in a detailed PV simulation when inputting weather via the weather data variable. It also enables annual outputs for an 8760 simulation that includes Feb 29 but not Dec 31, specifically for our reV/REeDS folks (looking at you @grantbuster @mrossol ). Fixes #494 . It does NOT enable handling 8784 data, which is still on our wishlist.

After merging into patch, can we ask either @grantbuster or @mrossol to test a new ssc library to make sure this works for you? If so, let us know which OS and we can get that built.


MRossol:
@janinefreeman  for the win! I think the easiest way to test this is either:
1) To have @dguittet  clone reV and run the tests from a python env with the new ssc
2) Have @dguittet  help us set-up a custom pysam env on eagle that we can run the tests from.

grantbuster:
@MRossol I think I can also help test from the reV side of things. @dguittet has given me an ssc executable before that I was able to patch into the reV env for a test.

ssc/common.cpp
```diff
@@ -1222,28 +1222,39 @@ double shading_factor_calculator::dc_shade_factor()
...
 {
+    //first, set up some timestep variables
 	int ts_per_hour = 0; //determine the number of timesteps in each hour
 	if (leapyear)
 		ts_per_hour = (int)(m_nRecords / 8784);
 	else
 		ts_per_hour = (int)(m_nRecords / 8760);
 	double ts_min = 60. / ts_per_hour; //determine the number of minutes of each timestep
+
+    //next, check if the data has leap day (feb 29). need to do this because some toold pass in 8760 data that contains feb 29 and not dec 31
```
dguittet, line 1233: typo in "tools"

dguittet (approved):
Looks good and thanks for the tests! What currently happens if there is a non-annual simulation with 8784 data?

dguittet:
@MRossol @grantbuster I'll upload a new pysam distribution to Box, and you can update the python env you're running reV in (whether locally or on Eagle) by doing `pip install <path to pysam.whl>`. What platform should I build it for?

MRossol:
@dguittet see the failed test here: https://github.com/NREL/reV/runs/1487453080?check_suite_focus=true.
Let's go with Linux, that way @grantbuster can test locally or I can test it on Eagle.

dguittet:
Here's the link to the [PySAM v 2.2.1 whl](https://app.box.com/s/34q6b38c7i6c11v9c6mytep602nd1h68) for linux Python 3.7. Install via `pip install <whl file>`. Please note it's not fully vetted yet and is intended just to check the leap year tests.

dguittet:
@MRossol I uploaded a build for the wrong branch! Sorry, just updating now. Do you have a ETA for when the testing will be done-- would it be possible to do before EOD tomorrow, when we plan to release the patch?

ssc/common.cpp
janinefreeman: Thanks!

janinefreeman:
@dguittet Non-annual data with length of 8784 will trigger the function to check if it's a year, but it won't pass, so it will be flagged as non-annual data. Feb 29 is allowed through in non-annual data now, though.

grantbuster:
Hey @dguittet it looks like our tests pass! Thanks for implementing that. 

Can you clarify what the improvements were on the ssc side of things? It is now more sensitive to the input dates but we still cant run non-8760s? Can we run partial timeseries, like a week of data?

janinefreeman:
@grantbuster Thanks for testing! We had made some changes in version 2020.2.29 to allow non-8760 data when entered using the weather data structure- so yes, you can run a week at a time (or any other length of time). The additions I had made to allow that made the code rely more on the timestamps in the weather data, and I neglected to have it allow Feb 29. However, when you run non-8760 data, you can't get "annual" results, because they would be meaningless. So the code checked that 8760 data (or multiples thereof) was in fact a complete, evenly spaced, Jan 1-Dec 31 year before it would report annual values. For the purposes of reV, this pull request also modifies that function so that it will now allow Jan 1-Dec 30 if Feb 29 is present, and still report annual values.

grantbuster:
Awesome thanks @janinefreeman !

Issue #535, opened 2021-02-12
Utility Rate tiers split by hour, not by kWh


brtietz:
When testing the block step rates, my test was slightly off (~$14 in the month of July). The root cause of the issue is that the utility rate code in ur_calc_timestep (net billing and buy all / sell all) is only allowing each hour's energy to be given to one tier. See below screenshots:

![image](https://user-images.githubusercontent.com/5530592/107828479-32104f00-6d46-11eb-85cc-2bdc741cbbdd.png)

and the resulting charges:

![image](https://user-images.githubusercontent.com/5530592/107828507-3a688a00-6d46-11eb-8e23-793a7046fe6e.png)

These are different from net metering, which accurately allocates all kWh in one tier before moving on to the next tier.

See attached SAM file:

[evenly_distributed_tiers_test.zip](https://github.com/NREL/ssc/files/5974545/evenly_distributed_tiers_test.zip)

Update ur_calc_timestep to split the kWh appropriately.

brtietz:
Closed in https://github.com/NREL/ssc/pull/538

PR #538, opened 2021-02-18
Ssc 535 fix tiers net billing

brtietz:
Tracks the surplus and deficit tiers (they can be different) differently for net billing and buy all / sell all. This code will correctly allocate kWh to lower tiers before moving on to upper tiers.

Utility rate:

image

Behavior in 2020.11.29: 

image

Behavior in this PR:

image

Test file available in https://github.com/NREL/ssc/issues/535

### DIFF
--- a/ssc/cmod_utilityrate5.cpp
+++ b/ssc/cmod_utilityrate5.cpp
@@ -1948,8 +1948,11 @@ public:
 		// single meter so single net accumulation
 		ssc_number_t daily_surplus_energy;
 		ssc_number_t monthly_surplus_energy;
+        ssc_number_t step_surplus_energy; // For tier rollovers
 		ssc_number_t daily_deficit_energy;
 		ssc_number_t monthly_deficit_energy;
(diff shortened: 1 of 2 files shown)


sjanzou (commented):
Hmmm... your breakdown and image provided for the tier breakdown are great! However, to get the full tier value at each bound, does it make sense that the breakdown is continuous rather than discrete at the timestep intervals as previously reported? That is, the first image in the 2020.11.29 release values...

brtietz:
You can see the changes from the first tier in January to the second in the attached image
Screenshot 2021-02-18 090641
The tier transition occurs occurs at 11 am on the 5th. 2020.11.29 allocates the entire hour to the higher tier, whereas this code splits it between tiers 1 and 2, resulting in a lower energy charge in that hour.

I've never seen text from a utility that supports the discrete allocations, they always cut it off strictly at the tier kWh.

sjanzou (approved):
Okay, so the weather file is the measurement of time for the production but the analog meter is the unit of time for the utility bill - we should list that in help...

PR #549, opened 2021-03-11
Batt fixes

dguittet:
For the NMC model, relying on differences between lifetime indices to compute whether a full day has elapsed or not had difficulties when using different time steps. That issue occurred when towards the end of the day, the hourly battery would "finish" its day, then get switched over to a smaller timestep (1 min) and would run into needing to "finish" its day again-- resulting in problems. The previous fix was to make the index when the day "finishes" equal to `lifetimeIndex % ts_per_day == ts_per_day - (size_t)(1. / params->dt_hr)` so that the repetition wouldn't happen.

The fix here explicitly calculates elapsed day as `fabs(state->nmc_li_neg->cum_dt - 1.) < 1e-7`.

Also, reduce the run time of `calculate_max_discharge_w` fxns.


shared/lib_battery_lifetime_nmc.cpp
```diff
@@ -194,12 +196,13 @@ void lifetime_nmc_t::runLifetimeModels(size_t lifetimeIndex, bool charge_changed
     state->nmc_li_neg->c2_dt += c2_dt_el;
 
     // Run capacity degradation model after every 24 hours
-    if (lifetimeIndex % ts_per_day == ts_per_day - (size_t)(1. / params->dt_hr)) {
+    if (fabs(state->nmc_li_neg->cum_dt - 1.) < 1e-7) {
```
brtietz, line 199: The PR title mentions worrying about switching to a smaller timestep, but might this cause trouble if the timestep got bigger? For example, what if for some reason the code or data switched to 30 minute timesteps at 11:45 pm? Or are there constraints elsewhere that enforce the larger stepsize hitting the top of the hour?

I'm specifically wondering why this is `fabs(state->nmc_li_neg->cum_dt - 1.) < 1e-7` rather than `state->nmc_li_neg->cum_dt >= 1`

brtietz (changes requested):
Looks good overall, but I have one question about edge cases in lib_battery_lifetime_nmc

shared/lib_battery_lifetime_nmc.cpp
dguittet: That's a good point. The point with using the precision of `1e-7` was because sometimes cum_dt was 0.99999999, and that's when the day should finish. The switching case you brought up would actually be a problem for the life model, so I may need to add some extra functionality and testing to account for that, thanks!

dguittet: Add an error check to ChangeTimestep to [revent changing from smaller to larger timestep in the middle of the larger time step: 
"battery_t dt_hr step size can only be changed to a higher step size when the current time step is at a time step common to both the previous and new step size. For instance, if running 30-min steps, step size can only be increased to 60-min step at the hour."

dguittet: For the `lifetime_nmc_t` class itself, I added code to break up a step size that would take cum_dt > 1 into two. This way the first part takes cum_dt to 1 and gets processed, then the remainder is added to the next step.

PR #550, opened 2021-03-12
Better enforce "don't export to grid" constraints

brtietz:
Two changes: First - pull adjustments out of else-if chain to ensure they are applied consistently. This also helps increase charging targets in price signals dispatch. Second, adjust the custom dispatch targets if they are too high, reducing iteration. Adjust tests as needed

- Includes a new test to trigger this specific bug
- Changes to two other custom dispatch tests were expected
- Price signals dispatch now has better access to charging power, due to a related constraint in the if-else chain
- The BattWatts changes were unexpected, but justified when I checked the tests in the debugger

### DIFF
--- a/shared/lib_battery_dispatch.cpp
+++ b/shared/lib_battery_dispatch.cpp
@@ -659,8 +659,9 @@ bool dispatch_automatic_t::check_constraints(double& I, size_t count)
 				I -= dI;
 			}
 		}
+
 		// Behind the meter
-		else if (m_batteryPower->meterPosition == dispatch_t::BEHIND)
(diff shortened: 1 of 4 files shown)


brtietz:
Companion SAM PR: https://github.com/NREL/SAM/pull/561

dguittet (approved):

dguittet:
@brtietz There was a failing test-- did you check that?

brtietz:
No, I didn't spot it until after I merged it into patch. Fix went directly to patch: https://github.com/NREL/ssc/commit/fdb3272f10a1b33e4b78159b8c45e86db36e2e49

Price signals has some minor differences running between Windows and Linux, I haven't had time to debug it yet.

PR #556, opened 2021-03-19
Made battery computed variables inouts for viewing in results

### LINKED ISSUE #555: Make battery computed variables output to results
**Describe the bug**
Make battery size computed variables available as output metrics

**To Reproduce**
1. Run battery case
2. Cannot see battery computed properties from results

**Expected behavior**
Change Input variables to InOut variables to allow for battery computed variables to display in outputs

mjprilliman:
-Battery computed variables now available for selection as outputs
-Closes #555


cpaulgilman (changes requested):
This is a good improvement. I think the variable labels should be changed to start with "Battery" to avoid confusion and to make them easier to find in alphabetized lists of variables. For example rename "Number of cells in series" to "Battery number of cells in series" to avoid potential confusion with fuel cells.

Thanks,
Paul.

cpaulgilman (approved):

PR #560, opened 2021-04-01
ensure num_tiers is enforced for the first tier of each month

brtietz:
Solves https://github.com/NREL/SAM/issues/563 - the variable to set the number of tiers in a month was being reset for each period, rather than enforcing the number of tiers in the first period of the month. This produces an existing error message in the SAM file attached to the bug, and responds correctly to different months having different numbers of tiers.


cpaulgilman (approved):
Looks good. I tried to make the error messages when there is a discrepancy between weekday/weekend schedules and demand or energy rate tables a little easier to understand.

PR #563, opened 2021-04-13
Add minute to PV irradiance processor error messages

cpaulgilman:
Adding minute to PV and PVWatts model irradiance error messages that show timestep m d y h.

Fixes SAM Issue 584: https://github.com/NREL/SAM/issues/584


sjanzou (approved):
Looks good - definitely helpful for sub-hourly simulations! Please merge into patch and develop when you get a chance. Thanks!

PR #566, opened 2021-04-22
Wave model time series

mjprilliman:
-Allow for time series wave data to be used in wave compute module
-Adds options for jpd and time series wavefile processing
-Wave_resource_data class for data array resource specification for use outside of SAM GUI


ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+            for (size_t r = 0; r < 21; r++)
+            {
+                getline(ifs, buf);
+                values.clear();
+                values = split(buf);
+                if (values.size() != 22)
+                {
+                    for (size_t c = 0; c < 2; c++)
+                    {
+
```
dguittet, line 311: What's happening here with the empty for loop?

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+        }
+        else if (as_integer("wave_resource_model_choice") == 0) {
+            ssc_number_t* mat = allocate("wave_resource_matrix", 21, 22);
+            for (size_t r = 0; r < 21; r++)
+            {
+                getline(ifs, buf);
+                values.clear();
+                values = split(buf);
+                if (values.size() != 22)
```
dguittet, line 307: Throw an error, and give a more descriptive error message for why there should be 22 columns. What is the label and units on this column?

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+        //ssc_number_t* wave_periods = allocate("wave_energy_period", numberRecords);
+        //if (values.size() != 22)
+        if (as_integer("wave_resource_model_choice") == 1)
+        {
+            size_t numberRecords = 0;
+            size_t year_index = 0, month_index = 0, day_index = 0, hour_index = 0, minute_index = 0, period_index = 0, height_index = 0;
+            getline(ifs, buf); //Skip past column labels for record counting
+            while (getline(ifs, buf))
+                numberRecords++;
+            assign("number_records", (int)numberRecords);
```
dguittet, line 183: is there an expected number of records? is subhourly ok? is a 8784 ok? Maybe do an error check here

ssc/cmod_wavefile.cpp
```diff
@@ -27,8 +27,11 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+    { SSC_INPUT,         SSC_NUMBER,      "wave_resource_model_choice",           "Hourly or JPD wave resource data",                                 "0/1",             "",             "Weather Reader",          "",                         "INTEGER",                  "" },
+    { SSC_INPUT,         SSC_STRING,      "wave_resource_filename",               "local weather file path",                     "",       "",                      "Weather Reader",      "?",                       "LOCAL_FILE",      "" },
+    { SSC_INPUT,         SSC_STRING,      "wave_resource_filename_ts",               "local weather file path",                     "",       "",                      "Weather Reader",      "?",                       "LOCAL_FILE",      "" },
+
+    { SSC_INPUT,         SSC_NUMBER,      "use_specific_wf_wave",               "user specified file",                     "0/1",       "",                      "Weather Reader",      "?=0",                       "INTEGER,MIN=0,MAX=1",      "" },
```
dguittet, line 34: What does this mean? How is this specified file related to wave_resource_filename_ts or wave_resource_filename?

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+
+            // header name value pairs
+            std::vector<std::string> keys = split(buf);
+            values = split(buf1);
+            int ncols = (int)keys.size();
+            int ncols1 = (int)values.size();
+
+            if (ncols != ncols1 || ncols < 13)
+            {
+                throw exec_error("wave_file_reader", "incorrect number of header columns: " + std::to_string(ncols));
```
dguittet, line 130: Explain a bit more how many are expected and why

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+            assign("data_source", var_data(values[11]));
+            assign("notes", var_data(values[12]));
+        }
+        // read in 21 rows x 22 columns
+
+        //ssc_number_t* mat = allocate("wave_resource_matrix", 21, 22);
+        //size_t numberRecords = 2920;
+        //ssc_number_t* wave_heights = allocate("wave_significant_height", numberRecords);
+        //ssc_number_t* wave_periods = allocate("wave_energy_period", numberRecords);
+        //if (values.size() != 22)
```
dguittet, line 175: remove if unneeded

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+                    //value_1 = split(buf);
+                    hour1 = (ssc_number_t)std::stod(values[hour_index]);
+                    hourdiff = hour1 - hour0;
+                }
+                month[r] = (ssc_number_t)std::stod(values[month_index]);
+                timecheck[r] = (ssc_number_t)std::stod(values[hour_index]);
+                if (r > 0) {
+                    if (timecheck[r] - timecheck[r - 1] != hourdiff && timecheck[r] != 0) {
+                        throw exec_error("wave_file_reader", "Time steps are nonuniform");
+                        timecheck[r] = 999;
```
dguittet, line 250: any code after a throw won't do anything

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+
+        //ssc_number_t* mat = allocate("wave_resource_matrix", 21, 22);
+        //size_t numberRecords = 2920;
+        //ssc_number_t* wave_heights = allocate("wave_significant_height", numberRecords);
+        //ssc_number_t* wave_periods = allocate("wave_energy_period", numberRecords);
+        //if (values.size() != 22)
+        if (as_integer("wave_resource_model_choice") == 1)
+        {
+            size_t numberRecords = 0;
+            size_t year_index = 0, month_index = 0, day_index = 0, hour_index = 0, minute_index = 0, period_index = 0, height_index = 0;
```
dguittet, line 179: as an error check, I would make these `int`s initialized to `-1` and then after line 212, check that all these indices have been properly identified (are no longer -1 / invalid)

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+                wave_heights[r] = (ssc_number_t)std::stod(values[height_index]);
+                wave_periods[r] = (ssc_number_t)std::stod(values[period_index]);
+
+                //Make JPD from time series data
+
+                ts_significant_wave_height = wave_heights[r];
+                ts_energy_period = wave_periods[r];
+                for (size_t j = 0; j < 21; j++) {
+
+                    if (r == 0) mat[(j + 1) * ncols] = (0.25 + j * 0.5);
```
dguittet, line 266: move this and line 273 to right under the instantiation of `mat` in line 224, since this appears to be initializing the bin widths of matrix. That would also be a good place to add in a comment that the rows are for height and columns are for period

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+                    if (abs(ts_significant_wave_height - (0.25 + (j - 1) * 0.5)) <= 0.25 && j != 0) {
+                        sig_wave_height_index = j;
+                        if (r != 0) break;
```
dguittet, line 269: won't need the if r!=0 statement when the mat initialization is done beforehand

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+                        energy_period_index = m;
+                        if (r != 0) break;
+                    }
+
+
+                }
+
+
+                //mat[sig_wave_height_index * ncols + energy_period_index] = mat[sig_wave_height_index * ncols + energy_period_index] + 1 / 2920 * 100;
+                mat[sig_wave_height_index * ncols + energy_period_index] += 0.0342465753;
```
dguittet, line 284: does the increment value to the PDF depend on the number of records or time steps or anything?

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+                //Set decimal values to 2 for JPD
+                if (r == numberRecords - 1) {
```
dguittet, line 286: Do this after the resource reading loop, like in line 295

ssc/cmod_wavefile.cpp
```diff
@@ -60,100 +78,255 @@ class cm_wave_file_reader : public compute_module
...
+                    //throw exec_error("wave_file_reader", "incorrect number of data columns: " + std::to_string(values.size()));
+                }
+                for (size_t c = 0; c < 22; c++)
+                {
+                    if (r == 0 && c == 0)
+                        mat[r * 22 + c] = 0.0;
+                    else
+                        mat[r * 22 + c] = std::stod(values[c]);
+                }
+            }
```
dguittet, line 322: probably want to check validity of the JDF to make sure the it all sums to ~1

dguittet (changes requested):
First round of review... mostly focused on usability and coding. Happy to re-review when you're ready!

ssc/cmod_wavefile.cpp
```diff
@@ -45,12 +48,27 @@ static var_info _cm_wave_file_reader[] = {
...
+	{ SSC_OUTPUT,        SSC_MATRIX,      "wave_resource_matrix",              "Frequency distribution of resource",                                  "m/s",   "",                       "Weather Reader",      "?",                        "",                            "" },
+   // { SSC_OUTPUT,        SSC_ARRAY,       "time_check",                        "Time check",                                                          "",      "",                       "Weather Reader",      "?",                        "",                            "" },
+   // { SSC_OUTPUT,        SSC_ARRAY,       "month",                        "Month",                                                          "",      "",                       "Weather Reader",      "?",                        "",                            "" },
+
+    { SSC_OUTPUT,        SSC_ARRAY,       "wave_significant_height",           "Wave height time series data",                                        "m",     "",                       "Weather Reader",      "?",                        "",                            "" },
```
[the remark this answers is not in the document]
mjprilliman: Was trying to match my usage across the wave modules but matching the attributes from the dataset makes more sense. Correcting

ssc/cmod_pvwattsv7.cpp
```diff
@@ -1229,6 +1230,21 @@ class cm_pvwattsv7 : public compute_module
...
+                                p_annual_energy_dist_time[m] = m;
+                            }
+                            if (wf.month == m && wf.hour == (h - 1)) {
+                                p_annual_energy_dist_time[h * 13 + m] += p_gen[idx] / step_per_hour;
+                                break;
+                            }
+                        }
+                    }
+                }*/
+
```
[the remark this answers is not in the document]
mjprilliman: Remnants of development of heatmap generation in UI. Removed as this is now stored in a common.cpp function

ssc/cmod_wavefile.cpp
```diff
@@ -45,12 +48,27 @@ static var_info _cm_wave_file_reader[] = {
...
+    { SSC_OUTPUT,        SSC_ARRAY,       "month",                   "Month",                            "mn",     "1-12",                  "Weather Reader",      "wave_resource_model_choice=1",                       "",                          "" },
+    { SSC_OUTPUT,        SSC_ARRAY,       "day",                     "Day",                              "dy",     "1-365",                 "Weather Reader",      "wave_resource_model_choice=1",                       "",                          "" },
+    { SSC_OUTPUT,        SSC_ARRAY,       "hour",                    "Hour",                             "hr",     "0-23",                  "Weather Reader",      "wave_resource_model_choice=1",                       "",                          "" },
+    { SSC_OUTPUT,        SSC_ARRAY,       "minute",                  "Minute",                           "min",    "0-59",                  "Weather Reader",      "wave_resource_model_choice=1",                       "",                          "" },
+    */
```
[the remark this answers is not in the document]
mjprilliman: Undecided on whether to include this or not as the current usage of the wavefile module results in these time series variables being in the results metrics. Had commented these out as I have not yet figured out how to run wavefile only in the UI callbacks rather than from setconfig() in startup.lk.

ssc/cmod_wavefile.cpp
mjprilliman: Currently this is done in the compute module rather than the file reader. Do you think it should be moved?

ssc/cmod_wavefile.cpp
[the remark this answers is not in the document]
dguittet: Matching across both the wave modules and the dataset would be ideal

ssc/cmod_wavefile.cpp
dguittet: It should be moved or duplicated such that any path using file reading performs the check.

dguittet (changes requested):
I think it looks better! Tests are passing now too. 

Could you please clean up the empty spaces & unneeded code, etc?

Also, there are several warnings for both cmod_mhk_wave and cmod_wavefile:
image
image


I created a test PySAM with these updates for Windows Python version 3.7 (let me know if you need another version) here for you to test before I release the new PySAM: https://nrel-my.sharepoint.com/:u:/g/personal/dguittet_nrel_gov/ESXd7aT7LJhOkf7u4qACtLMBHCqdZ6rTXNBZGngVnRVqRQ?e=Wwd63m

dguittet (approved):
Examples in PySAM workbook worked! Nice job on finishing this up!

cpaulgilman (changes requested):
This looks good. A few comments:

For the wave_resource_model_choice exclusive var, the labels "Frequency of Occurrence" and "Time Series" would be easier to understand for the blue dropdown at the top of the Wave Resource page.

We use the term "SAM CSV" for the solar resource file format to differentiate from other formats like EPW, TMY3, etc. I think I would avoid using it in the file name for the wave resource data since there aren't other formats to deal with.

Please remove the apostrophe from the The Jannette's Pier weather file for both the frequency and time series file to avoid possible problems on some operating systems.

For the solar resource files, I've avoided using spaces in the file names. A recent convention is to use hyphens instead of underscores in file names because underscores are hidden in underlined links, so you might consider using that convention here.

I think we can avoid having multiple filenames for these wave files in the UI form. The Solar Resource Data form has different file names to try to maintain backwards compatibility with older versions. I think you can remove the equation functions involving use_specific-wf_wave etc. Hopefully we can get away with one or two resource file names instead of several:

image


The variable name file_name might not be the best, maybe wave_data_file_freq and wave_data_file_ts?

And one issue:

**1. Error message about Solar Resource window**

1. Click **Add/remove weather file folders**.
2. Click **Cancel**.

image

cpaulgilman (approved):

Issue #565, opened 2021-04-21
voltage_table_t::calculate_max_discharge_w breaks dispatch constraints


brtietz:
Lead acid batteries using the voltage table can export to grid, since the dispatch current can be greater than the load. The root cause of the issue is that voltage_table_t::calculate_max_discharge_w specifies unreasonably high output currents, and this is the last constraint that is enforced.

Unclear whether this is a units issue with qmax, or something else. A brute force solution would be to enforce dod is >0 and <100 in the function.

**To Reproduce**
1. Create a PV-Battery residential case with a lead acid battery that uses the voltage table
2. Run the model with any dispatch
3. Note that (a) the SOC goes well below the SOC limits specified on the dispatch page, and (b) the electricity from the battery is greater than the load, meaning this constraint is violating other constraints.

**Expected behavior**
voltage_table_t::calculate_max_discharge_w should respect charge limits specified in the dispatch functions. 

**Screenshots**
Peak shaving:

![image](https://user-images.githubusercontent.com/5530592/115620963-19c33000-a2b3-11eb-9e59-72b774fc971b.png)

Specified limits:

![image](https://user-images.githubusercontent.com/5530592/115620996-23e52e80-a2b3-11eb-80c2-b46a60e9988a.png)

**Desktop (please complete the following information):**
 - Version 2020.11.29 r1 - problem exists on patch, and the code is the same on develop.

brtietz:
It looks like the issue is that the voltages in the default lead acid voltage table are bank voltages, not cell voltages. Three recommendations for resolution:

(1) Update the Lead Acid default to a lower table voltage (should be ~2 V @ 50%)
(2) Do add the "brute force" dod constraints in the discharge function above.
(3) Consider a warning for the voltage table values: for example if all of the values are strictly above or strictly below the cell voltage, prompt the user to adjust the cell voltage.

dguittet:
Can you link to where `voltage_table_t::calculate_max_discharge_w` is leading to the over-dispatch? This function is meant to calculate the maximum power available to be discharged. Regardless of whether there is an issue in bank vs cell voltages which should definitely be fixed, this function is meant to be used by the dispatcher to determine max limits, and not vice versa ( no dependency of voltage model on dispatch models ). Perhaps a new function is required such as "calculate_current_for_SOC" or something.

[...]

PR #568, opened 2021-04-27
Ssc 565 fix voltage discharge

brtietz:
Fix the ssc side of #565 - SSC will now throw an exception if the nominal voltage is outside of the values provided in the voltage table. Additionally add the depth of discharge guardrails to voltage_table_t::calculate_max_discharge_w - this reduces the over-dispatch issue significantly (see screenshot below) but does not solve the SOC issue reported in 565. Another issue will be created to handle that in a future patch.

New behavior:

image

Improvement in SOC and dispatch relative to the filed issue. Still need to fully restrict dispatch from exceeding SOC limit with a new function.

### DIFF
--- a/shared/lib_battery_voltage.cpp
+++ b/shared/lib_battery_voltage.cpp
@@ -90,9 +90,20 @@ void voltage_table_t::initialize() {
     // save slope and intercept for every set of points to interpolate between
     std::sort(params->voltage_table.begin(), params->voltage_table.end(),
               [](std::vector<double> a, std::vector<double> b) { return a[1] > b[1]; });
+    bool need_less_than_nom = true;
+    bool need_greater_than_nom = true;
+
     for (size_t i = 0; i != params->voltage_table.size(); i++) {
         double DOD = params->voltage_table[i][0];
         double V = params->voltage_table[i][1];
+
+        if (need_less_than_nom && V < params->Vnom_default) {
+            need_less_than_nom = false;
+        }
+        else if (need_greater_than_nom && V > params->Vnom_default) {
+            need_greater_than_nom = false;
+        }
+
         double slope = 0;
         double intercept = V;
         if (i > 0) {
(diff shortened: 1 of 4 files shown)


brtietz:
Interesting, those pass on Windows. I have a hunch that std::sort is giving us different results on Linux and Windows. Thoughts on transitioning to std::stable_sort https://stackoverflow.com/questions/23985891/what-is-the-difference-between-stdsort-and-stdstable-sort/23986246 ? That was a recommended fix here: https://jvet.hhi.fraunhofer.de/trac/vvc/ticket/1467 and may solve some of the other issues I've been seeing (such as https://github.com/NREL/ssc/pull/571/commits/10ff44bfb41348ed29d2eee06f767653dc42336f)

brtietz:
Good news - this turned out to be a "I was having trouble building" problem instead of a "Windows vs Linux" problem. 

Bad news - This has turned into a cascade of failing tests. I fixed the exceptions by lowering vnom, and updated the values in voltage_t where it made sense. The core of the issue is that  battery_t::calculate_current_for_power_kw does not enforce SOC limits the same way as calculate_max_discharge_kw and calculate_max_charge_kw. This is shown in DischargeBatteryModelSubHourlyWSOCLimits, and the new values in voltage_table_lib_battery_voltage_test::calculateMaxDischargeHourly_table. In the latter test, the final current value should be ~9, and the final SOC ~5 - but the way the SOC limits are applied in calculate_max_discharge_kw skew how that function reads the voltage table. 

If I remove the SOC limits from that function, it affects things like lib_battery_test::AdaptiveTimestepNMC, which is a pretty significant scope creep. Do you think it's possible that the SOC limits are causing problems in all of the models and that things need to be updated that significantly?

cpaulgilman (approved):
This looks good for the example given in #565. 

I made a couple of  small revisions to the error messages. I also added an explanation in Help for 2020.11.29 Patch 2.

Is the cell nominal voltage required to be within the range shown in the UI voltage curve graph for the electrochemical model?

dguittet:
@cpaulgilman There isn't a check in the UI voltage curve graph for the voltage table's nominal voltage being included. Currently the user will be notified only after trying to simulate.

PR #570, opened 2021-04-29
Remove fuelcell and battery outputs from cases without those systems

mjprilliman:
-Added assignment checks to only output fuel cell and battery OandM and replacement cost outputs when there is a fuelcell or battery, respectively
-Removed battery efficiency, AC loss, and DC loss for cases without a battery
-Closes https://github.com/NREL/SAM/issues/510
-Closes https://github.com/NREL/SAM/issues/448

### DIFF
--- a/ssc/cmod_merchantplant.cpp
+++ b/ssc/cmod_merchantplant.cpp
@@ -2996,15 +2996,16 @@ public:
 		save_cf( CF_property_tax_assessed_value, nyears, "cf_property_tax_assessed_value" );
 		save_cf( CF_property_tax_expense, nyears, "cf_property_tax_expense" );
 		save_cf( CF_insurance_expense, nyears, "cf_insurance_expense" );
-        if ((as_integer("en_batt") == 1) && (as_integer("batt_replacement_option") > 0)) {
+        if (as_integer("en_batt") == 1) {
             save_cf(CF_battery_replacement_cost, nyears, "cf_battery_replacement_cost");
             save_cf(CF_battery_replacement_cost_schedule, nyears, "cf_battery_replacement_cost_schedule");
         }
-        if (is_assigned("fuelcell_replacement_option") && (as_integer("fuelcell_replacement_option") > 0)) {
+        if (is_assigned("fuelcell_replacement_option")) {
             save_cf(CF_fuelcell_replacement_cost, nyears, "cf_fuelcell_replacement_cost");
             save_cf(CF_fuelcell_replacement_cost_schedule, nyears, "cf_fuelcell_replacement_cost_schedule");
         }
-		save_cf( CF_ebitda, nyears, "cf_ebitda" );
+        save_cf(CF_operating_expenses, nyears, "cf_operating_expenses");
+        save_cf( CF_ebitda, nyears, "cf_ebitda" );
 		save_cf( CF_net_salvage_value, nyears, "cf_net_salvage_value" );
 		save_cf( CF_total_revenue, nyears, "cf_total_revenue" );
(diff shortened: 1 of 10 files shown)


ssc/cmod_merchantplant.cpp
```diff
@@ -2992,11 +2996,14 @@ class cm_merchantplant : public compute_module
 		save_cf( CF_property_tax_assessed_value, nyears, "cf_property_tax_assessed_value" );
 		save_cf( CF_property_tax_expense, nyears, "cf_property_tax_expense" );
 		save_cf( CF_insurance_expense, nyears, "cf_insurance_expense" );
-		save_cf(CF_battery_replacement_cost, nyears, "cf_battery_replacement_cost");
-		save_cf(CF_battery_replacement_cost_schedule, nyears, "cf_battery_replacement_cost_schedule");
-		save_cf(CF_fuelcell_replacement_cost, nyears, "cf_fuelcell_replacement_cost");
-		save_cf(CF_fuelcell_replacement_cost_schedule, nyears, "cf_fuelcell_replacement_cost_schedule");
-		save_cf( CF_operating_expenses, nyears, "cf_operating_expenses" );
+        if ((as_integer("en_batt") == 1) && (as_integer("batt_replacement_option") > 0)) {
```
brtietz, line 2999: Mind removing the second half of this if statement ("batt_replacement_option")? Some of the work I'm doing right now compares cases using "no replacement" to various replacement strategies, so having zeroes come out of SSC is useful to that workflow.

ssc/cmod_singleowner.cpp
```diff
@@ -3260,10 +3264,14 @@ class cm_singleowner : public compute_module
...
 		save_cf( CF_insurance_expense, nyears, "cf_insurance_expense" );
-		save_cf(CF_battery_replacement_cost, nyears, "cf_battery_replacement_cost");
-		save_cf(CF_battery_replacement_cost_schedule, nyears, "cf_battery_replacement_cost_schedule");
-		save_cf(CF_fuelcell_replacement_cost, nyears, "cf_fuelcell_replacement_cost");
-		save_cf(CF_fuelcell_replacement_cost_schedule, nyears, "cf_fuelcell_replacement_cost_schedule");
+        if ((as_integer("en_batt") == 1) && (as_integer("batt_replacement_option") > 0)) {
+            save_cf(CF_battery_replacement_cost, nyears, "cf_battery_replacement_cost");
+            save_cf(CF_battery_replacement_cost_schedule, nyears, "cf_battery_replacement_cost_schedule");
+        }
+        if (is_assigned("fuelcell_replacement_option") && (as_integer("fuelcell_replacement_option") > 0)) {
```
brtietz, line 3271: Might want to apply the same logic to fuel cell replacement option as batteries above.

brtietz (changes requested):
Thanks for cleaning all of this up! I had one request on the replacement options, since I consider it easier to handle zeroes than nans when comparing the "no replacements" cases to other cases. Other than that looks good!

cpaulgilman (changes requested):
Changes look good to me, and agree with Brian's suggestions.

Since this touches cmod_singleowner,cpp, could you fix the group name for issuance_of_equity in the inputs var table? It looks like someone accidentally pasted into it. It should be "Cash Flow Total and Returns" instead of "TCash Flow Total and Returnsotal Cash Flow and Return":

https://github.com/NREL/ssc/blob/158dcee9bc291dfb043c3fbfa6e90d5960703870/ssc/cmod_singleowner.cpp#L628

brtietz (approved):

PR #571, opened 2021-04-30
tweaks to price signals dispatch for the pvsc paper. 

brtietz:
Decrease the value of stored energy so things perform better in situations where manual dispatch performs well. Now out performing manual by $500 (out of $3.6M), a 1.4% improvement over the previous dispatch.


shared/lib_battery_dispatch_automatic_btm.cpp
```diff
@@ -640,23 +640,7 @@ void dispatch_automatic_behind_the_meter_t::plan_dispatch_for_cost(dispatch_plan
         }
     }
 
-    for (i = 0; i < _steps_per_hour && (i < sorted_grid.size()); i++)
```
brtietz, line 643: Since steps per hour was usually 1, this wasn't actually doing anything. The decrease in stored energy value associated with choosing a lower marginal cost period resulted in a substantially lower NPV.

dguittet (approved):

PR #572, opened 2021-04-30
Change PV open circuit current labels to fix SAM 562

cpaulgilman:
See https://github.com/NREL/SAM/issues/562


janinefreeman:
I think this may be a misleading change- the Isc across a module is the same as the Isc of a string because modules are connected in series. In SAM, this output is assigned on line 1972ish of cmod_pvsamv1.cpp:

`PVSystem->p_currentShortCircuit[nn][idx] = (ssc_number_t)Subarrays[nn]->Module->currentShortCircuit;`

This is likely why it was not originally labeled to belong to either a module or a string, because it's the same for both. Thoughts on how to make that not confusing?

cpaulgilman:
I agree that the current from a string is the same as the current from any given single module in that string.

I think the confusion is that the current label "Subarray 1 Short circuit DC current" makes it seem like this is the current from the entire subarray rather than the current through either a module or string in the subarray. See this post on the SAM forum: https://sam.nrel.gov/forum/forum-general/3365 and issue on the SAM repo: https://github.com/NREL/sam/issues/562.

The label "Subarray 1 String short circuit DC current" is helpful because it suggests that to calculate the subarray current you should multiply the string current by the number of strings. The alternative "Subarray 1 Module short circuit DC current" would also be accurate, but might tempt someone to multiply the module Isc by the number of modules.

janinefreeman:
Ahhh now I understand- I had been looking at issue 562 in SSC and was very confused. String current makes sense to me, feel free to keep it and close!

PR #575, opened 2021-05-10
Replaces pull request ssc 574

sjanzou:


Matthew-Boyd (approved):
Pulled and tested--runs good for the trough models.

PR #577, opened 2021-05-12
Removed default assignment on fuelcell replacement cost option 

mjprilliman:
-Removed ?=0 constraint on fuelcell_replacement_option to not have fuel cell replacement costs show up on every configuration
-See #570


brtietz (approved):
Changes look good! Thanks for the quick turn around!

PR #578, opened 2021-05-14
Fix warnings

dguittet:
@Matthew-Boyd I didn't notice the warnings being emitted out of the csp equation files during the review, so I went ahead and fixed them, along with other more recently-introduced ones. We're nowhere near 0 warnings for ssc but I'd be remiss to add more.


Matthew-Boyd (approved):
Thank you @dguittet. Looks good.

PR #579, opened 2021-05-25
NREL NMC Lifetime Model Validation

dguittet:
Update SAM implementation of the NREL NMC Battery Life Model and perform validation based on `V5_GrNMC_Kokam_75Ah` model implementation for Kokam 75Ah Gr/NMC cell performance in K. Smith, et al, Life Prediction Model for Grid-Connected Li-ion Battery Energy Storage System, 2017

Changes:
 - Update cycle counter to track the most recent cycle's range and max DOD
 - Track the range and max DOD of all the cycles that occur in each day
 - Update predictive life model results and tests
     - Interpolate Uneg and Voc from table
     - If no cycles have yet elapsed, predict range of cycles that will occur and predict average SOC of cycles that will occur in that day; otherwise use DOD range and SOC average of cycles from current day
 - EstimateCycleDamage using cumulative loss and total number of cycles for average
 - Rename some cycle and NMC variables for clarity
 - BatteryStateful INOUT variables
 
Validation:
 - TestAgainstKokamData in lib_battery_lifetime_nmc_test tests against reference data and model
 
image

 - Pretty good agreement to original model, differences due primarily to how SAM handles time and SOC profiles
 - The difference for cell 6 & 7 grows after Qneg overtakes QLi and is primarily due to completed/integer cycles per day  underestimating vs using incomplete/floating pt cycle counting. For instance, a sequence of number of cycles per day of (8, 9, 8, 9, ...) vs (8.5, 8.5, 8.5, 8.5, ...). The difference would probably be less with less aggressive cycling

FYI @nmerket @rchintala13, the TLDR here is this update results in better agreement with original model and data


nmerket:
Wow. This looks like a significant improvement. We will be definitely interested incorporating it in EnergyPlus once it is part of a release.

shared/lib_battery_lifetime_nmc.cpp
```diff
@@ -82,33 +86,23 @@ lifetime_t * lifetime_nmc_t::clone() {
...
-
 double lifetime_nmc_t::calculate_Uneg(double SOC) {
-    double Uneg;
-    if (SOC <= 0.1)
-        Uneg = ((0.2420 - 1.2868) / 0.1) * SOC + 1.2868;
-    else
-        Uneg = ((0.0859 - 0.2420) / 0.9) * (SOC - 0.1) + 0.2420;
+    size_t prev_ind = floor(SOC * 10);
+    size_t next_ind = prev_ind + 1;
+    double Uneg = unegs[prev_ind] + (unegs[next_ind] - unegs[prev_ind]) / 0.1 * (SOC - (double)prev_ind / 10);
```
brtietz, line 92: This crashes at SOC of 100% - prev_ind ends up being 10, which pushes next_ind to 11, which is out of bounds. It looks like calculate_Voc will have the same problem.

dguittet: Good catch, thanks

brtietz: When I run the tests locally, the fix has the same problem. It looks like line 93 should be `>=`

dguittet: Yes, thanks! Updated.

brtietz (changes requested):
Code looks good, and I love the idea of having JSON files with the NMC data in the tests. That said, I didn't see any calls to lifetime_nmc_t::replaceBattery in the tests. Is that worth adding to this PR?

brtietz (approved):

Issue #581, opened 2021-05-27
Snow modeling assumptions for true-tracking vs back-tracking in PVWatts7 appear to differ unexpectedly


cpr-chas:
The snow model library makes an assumption on the number of strings per slant height based on the tracking type:
cmod_pvwatts7.cpp

            if (pv.type == ONE_AXIS)
                pv.nmody = 1; // e.g. Nextracker or ArrayTechnologies single portrait
            else
                pv.nmody = 2; // typical fixed 2 up portrait

Based on this logic, a backtracking system (type == ONE_AXIS_BACKTRACKING) would get the 2-up version.  I'm guessing this was unintentional and that all single-axis tracker types should get nmody of 1.

I'd be happy to contribute a fix and test coverage after the desired behavior is confirmed.
I haven't done any specific testing yet - just found this while learning about snow modeling.

janinefreeman:
@cpr-chas You're absolutely right, the intended behavior is what you describe and not what's happening in the code. If you have a chance to contribute a fix and test, that would be fantastic! Thanks! :)

cpr-chas:
Sure.  The change is obviously simple, but it might take me a day or two to figure out the test side.

janinefreeman:
Let us know if you need any support on that side, we're happy to help!

janinefreeman:
Hi @cpr-chas ! Just checking in on this as we start to corral issues for a fall release. Thanks!

PR #582, opened 2021-05-28
Fix backtracking configuration for snow shedding model in PVWatts7

cpr-chas:
The snow shedding model requires a number of strings or panels along the height of the row slant height.  The PVWatts7 model estimates this based on the tracking type, making the assumption that most fixed systems have strings of panels in each row and single-axis have only one.  However, for single-axis trackers that support backtracking it was falling through to the fixed assumption while it should be consistent with the single-axis true-tracker assumption.

Added some test cases to verify behavior of the snow model in all three cases.

This address issue #581.

### DIFF
--- a/ssc/cmod_pvwattsv7.cpp
+++ b/ssc/cmod_pvwattsv7.cpp
@@ -555,7 +555,7 @@ public:
                 log(util::format("system size is too small to accurately estimate regular row-row self shading impacts. (estimates: #modules=%d, #rows=%d).  disabling self-shading calculations.",
                 (int)pv.nmodules, (int)pv.nrows), SSC_WARNING);*/
 
-            if (pv.type == ONE_AXIS)
+            if (pv.type == ONE_AXIS || pv.type == ONE_AXIS_BACKTRACKING)
                 pv.nmody = 1; // e.g. Nextracker or ArrayTechnologies single portrait
             else
                 pv.nmody = 2; // typical fixed 2 up portrait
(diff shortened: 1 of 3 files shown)


cpr-chas:
I have a concern about the tests in that they validate against specific monthly totals.  It's possible that other modeling changes could affect these totals in a way that breaks the tests - ie, they don't really isolate the snow model.  I considered writing a test that ran the model both with and without snow modeling, but that makes the tests take twice as long, is a little odd because I have to manually call Teardown and Setup in the middle, and I discovered that enabling snow modeling sometimes increases power output slightly (probably some odd rounding error in the end calculations), so I couldn't use EXPECT_LE, at least not without scaling the non-snow case up by a fraction of a percent.  Any thoughts on how these tests could be improved?

cpr-chas:
I don't know what branch I had originally been based on but it must have been pretty old.  I see that the more recent tests do follow a model of calling SSC twice so I'll move towards doing that and verifying the snow loss as a rough percentage of overall power per month and post an update when that's done..

janinefreeman:
@cpr-chas Sorry for the slow response! One alternate approach, since in this case we just want to check that the snow model IS applied for backtracking, would be to run the annual system for the backtracking config with a snowy weather file and just check that the snow loss output `dcsnowloss` is not zero.

Thank you so much for this contribution!!

janinefreeman:
This pull request was used to create pull request #643 , with some updates from NREL due to my delays in merging. Thank you so much @cpr-chas !!

PR #583, opened 2021-06-14
Make heatmap of system power generated dynamic to time interval of weather file

mjprilliman:
-Fix calculation of day of year that timestep fits into based on number of timesteps within each hour
-Remove unused code


cpaulgilman (approved):
Tested with PVWatts and 5-minute data. Looks good to me.

PR #585, opened 2021-06-29
NMC life model replacement & small dt tests

dguittet:
Two additional tests for NMC life model: replacement (fixing code also) and small dt


dguittet:
looks like the changes in this branch were already merged in previous pull request(s)

PR #587, opened 2021-07-06
TOD multipliers for LCOS calculations

mjprilliman:
-Added array size checks for sub-hourly simulations
-Added check for schedule or array ppa multiplier model to build array
-Using var table inputs now rather than extra input in lcos_calc function; Removed extra variable


sjanzou (approved):
Thanks for fixing this Matt!

PR #588, opened 2021-07-07
Battery Life Model for LMO/LTO chemistry

dguittet:
Beta version of a battery life model for LMO/LTO chemistry. The model is based on test data being collected as part of BTMS, and is in-progress as the data collection is still ongoing. The final version will be similar in structure to the version in this PR, and the results will probably not change a lot. It will be published in a paper, so this model needs to be updated in the future with the reference also. 

I'd like to include this version in SAM right now, in develop, in order for it to be used in PySAM for BTMS, as we're planning to start using this beta version for some analyses.

Changes:
- lib_battery_lifetime_lmolto.cpp / .h files : contains the new data structure and class
- `LMOLTO` option to cmod_battery, cmod_pvsamv1, and cmod_batterystateful
- Add state variables of NMC and LMOLTO battery life models as SSC_INOUTs to battery stateful
- Rename enum for life model from `NRELNMC` to `NMC` for simplicity
- Refactor daily cycle tracking state variables & functions from lifetime_nmc_t to lifetime_cycle_t for access from multiple classes
- Tests and validation against test data: `test/input_cases/battery_lmolto_life/lifetime_validation_cell_1.json` contains a dictionary for running the life model and expected outputs for 16 different test conditions

image


shared/lib_battery_lifetime_calendar_cycle.cpp
```diff
@@ -210,6 +212,66 @@ double lifetime_cycle_t::average_range() { return state->average_range; }
 
 double lifetime_cycle_t::capacity_percent() { return state->cycle->q_relative_cycle; }
 
+void lifetime_cycle_t::resetDailyCycles() {
```
brtietz, line 215: I like that this code has moved into the cycle model, good thinking!

shared/lib_battery_lifetime_calendar_cycle.h
```diff
@@ -70,6 +70,13 @@ struct cycle_state {
     int rainflow_jlt;                       // last index in Peaks, i.e, if Peaks = [0,1], then jlt = 1
     std::vector<double> rainflow_peaks;
 
+    // values for current day, refreshed end of each day; used only in analytic life models not calendar_cycle model
```
brtietz, line 73: This header file could also use some additional comments on constructor usage. For example, is it correct that initializing with only params is only used for first time setup, where as params and state are for resuming from a stateful run?

shared/lib_battery_lifetime_lmolto.h
```diff
@@ -0,0 +1,101 @@
...
+extern const double T_ref;
+extern const double F;
+
+struct lifetime_lmolto_state {
+    double dq_relative_cal;                 // %
+    double dq_relative_cyc;                 // %
+    double EFC;                             // Equivalent Full Cycles
+
+    // lifetime capacity updated after 24 hours elapsed using below values, which are all reset each day
+    double EFC_dt;                          // 1
```
brtietz, line 45: What is the // 1 comment? Shouldn't dt have units of hrs or days?

shared/lib_battery_lifetime_lmolto.h
```diff
@@ -0,0 +1,101 @@
...
+    double temp_avg;                        // 1
```
brtietz, line 46: Ditto on units here.

shared/lib_battery_lifetime_lmolto.h
```diff
@@ -0,0 +1,101 @@
...
+    double estimateCycleDamage() override;
+
+    void replaceBattery(double percent_to_replace) override;
+
+protected:
+
+    std::unique_ptr<lifetime_cycle_t> cycle_model;
+
+    /// Capacity degradation due to calendar loss
+    double q1_b0 = 3.4984e-5;
```
brtietz, line 77: Please add units here.

test/shared_test/lib_battery_lifetime_lmolto_test.cpp
```diff
@@ -0,0 +1,416 @@
...
+    auto state = model->get_state();
+
+    EXPECT_EQ(state.n_cycles, 43);
+    EXPECT_NEAR(state.lmo_lto->dq_relative_cal, 0.404, 1e-3);
+    EXPECT_NEAR(state.lmo_lto->dq_relative_cyc, 0.0637, 1e-3);
+    EXPECT_NEAR(state.q_relative, 99.532, 1e-3);
+    EXPECT_NEAR(state.day_age_of_battery, 88, 1e-3);
+}
+
+/** Test focusing on how different time steps affect the integration of a day's degradation in the NMC life model.
```
brtietz, line 185: Comment says "NMC life model"

brtietz (changes requested):
I love the new code structure. Questions & revisions mostly deal with areas that need improved comments.

shared/lib_battery_lifetime_lmolto.cpp
```diff
@@ -0,0 +1,183 @@
...
+}
+
+lifetime_t * lifetime_lmolto_t::clone() {
+    return new lifetime_lmolto_t(*this);
+}
+
+double lifetime_lmolto_t::runQcal() {
+    double SOC_avg = cycle_model->predictAvgSOC(state->cycle->DOD_max - state->cycle->DOD_min);
+    double q1 = q1_b0 * exp(q1_b1 * (1 / pow(state->lmo_lto->temp_avg, 3.)) * sqrt(SOC_avg)) *
+                      exp(q1_b2 * (1 / pow(state->lmo_lto->temp_avg, 2.)) * sqrt(SOC_avg));
```
mjprilliman, line 79: Are these variables going to be in the reference? Comments might be helpful.

shared/lib_battery_lifetime_lmolto.cpp
```diff
@@ -0,0 +1,183 @@
...
+    double energy_throughput = fabs(DOD - prev_DOD) * 0.01 * 0.5;
+    state->lmo_lto->EFC_dt += energy_throughput;
+    state->lmo_lto->EFC += energy_throughput;
+    state->lmo_lto->temp_avg += T_battery * dt_day;
+    state->cycle->cum_dt += dt_day;
+}
+
+void lifetime_lmolto_t::runLifetimeModels(size_t lifetimeIndex, bool charge_changed, double prev_DOD, double DOD,
+                                          double T_battery) {
+    T_battery += 273.15;
```
mjprilliman, line 139: Unless I am mistaken T_battery is converted from Celsius to Kelvin for all the lifetime models. Are there other reasons that this input is being passed in Celsius (`thermal->T_battery()`) ?

mjprilliman (commented):
I don't have much to add. Looks good!

shared/lib_battery_lifetime_lmolto.cpp
dguittet: You're correct about that, but I think the reason Celsius is used and passed around rather than Kelvin is because of the thermal model, and the "cap_vs_temp" input, which is the relationship between the instantaneous thermal capacity and the temp in Celsius.

test/shared_test/lib_battery_lifetime_lmolto_test.cpp
dguittet: thanks, replaced all mention of "nmc" in the file

shared/lib_battery_lifetime_lmolto.cpp
dguittet: Yes, they will be in the reference. They are parameters to the equations, describing the effect of each term in the equation, so not too much to explain beyond that, but I added some comments about the equations

shared/lib_battery_lifetime_lmolto.h
dguittet: Fixed

shared/lib_battery_lifetime_calendar_cycle.cpp
dguittet: Thanks!

shared/lib_battery_lifetime.cpp
```diff
@@ -78,6 +80,16 @@ lifetime_state::lifetime_state(const std::shared_ptr<lifetime_nmc_state>& nmc) {
     q_relative = fmin(nmc->q_relative_li, nmc->q_relative_neg);
 }
 
+lifetime_state::lifetime_state(const std::shared_ptr<lifetime_lmolto_state> &lmolto) {
```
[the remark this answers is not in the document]
dguittet: The idea behind this set of constructors where the inputs are a single component state class (for the calcyc, nmc, or the lmo/lto) doesn't make sense, thanks for bringing my attention to it. I've deleted those 3 constructors since they weren't even being used!

brtietz (approved):
Thanks for updating the comments and constructors!

shared/lib_battery_lifetime.h
```diff
@@ -56,13 +58,18 @@ struct lifetime_state {
...
+
+    // Cycling model state
     std::shared_ptr<cycle_state> cycle;
 
-    // NREL NMC model state
+    // NMC model state
     std::shared_ptr<lifetime_nmc_state> nmc_li_neg;
 
+    // LMO/LTO model state
+    std::shared_ptr<lifetime_lmolto_state> lmo_lto;
```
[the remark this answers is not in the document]
dguittet: That's a good point. In the current code, all these shared_ptrs are not NULL even when they aren't being used, such as the lmolto pointer when using an nmc model. I added comments but also the ability to create only the sub-state structs that will be used. 

https://github.com/NREL/ssc/blob/cc40cae97080011cddf9594c9182f198194c49b6/shared/lib_battery_lifetime.cpp#L50

That'll make it cleaner, and hopefully also clear up a little bit how the cycle model pointer here relates to the one used within the NMC & LMO/LTO models. Specifically, they are the same pointer. 

https://github.com/NREL/ssc/blob/cc40cae97080011cddf9594c9182f198194c49b6/shared/lib_battery_lifetime_nmc.cpp#L32

The state created here has a cycle state that then gets passed to the constructor of the cycle model. 

I'm not sure what you meant by breaking up the cycle counting functions?

PR #589, opened 2021-07-07
fix check_continuous_single_year

dguittet:
Function to check whether solar resource data is continuous was incorrectly stepping through minutes and leading to false negative for a valid 5 min file.

see https://github.com/NREL/pysam/issues/88


sjanzou (approved):
Looks good!

dguittet:
Thanks for the quick turnaround!

PR #595, opened 2021-07-22
PV Smoothing 

sjanzou:
lib_dispatch_pv_smoothing_fom shared library and test
No weather file forecast until merged into develop


shared/lib_battery_dispatch_pvsmoothing_fom.cpp
```diff
@@ -0,0 +1,440 @@
...
+void dispatch_pvsmoothing_front_of_meter_t::update_dispatch(size_t year, size_t hour_of_year, size_t , size_t lifetimeIndex)
+{
+	// Initialize
+	m_batteryPower->powerBatteryDC = 0;
+	m_batteryPower->powerBatteryAC = 0;
+	m_batteryPower->powerBatteryTarget = 0;
+
+    size_t ndx_sampled = lifetimeIndex / m_batt_dispatch_pvs_timestep_multiplier;
+
+    bool linear_interpolation = false; // updated per 7/14/2021 meeting
```
brtietz, line 152: Under what conditions might this be true in the future? If it's unlikely to be used, I'd prefer removing the unused code to leaving it, as I think this implementation (hardcoded to false) is likely to cause confusion.

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
```diff
@@ -0,0 +1,440 @@
...
+        if (AC_upper_bound_on) {
+            if (out_power > AC_upper_bound)
+                out_power = AC_upper_bound;
+        }
+        if (AC_lower_bound_on) {
+            if (out_power < AC_lower_bound)
+                out_power = AC_lower_bound;
+        }
+
+        //                  #calculate desired(unconstrained) battery power
```
brtietz, line 369: Can these be reformatted as C++ comments? The whitespace is hard to read.

test/ssc_test/cmod_pvsamv1_test.cpp
```diff
@@ -743,9 +747,14 @@ TEST_F(CMPvsamv1PowerIntegration_cmod_pvsamv1, NonAnnual)
 {
     //set up a weather data array and unassign the solar resource file
 
-    auto weather_data = create_weatherdata_array(24);
+//    auto weather_data = create_weatherdata_array24(); // runs without error
+    auto weather_data = create_weatherdata_array(24); // throws SEH exception in release mode only from cmod_pvsamv1 on Windows
+   // var_data weather_data;
+   // create_weatherdata_array(weather_data, 24);
```
brtietz, line 753: Please remove or explain commented out code.

test/ssc_test/cmod_pvwattsv7_test.cpp
```diff
@@ -7,6 +7,74 @@
...
 
+TEST_F(CMPvwattsV7Integration_cmod_pvwattsv7, NonAnnual)
+{
+    //set up a weather data array and unassign the solar resource file
+
+    var_data* weather_data = create_weatherdata_array(24); // runs without error
+    ssc_data_unassign(data, "solar_resource_file");
+    ssc_data_set_table(data, "solar_resource_data", &weather_data->table);
+    //   delete weather_data;
+    /*    var_table* vt = static_cast<var_table*>(data);
```
brtietz, line 18: Please remove or explain commented out code.

test/ssc_test/cmod_windpower_test.cpp
```diff
@@ -30,7 +30,8 @@ TEST_F(CMWindPowerIntegration, HubHeightInterpolation_cmod_windpower) {
     ssc_data_get_number(data, "annual_energy", &annual_energy);
     EXPECT_GT(annual_energy, 4e06) << "Annual energy should be higher than height at 90";
 
-    free_winddata_array(windresourcedata);
+//    delete[] windresourcedata;
```
brtietz, line 33: Why is this free no longer required?

brtietz (changes requested):
SSC code looks good overall, mostly some cleanliness requests regarding the transition from development and debugging to production. Nice work!

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
sjanzou: Left in for next meeting review - I prefer to have it removed but they seem to go back and fortth with the step curves in the time series.

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
sjanzou: Yes, they were originally taken directly from the Python code from EPRI

test/ssc_test/cmod_pvsamv1_test.cpp
sjanzou: These were added while trying to remove the occasional SEH test errors on Windows

test/ssc_test/cmod_windpower_test.cpp
sjanzou: The free was moved to the test tear down.

brtietz:
Travis seems to be having trouble with the tests, and I'm also having trouble locally:

From PV_Phoeniz_all_on:
```
    !! This test has probably CRASHED !!
    Test output:
    
    C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\vector(1566) : Assertion failed: vector subscript out of range
```
I dug into this on Generic_w_PV_input_all_on and it looks like the error is in the LCOS calculations in common_financial:

image

charged_pv only has length 1. batt_annual_charge_from_system appears to be empty.

I'm not sure whether this means the test needs to change to ensure this value is reported or the code in the financial models needs to be more robust. This metric appears to work in the GUI.

sjanzou:
@mjprilliman, why is the generic system failing here? The tests run fine on my Windows box. Thanks!

mjprilliman:
> Travis seems to be having trouble with the tests, and I'm also having trouble locally:
> 
> From PV_Phoeniz_all_on:
> 
> ```
>     !! This test has probably CRASHED !!
>     Test output:
>     
>     C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\include\vector(1566) : Assertion failed: vector subscript out of range
> ```
> 
> I dug into this on Generic_w_PV_input_all_on and it looks like the error is in the LCOS calculations in common_financial:
> 
> image
> 
> charged_pv only has length 1. batt_annual_charge_from_system appears to be empty.
> 
> I'm not sure whether this means the test needs to change to ensure this value is reported or the code in the financial models needs to be more robust. This metric appears to work in the GUI.

Must be something to do with analysis_period being set to 1. When that happens the array doesn't seem to have a year 0 like I was assuming so there's the size 1 array @brtietz. My code assumes we skip the first index in a lot of places because that is year 0. Is that the correct array sizing behavior @sjanzou?

mjprilliman:
@brtietz @sjanzou 

image

This is in cmod_battery.cpp. Why make it 1 rather than 2?

sjanzou:
Agreed that should be 2. Same issue in cmod_fuelcell that prevented single year analysis periods to be run.

sjanzou:
PVSmoothing tests working fine on Windows and Travis-CI
image
image

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
sjanzou: removed with last commit

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
sjanzou: Resolved with last commit

test/ssc_test/cmod_pvsamv1_test.cpp
sjanzou: Reverted to develop and SEH windows errors occur with or without changes

test/ssc_test/cmod_pvwattsv7_test.cpp
sjanzou: Resolved with last commit

test/ssc_test/cmod_windpower_test.cpp
sjanzou: reverted to develop tests

sjanzou:
Latest Windows results with all test (usual widows weather_file_inputs test failure - usually different each time)
image

PR #598, opened 2021-07-29
Test bionic travis container

brtietz:
Creating a pull request to force travis to run the tests. Will review and merge if this passes.


brtietz:
Upgrading did not solve crash. May consider doing this in the future, but needs more coordination with other repositories.

PR #602, opened 2021-08-13
Refactor battery enums

brtietz:
Reduce the number of dispatch algorithms to places where the underlying algorithm is different, and add new enums for weather (generation) forecast and load forecasting. These are fully fleshed out for FOM, BTM will come in another pull request. Pairs with SAM PR: https://github.com/NREL/SAM/pull/681

New forecast enum numbers:
FOM:
- 0 - Economic
- 1 - PV Smoothing
- 2 - Custom
- 3 - Manual

BTM:
- 0 - Peak Shaving
- 1 - Input Grid Target
- 2 - Custom
- 3 - Manual
- 4 - Price signals


ssc/cmod_pvsamv1.cpp
```diff
@@ -2238,6 +2250,16 @@ void cm_pvsamv1::exec()
                 }
 			}
 
+            // Re-compute PV AC forecast for AC connected batteries
```
brtietz, line 2253: @sjanzou This code change is the reason the number of violations increased in the pv smoothing test. The new AC look ahead number includes additional losses and therefore I'd argue it's more accurate. Let me know if it makes sense to update the numbers in the test or solve the test failure some other way.

sjanzou (approved):
looks good!

ssc/cmod_pvsamv1.cpp
sjanzou: We should probably re-evaluate with the code from EPRI - although it will not include the additional losses - maybe if we use the updated pv profile as input to the Python from EPRI - we can match the violation count to the new SAM count...

ssc/cmod_battery.cpp
```diff
@@ -1231,29 +1232,25 @@ void battstor::parse_configuration()
         prediction_index = 0;
         if (batt_meter_position == dispatch_t::BEHIND)
         {
-            if (batt_dispatch == dispatch_t::LOOK_AHEAD || batt_dispatch == dispatch_t::MAINTAIN_TARGET || batt_dispatch == dispatch_t::FORECAST)
+            if (batt_dispatch == dispatch_t::PEAK_SHAVING || batt_dispatch == dispatch_t::MAINTAIN_TARGET || batt_dispatch == dispatch_t::FORECAST)
             {
-                look_ahead = true;
+                look_ahead = batt_weather_forecast == dispatch_t::WEATHER_FORECAST_CHOICE::WF_LOOK_AHEAD;
+                look_behind = batt_weather_forecast == dispatch_t::WEATHER_FORECAST_CHOICE::WF_LOOK_BEHIND;
+                input_forecast = batt_weather_forecast == dispatch_t::WEATHER_FORECAST_CHOICE::WF_CUSTOM;
```
dguittet, line 1239: I think the standard semantics for this is the `case` keyword. That'll also make sure `batt_weather_forecast` is within the correct bounds for `dispatch_t::WEATHER_FORECAST_CHOICE1

ssc/cmod_battery.cpp
```diff
@@ -1231,29 +1232,25 @@ void battstor::parse_configuration()
...
-            }
-            else if (batt_dispatch == dispatch_t::FOM_LOOK_BEHIND) {
-                look_behind = true;
-            }
-            else if (batt_dispatch == dispatch_t::FOM_FORECAST) {
-                input_forecast = true;
+            if (batt_dispatch == dispatch_t::FOM_AUTOMATED_ECONOMIC || batt_dispatch == dispatch_t::FOM_PV_SMOOTHING) {
+                look_ahead = batt_weather_forecast == dispatch_t::WEATHER_FORECAST_CHOICE::WF_LOOK_AHEAD;
+                look_behind = batt_weather_forecast == dispatch_t::WEATHER_FORECAST_CHOICE::WF_LOOK_BEHIND;
+                input_forecast = batt_weather_forecast == dispatch_t::WEATHER_FORECAST_CHOICE::WF_CUSTOM;
```
dguittet, line 1253: same, `look_ahead`, etc should all be False to start then one should be toggled on

ssc/cmod_pvsamv1.cpp
```diff
@@ -1093,6 +1085,22 @@ void cm_pvsamv1::exec()
         if (!Simulation->annualSimulation)
             throw exec_error("pvsamv1", "The PV+Battery configuration requires a simulation period that is continuous over one or more years.");
 
+        int batt_forecast_choice = as_integer("batt_dispatch_wf_forecast_choice");
+        if (is_assigned("batt_pv_clipping_forecast")) {
+            p_pv_clipping_forecast = as_vector_ssc_number_t("batt_pv_clipping_forecast");
```
dguittet, line 1090: I remember you asking about lifetime forecasts, but here`batt_pv_clipping_forecast` is 1 year only? Please specify in var table if its 1 yr forecast, if it isn't already

dguittet (approved):
Minor requests from me. After those, I think it's good to go!

ssc/cmod_battery.cpp
brtietz: These are flagged to false in the header file. I'll proceed with the case statement for clarity.

PR #603, opened 2021-08-13
Me fixes

mjprilliman:
-Cleanup of var tables for Marine Energy models


cpaulgilman (approved):
I set the UI Group for all cost inputs so they appear organized in input variable lists for LK editor, parametrics, etc.

PR #606, opened 2021-08-19
Weather and load forecasts for BTM

brtietz:
Add options for load forecasts for behind the meter systems. Load forecast functions the same as actual load - can be either hourly for a single year (8760) or match the weather file timestep. Interpolation is handled by single_year_to_lifetime_interpolated.

Additionally, extend forecast options to generic system.

Pairs with SAM pull request: https://github.com/NREL/SAM/pull/688


ssc/cmod_battery.cpp
```diff
@@ -1352,6 +1357,39 @@ void battstor::initialize_automated_dispatch(std::vector<ssc_number_t> pv, std::
...
+                    for (size_t idx = 0; idx != nrec; idx++) {
+                        load_prediction.push_back(load[idx]);
+                    }
+                }
+            }
+            else if (load_look_behind) {
+                // day one is zeros
+                for (size_t idx = 0; idx != 24 * step_per_hour; idx++)
+                {
+                    load_prediction.push_back(0);
```
dguittet, line 1374: why isn't this from the last 24 hours of the load forecast?

ssc/cmod_battery.cpp
```diff
@@ -1944,15 +1985,57 @@ class cm_battery : public compute_module
...
+            if (is_assigned("batt_load_ac_forecast"))
+            {
+                p_load_forecast_in = as_vector_ssc_number_t("batt_load_ac_forecast");
+                size_t nload = p_load_forecast_in.size();
+                if (nload == 1) {
+                    // Length 1 is "empty" to UI lk
+                    p_load_forecast_in.clear();
+                }
+            }
+            if (p_load_forecast_in.size() > 0) {
```
dguittet, line 2014: shouldn't this be `p_load_forecast_in.size() < weather file rec * analysis period`? a full life time forecast also has `p_load_forecast_in.size()` > 0

ssc/cmod_pvsamv1.cpp
```diff
@@ -1165,6 +1181,26 @@ void cm_pvsamv1::exec()
         p_load_full = p_load_in;
     }
 
+    if (en_batt) {
+        if (p_load_forecast_in.size() > 0) {
```
dguittet, line 1185: same here for size of p_load_forecast_in

ssc/cmod_battery.cpp
brtietz: I was matching the convention from the weather file forecast: https://github.com/NREL/ssc/blob/fcff96bd16bf1a8dbc38966a94bca69d2ab99ec0/ssc/cmod_battery.cpp#L1328 is it worth changing both?

ssc/cmod_battery.cpp
brtietz: The main load functions don't allow for lifetime loads. This is responding to the clear() on line 2011.

The length of load is far better enforced in pvsamv1 than it is in cmod_battery. I'll copy some of those functions over.

ssc/cmod_pvsamv1.cpp
brtietz: See https://github.com/NREL/ssc/blob/fcff96bd16bf1a8dbc38966a94bca69d2ab99ec0/ssc/cmod_pvsamv1.cpp#L1084 for the length constraints that have been enforced prior to this line.

ssc/cmod_battery.cpp
dguittet: I think so. I'd imagine if we're in year 2, and if we were doing lookahead, we'd use jan data for dec 31, and if doing lookbehind, we'd do the counterpart. This keeps continuity. What do you think?

ssc/cmod_battery.cpp
dguittet: ok thanks

dguittet (approved):

PR #608, opened 2021-08-24
Battstor outputs for batt_life_model

dguittet:
Fix some inputs to run battery lifetime models from the GUI


dguittet:
Merged in latest from develop without merge conflict. Not sure how to automate the spacing in the vartables. For editorconfig, we're using the one that's included from wxWidgets but the settings are fairly short and don't account for specific use cases like this.

PR #609, opened 2021-08-24
More flexible powerflows to match REopt output

brtietz:
Restrictions in SAM's powerflow code required that AC connected, behind the meter batteries only charge when the renewable system's power exceeded load, and only discharge when the load exceeded the renewable system's power. DC connected batteries were subject to the latter restriction as well.

In order to accurately process optimal dispatch from REopt, these powerflow constraints need to be lifted, as some timesteps will violate these constraints. This increases the power that SAM deliveres from the battery to the load while running a REopt dispatch case from 93% of REopt's prediction to 99.94%  - and should help with other custom dispatch cases as well.

Pairs with SAM PR: https://github.com/NREL/SAM/pull/692


ssc/cmod_battery.cpp
```diff
@@ -149,11 +149,12 @@ var_info vtab_battery_inputs[] = {
...
+    { SSC_INPUT,        SSC_NUMBER,     "batt_dispatch_auto_can_charge",               "System charging allowed for automated dispatch?",            "0/1",       "",                "BatteryDispatch",       "",                           "",                             "" },
+    { SSC_INPUT,        SSC_NUMBER,     "batt_dispatch_auto_can_clipcharge",           "Battery can charge from clipped power for automated dispatch?", "0/1",   "",                 "BatteryDispatch",       "",                           "",                             "" },
+    { SSC_INPUT,        SSC_NUMBER,     "batt_dispatch_charge_only_system_exceeds_load",  "Battery can charge from system only when system output exceeds load", "0/1",   "",        "BatteryDispatch",       "en_batt=1&batt_meter_position=0",                           "",                             "" },
+    { SSC_INPUT,        SSC_NUMBER,     "batt_dispatch_discharge_only_load_exceeds_system","Battery can discharge battery only when load exceeds system output", "0/1",   "",        "BatteryDispatch",       "en_batt=1&batt_meter_position=0",                           "",                             "" },
```
dguittet, line 157: the dischargeOnlyLoadExceedSystem seems to be an option for AC and DC batteries, but here it says `batt_meter_position=0`

ssc/cmod_battwatts.cpp
```diff
@@ -241,6 +241,8 @@ battwatts_create(size_t n_recs, size_t n_years, int chem, int meter_pos, double
     }
     batt_vars->batt_dispatch_auto_can_charge = true;
     batt_vars->batt_dispatch_auto_can_gridcharge = true;
+    batt_vars->batt_dispatch_charge_only_system_exceeds_load = true;
+    batt_vars->batt_dispatch_discharge_only_load_exceeds_system = true;
```
dguittet, line 245: Perhaps this should be false so the ReOpt runs work better even in Battwatts

test/shared_test/lib_battery_powerflow_test.cpp
```diff
@@ -2299,3 +2299,431 @@ TEST_F(BatteryPowerFlowTest_lib_battery_powerflow, DC_interconnection_limit_exce
...
+
+    EXPECT_NEAR(m_batteryPower->powerBatteryAC, -50, error);
+    EXPECT_NEAR(m_batteryPower->powerSystemToLoad, 50, error);
+    EXPECT_NEAR(m_batteryPower->powerSystemToBattery, 50, error);
+    EXPECT_NEAR(m_batteryPower->powerGridToBattery, 0, error);
+    EXPECT_NEAR(m_batteryPower->powerGridToLoad, 0, error);
+    EXPECT_NEAR(m_batteryPower->powerSystemToGrid, 0, error);
+    EXPECT_NEAR(m_batteryPower->powerBatteryToLoad, 0, error);
+    EXPECT_NEAR(m_batteryPower->powerConversionLoss, 2.0, error);
+    EXPECT_NEAR(m_batteryPower->powerSystemLoss, 0.0, error);
```
dguittet, line 2327: can we have some power battery to grid tests?

test/shared_test/lib_battery_powerflow_test.cpp
```diff
@@ -2299,3 +2299,431 @@ TEST_F(BatteryPowerFlowTest_lib_battery_powerflow, DC_interconnection_limit_exce
...
+
+    double gen = m_batteryPower->powerSystem + m_batteryPower->powerBatteryAC;
+    EXPECT_NEAR(m_batteryPower->powerGeneratedBySystem, gen, error);
+    EXPECT_NEAR(m_batteryPower->powerLoad, 50, error);
+
+    // Try to charge more than is available from PV, allowed due to powerflow
+    m_batteryPower->powerBatteryDC = -100;
```
dguittet, line 2334: what happens if powerBatteryDC > pv production?

dguittet (changes requested):
Curious why isn't chargeOnlySystemExceedLoad for DC batteries, but then in the tests both `chargeOnlySystemExceedLoad` and `dischargeOnlyLoadExceedSystem` can be false? In both the cmod_battery var_table where meter_position is fixed and in the powerflow calculations where I didn't see changes to the charging part of the code.

brtietz:
> chargeOnlySystemExceedLoad for DC batteries,

I'll remove this variable from the DC tests. The DC code is structured in a way such that this variable is effectively always false. I worry the efficiency calculations required to enforce this on the DC side would be extremely difficult, but happy to discuss if it's important.

ssc/cmod_battery.cpp
brtietz: `batt_meter_position=0` is behind the meter, not AC/DC.

shared/lib_battery_powerflow.cpp
```diff
@@ -333,22 +355,31 @@ void BatteryPowerFlow::calculateACConnected()
...
+            }
+            else {
+                P_fuelcell_to_load_ac = std::fmin(P_fuelcell_ac, calc_load_ac - P_pv_to_load_ac);
+                P_batt_to_load_ac = std::fmin(P_battery_ac - P_system_loss_ac, calc_load_ac - P_pv_to_load_ac - P_fuelcell_to_load_ac);
+            }
         }
         else {
-            P_fuelcell_to_load_ac = std::fmin(P_fuelcell_ac, calc_load_ac - P_pv_to_load_ac);
-            P_batt_to_load_ac = std::fmin(P_battery_ac - P_system_loss_ac, calc_load_ac - P_pv_to_load_ac - P_fuelcell_to_load_ac);
+            P_batt_to_load_ac = P_battery_ac;
```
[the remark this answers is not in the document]
brtietz: I was trying to avoid doing batt to grid here, as that increases the scope to the dispatch code. The fmin addition is a good idea, I'll get that in the next push.

test/shared_test/lib_battery_powerflow_test.cpp
brtietz: Should always be zero until we tackle https://github.com/NREL/SAM/issues/338 - worth adding anyway?

ssc/cmod_battery.cpp
dguittet: Ah right thanks

test/shared_test/lib_battery_powerflow_test.cpp
dguittet: I didn't realize this didn't allow exporting to grid, no need adding for now then!

test/shared_test/lib_battery_powerflow_test.cpp
brtietz: Test added in latest push. Like the DC case, since grid charging is not allowed powerBatteryDC is reduced to the PV output.

ssc/cmod_battwatts.cpp
brtietz: Changed to false in latest push. Corresponding annual energy and lcoe changes pushed to paired SAM branch.

dguittet (approved):

PR #611, opened 2021-08-26
Improve cycle rainflow model

brtietz:
Move the cycle life model away from using the average range for degradation. This means that a large number of small cycles will no longer impede degradation due to a large cycle. Add a table that allows the cycle life algorithm to sort cycles by depth of discharge, and apply the appropriate number of cycles for that curve.

Results of cycling tests in SAM-analysis/rainflow-validation from develop, where low is 2000 cycles of 20% DOD, and high is specified:

> Cycle life, high = 40% DOD
High, then low, cycle life 61.50770913594484
Low, then high, cycle life 61.70228902887189

>Cycle life, high = 80% DOD
High, then low, cycle life 56.293643046055735
Low, then high, cycle life 61.25222517557856

>Cycle life, high = 100% DOD
High, then low, cycle life 49.78814668141695
Low, then high, cycle life 61.069726777259305

Results from this branch:
>Cycle life, high = 40% DOD
High, then low, cycle life 61.57734478757392
Low, then high, cycle life 61.26046846007392

>Cycle life, high = 80% DOD
High, then low, cycle life 57.87701862625535
Low, then high, cycle life 55.39861991443408

>Cycle life, high = 100% DOD
High, then low, cycle life 53.47991635538674
Low, then high, cycle life 53.98541896613945

So "low then high" is much more affected by the high DOD cycles, as would be expected.

Test changes are due to updated cycle life model, where the test conducts enough cycles that it would be sensitive to this change.

Paired SAM PR with updated results from test_script_ow.lk: https://github.com/NREL/SAM/pull/695


shared/lib_battery_lifetime_calendar_cycle.cpp
```diff
@@ -42,9 +42,25 @@ void lifetime_cycle_t::initialize() {
...
+        double dod = params->cal_cyc->cycling_matrix.at(i, calendar_cycle_params::DOD);
+        if (std::find(DOD_levels.begin(), DOD_levels.end(), dod) == DOD_levels.end()) {
+            DOD_levels.push_back(dod);
+        }
+    }
+    std::sort(DOD_levels.begin(), DOD_levels.end());
+    state->cycle->cycle_counts.resize_fill(DOD_levels.size(), 2, 0.0);
+    for (size_t i = 0; i < DOD_levels.size(); i++) {
+        state->cycle->cycle_counts.set_value(DOD_levels[i], i, cycle_state::DOD);
+    }
```
dguittet, line 61: does it make sense to restrict the reporting of the cycle_count matrices' DOD levels to what is provided in the cycling_matrix? This isn't how the cycle model actually works because it interpolates between DODs. I may prefer a fixed number of equal-width bins for reporting (like 10: 0-9, 10-19, ... 90-99).

ssc/cmod_battery_stateful.cpp
```diff
@@ -244,6 +245,7 @@ void write_battery_state(const battery_state& state, var_table* vt) {
     if (choice == lifetime_params::CALCYC) {
         vt->assign_match_case("q_relative_calendar", lifetime->calendar->q_relative_calendar);
         vt->assign_match_case("dq_relative_calendar_old", lifetime->calendar->dq_relative_calendar_old);
+        vt->assign_match_case("cycle_counts", lifetime->cycle->cycle_counts);
```
dguittet, line 248: since these are matrices, they'll have to follow the protection against discrepancy between empty arrays and matrix_t always having a num value-- like in line 236

ssc/cmod_battery_stateful.cpp
```diff
@@ -355,6 +357,7 @@ void read_battery_state(battery_state& state, var_table* vt) {
     if (choice == lifetime_params::CALCYC) {
         vt_get_number(vt, "q_relative_calendar", &lifetime->calendar->q_relative_calendar);
         vt_get_number(vt, "dq_relative_calendar_old", &lifetime->calendar->dq_relative_calendar_old);
+        vt_get_matrix(vt, "cycle_counts", lifetime->cycle->cycle_counts);
```
dguittet, line 360: same here

dguittet (changes requested):
Great improvement!

ssc/cmod_battery_stateful.cpp
dguittet: Nevermind, this matrix is never empty so this code and the one below are fine as is

shared/lib_battery_lifetime_calendar_cycle.cpp
brtietz: This is used by the model in lines 183 through 192 to determine how many cycles to apply in the bilinear interpolation. I was assuming that the user would specify any non-linear dynamics on different DOD vs cycle curves, and that maintaining the bins specified on chart would allow us to hit roughly the right point on each line.

I'm open to reporting smaller bins in a second output variable, but I think the correlation with the chart is worth maintaining if we're using cycle_counts in the degradation computation.

dguittet (approved):

PR #612, opened 2021-08-30
Lcoe lppa energy sales

mjprilliman:
-Change LCOE, LPPA denominator to use energy sales rather than net energy
-Address all main financial models
-Add cash flows for energy sales and energy purchases where they did not previously exist

### DIFF
--- a/ssc/cmod_levpartflip.cpp
+++ b/ssc/cmod_levpartflip.cpp
@@ -884,7 +884,10 @@ enum {
 	CF_battery_replacement_cost_schedule,
 	CF_battery_replacement_cost,
 
-	
+    CF_energy_sales,
+    CF_energy_sales_value,
+    CF_energy_purchases,
+    CF_energy_purchases_value,
 
     CF_energy_charged_grid,
     CF_energy_charged_pv,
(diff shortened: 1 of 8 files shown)


ssc/cmod_levpartflip.cpp
```diff
@@ -884,7 +884,10 @@ enum {
 	CF_battery_replacement_cost_schedule,
 	CF_battery_replacement_cost,
 
-	
+    CF_energy_sales,
+    CF_energy_sales_value,
+    CF_energy_purchases,
+    CF_energy_purchases_value,
```
brtietz, line 890: Usually best practice is to add new values to the end of an enumeration. I'm concerned that adding these in the middle might have thrown off some of the cash flow sheets. This is from PVWatts, partnership flip with no debt:

image

ssc/cmod_mhk_tidal.cpp
```diff
@@ -169,32 +169,49 @@ class cm_mhk_tidal : public compute_module
...
         max_velocity = tidal_resource_matrix.at(size_t(number_rows) - 1, 0);
 
-		for (int i = 0; i < number_rows; i++) {
+        for (int i = 0; i < number_rows; i++) {
             size_t n = i;
-            if (tidal_resource_matrix.at(n, 1) != 0 && tidal_resource_matrix.at(n - 1, 1) == 0 && n != 0)
-            {
-                tidal_resource_start_velocity = tidal_resource_matrix.at(n, 0);
+            if (i != 0) {
+                if (tidal_resource_matrix.at(n, 1) != 0 && tidal_resource_matrix.at(n - 1, 1) == 0)
```
brtietz, line 175: Can you explain these changes in the pull request title? I'm all for fixing tests, but this seems a little out of place in this PR.

brtietz (changes requested):
I suspect some of the issues in https://github.com/NREL/SAM/pull/696 could be caused by the enum ordering, but @cpaulgilman will know better how to fix those cash flow sheets.

ssc/cmod_levpartflip.cpp
mjprilliman: At some point the outputs for cf_annual_costs and npv_annual_costs were lost from the var_table. I tried to put them back for FOM at this point but I will need to check all financial models.

PR #617, opened 2021-09-22
Move batt stateful sizing to ssc equations

brtietz:
Convert two functions currently in pysam's battery tools to ssc equations such that they can be used by the API in all languages.

Future pull requests will add additional cmod_battery functions.

### DIFF
--- a/ssc/cmod_battery_eqns.cpp
+++ b/ssc/cmod_battery_eqns.cpp
@@ -1,0 +1,80 @@
+/**
+BSD-3-Clause
+Copyright 2019 Alliance for Sustainable Energy, LLC
+Redistribution and use in source and binary forms, with or without modification, are permitted provided
+that the following conditions are met :
+1.	Redistributions of source code must retain the above copyright notice, this list of conditions
+and the following disclaimer.
+2.	Redistributions in binary form must reproduce the above copyright notice, this list of conditions
+and the following disclaimer in the documentation and/or other materials provided with the distribution.
+3.	Neither the name of the copyright holder nor the names of its contributors may be used to endorse
+or promote products derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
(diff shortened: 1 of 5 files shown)


ssc/cmod_battery_eqns.cpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    double mass, surface_area, original_capacity, desired_capacity, module_capacity, module_surface_area;
+
+    vt_get_number(vt, "mass", &mass);
+    vt_get_number(vt, "surface_area", &surface_area);
+    vt_get_number(vt, "original_capacity", &original_capacity);
+    vt_get_number(vt, "desired_capacity", &desired_capacity);
+
+    double mass_per_specific_energy = mass / original_capacity;
```
dguittet, line 61: There is an edge case where the original_capacity is 0 and then resized to non-zero. This happens when you size the battery to 0, perhaps as part of a loop across a range of numbers, then the mass and surface area will get set to 0 also. Then for later sizing calls, the function will no longer be able to use prior information about the mass to calculate an appropriate mass, and here it'll cause a divide by 0.

Other cmods are not able to run with a strictly 0 system capacity (pvsamv1, windpower) while others can (pvwatts). While this is a still-unfixed issue in the PySAM code, I think the simplest way to handle this is to not let the desired_capacity be 0.

dguittet (changes requested):

ssc/cmod_battery_eqns.cpp
brtietz: Great point! In the latest commit I used the same error convention as cmod_windpower_eqns - writing the error to an "error" variable in the vartable. Am I correct to assume it's the responsibility of the calling code to check this value? When I tried to throw an exec error, it crashed Julia. Given the difficulty of error handling in the SDK, I assume that's not a valid option.

dguittet: Yeah, you can't throw exceptions across a DLL boundary so that's not possible. The error will have to be done via a return string. Perhaps we could change the function signature of SSC equations so the return type is bool instead of void. I don't think that would be too difficult and would make error handling easier

dguittet: Let me know if you want me to change the function signature to bool

dguittet: Apologies for the delay, but I've fixed up all the equations: changing the return type, fixing up the error handling from exceptions to error strings.

dguittet:
I'll re-review once tests are done

PR #619, opened 2021-09-28
Add missing vtab_lcos_input so that cash_loan will run from lk script

sjanzou:


shared/lib_battery_powerflow.cpp
```diff
@@ -490,7 +490,8 @@ void BatteryPowerFlow::calculateACConnected()
         if ((fabs(P_grid_ac - P_grid_to_load_ac) > tolerance) && (-P_grid_ac > P_grid_to_load_ac)) {
             P_battery_ac = P_pv_ac - P_pv_to_grid_ac - P_pv_to_load_ac - P_system_loss_ac;
             P_battery_ac = P_battery_ac > 0 ? P_battery_ac : 0; // Don't swap from charging to discharging
-            m_BatteryPower->powerBatteryDC = -P_battery_ac * m_BatteryPower->singlePointEfficiencyACToDC;
+//            m_BatteryPower->powerBatteryDC = -P_battery_ac * m_BatteryPower->singlePointEfficiencyACToDC;
+            m_BatteryPower->powerBatteryDC = P_battery_ac * m_BatteryPower->singlePointEfficiencyACToDC;
```
brtietz, line 494: This code will only run when the battery is charging (see line 489), therefore the sign on line 493 is correct. We don't want this code swapping from charging to discharging, which is what line 494 now does. 

If there's a less confusing way to do the sign swap between 491 and 493, I'm open to that, but at the end of this section powerBatteryDC needs to be a negative number. The StackOverflow issue needs to be solved between lines 304 and 385. (I'll try to narrow that down today)

brtietz: My hunch on the lines was incorrect - proposed fix here: https://github.com/NREL/ssc/pull/620/files

Happy to accept a PR with just the "batt_annual_charge_energy" change.

brtietz:
I cherry-picked https://github.com/NREL/ssc/pull/619/commits/23e4bacf5c3c1371cdfc2e0a1aecdac4a9c80d6d over to develop so we get that change included while I debug the powerflow stack overflow.

sjanzou:
Great!

I will delete pull request 619 if you have not already.

Thanks,
steve

PR #618, opened 2021-09-27
Connect outage vars to powerflow

brtietz:
Adds crit_load_escalation to the relevant cmods. Brings the variables from the vartables into battery->advance such that the powerflow variables can use the outage code.

Needs to be paired with a SAM PR based on critical-load-ui branch, @cpaulgilman to create.

A future PR (ssc branch dispatch_for_grid_outage) will implement the outage behavior in dispatch, and clean up edge cases like the inverter night time loads.

Via email we had proposed making grid_outage and crit_load SSC outputs - should that happen on this PR or a different PR? If this one, which cmods are responsible for the output?

PR #620, opened 2021-09-28
add tolerance to grid charging check to allow for floating point issues

brtietz:
A floating point error (~1e-15) was causing a stack overflow in the powerflow code. Add a tolerance to fix the issue.

floating_point_stack_overflow

Alternative fix would be to set gen to zero if it's smaller than some tolerance. I think "tolerance" in the powerflow code is too high for this approach, I'd want 1e-7 or smaller for this approach.

Test with
[Test_Case.zip](https://github.com/NREL/ssc/files/7246382/Test_Case.zip)

### DIFF
--- a/shared/lib_battery_powerflow.cpp
+++ b/shared/lib_battery_powerflow.cpp
@@ -469,7 +469,7 @@ void BatteryPowerFlow::calculateACConnected()
 
     // Error checking trying to charge from grid when not allowed
     if (!m_BatteryPower->canGridCharge && P_battery_ac < -tolerance) {
-        if ((fabs(P_grid_ac - P_grid_to_load_ac) > tolerance) && (-P_grid_ac > P_grid_to_load_ac)) {
+        if ((fabs(P_grid_ac - P_grid_to_load_ac) > tolerance) && (-P_grid_ac > P_grid_to_load_ac + tolerance)) {
             P_battery_ac = P_pv_ac - P_pv_to_grid_ac - P_pv_to_load_ac - P_system_loss_ac;
             P_battery_ac = P_battery_ac > 0 ? P_battery_ac : 0; // Don't swap from charging to discharging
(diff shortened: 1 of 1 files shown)


sjanzou (approved):

sjanzou:
Still failing for case 3 of attached file, hangs at 66.6%
image

Case 3 of attached file - grid target 0 and time series critical load hangs as above image with these changes...
[Test_Case.zip](https://github.com/NREL/ssc/files/7247294/Test_Case.zip)

shared/lib_battery_powerflow.cpp
```diff
@@ -469,7 +469,7 @@ void BatteryPowerFlow::calculateACConnected()
 
     // Error checking trying to charge from grid when not allowed
     if (!m_BatteryPower->canGridCharge && P_battery_ac < -tolerance) {
-        if ((fabs(P_grid_ac - P_grid_to_load_ac) > tolerance) && (-P_grid_ac > P_grid_to_load_ac)) {
+        if ((fabs(P_grid_ac - P_grid_to_load_ac) > tolerance) && (-P_grid_ac > P_grid_to_load_ac + tolerance)) {
             P_battery_ac = P_pv_ac - P_pv_to_grid_ac - P_pv_to_load_ac - P_system_loss_ac;
```
sjanzou, line 473: Fixes peak shaving stack overflow but not the grid target of zero with time series resilency calculations per case 3 in the attached sam project file in the comments

brtietz: Case three has different problems depending on which branch you're on. On connect_outage_vars... it gets into the stack overflow because the unmet critical load is drawn from the grid. On dispatch_for_grid_outage it fails because the resiliency runner isn't properly updating the battery power flow variables as it steps forward through the hypothetical outage.

I'm inclined to fix the problem on dispatch_for_grid_outage (since that branch has already fixed the first problem) but what do we do about #618 in that case?

PR #623, opened 2021-10-05
Add critical load, and accumulators for unmet load and percentages

sjanzou:


sjanzou:
Critical loads and percentages added on monthly and annual basis for Puerto Rico work for minimizing unmet load percentages on different time scales.

Added outputs when there is a critical load are shown in red
image

brtietz (approved):
Looks great, thanks for adding all of this!

Issue #631, opened 2021-10-27
"Inverter total power loss" can be negative when DC connected batteries grid charge


brtietz:
**Describe the bug**
DC connected batteries can cause the "inverter total power loss" metric to display a negative number when grid charging

**To Reproduce**
1. Set up a DC connected battery with grid charging
2. Run the simulation
3. Look at the data table output for inverter total power loss during a step with grid charging
4. Observe that the number is negative

**Expected behavior**
It looks like a calculation needs to use fabs that isn't. This number should always be non-negative, like many of the other loss categories.

**Screenshots**
![image](https://user-images.githubusercontent.com/5530592/139145514-c00f449d-582e-4b21-b9d5-0026104e795e.png)

Observed on develop, likely older than that.

brtietz:
Closed in https://github.com/NREL/ssc/commit/9cf27a1bb29f9488eb1b13f95a3ce35315d3ea9b

Issue #590, opened 2021-07-07
Battery losses do not affect resilience runner


brtietz:
**Describe the bug**
Battery losses are not applied to the calculations of "hours of autonomy during outage" in the resiliency runner. The losses are applied to a source (PV, grid, battery) in the powerflow code, and the resiliency runner does not interface with this code. This generates behavior such as:

(a) The no-loss case and the loss case having the same hours of autonomy (given the same SOC at that step).
(b) Overestimating the survival hours. For example, a battery with 3.5 kWh of charge and a 1 kW discharging loss reports 5 hours of survival.

**To Reproduce**
1. Load the attached SAM file: 
[losses_bug.zip](https://github.com/NREL/ssc/files/6780086/losses_bug.zip)
2. Run both cases
3. Observe the hours of survival issues stated above.

**Expected behavior**
The resiliency code should account for battery losses.

Exists in 2020.11.29, and also in develop. The approach we took in solving https://github.com/NREL/ssc/issues/379 did not touch the code for the resiliency runner.

brtietz:
This fix is in the dispatch_for_grid_outage branch, which should be ready for a pull request next week.

brtietz:
Closed in https://github.com/NREL/ssc/commit/9cf27a1bb29f9488eb1b13f95a3ce35315d3ea9b

PR #632, opened 2021-10-28
Update dispatch for grid outages

brtietz:
Brings the grid outage code into the mainline battery dispatch. Allows the resiliency runner and the outage dispatch to use the same code. Pairs with SAM pull request https://github.com/NREL/SAM/pull/722, please pull dispatch_for_grid_outage branch to test.

Solves https://github.com/NREL/ssc/issues/631 and https://github.com/NREL/ssc/issues/590

- Adds a second min SOC variable "batt_minimum_outage_SOC", so that the battery can have a different min SOC during nominal operation and an outage.
  - Also prevents the SOC from going to zero during the default outage behavior, allowing the battery to black start the PV inverter in the morning
- Add new output variables:
  -  batt_to_system_load - tracks the battery covering the inverter night time losses
  - outage_losses_unmet - tracks unmet losses separately from the unmet critical load
- Updates the DC resiliency runner to better account for efficiency updates to meet the critical load
- Modifies the dispatch planning code such that the dispatch algorithms re-plan after an outage
- Modifies the losses code in pvsamv1 so that the resiliency code can dispatch against AC losses
- Updates the test values to account for all of the above

Recommended tests:
- Ensure that grid usage is zero during outage periods
- Ensure that the resiliency runner predicts when the battery will hit the minimum state of charge
- Ensure the outage code works with pvsamv1, battwatts, generic battery and fuel cell batttery


brtietz:
The file I've been using to test this is attached here:
[test_grid_outage_2.zip](https://github.com/NREL/ssc/files/7456215/test_grid_outage_2.zip)

shared/lib_battery_dispatch_automatic_btm.h
```diff
@@ -80,7 +80,8 @@ class dispatch_automatic_behind_the_meter_t : public dispatch_automatic_t
         std::vector<double> battCycleCost,
         double interconnection_limit,
         bool chargeOnlySystemExceedLoad,
-        bool dischargeOnlyLoadExceedSystem
+        bool dischargeOnlyLoadExceedSystem,
+        double SOC_min_outage
```
sjanzou, line 84: at some point we should stick with one naming convention and c++ instead of java structure. Just a comment and nothing specific to this pull request

shared/lib_shared_inverter.cpp
```diff
@@ -375,7 +375,10 @@ void SharedInverter::convertOutputsToKWandScale(double tempLoss, double powerAC_
     powerConsumptionLoss_kW *= m_numInverters * util::watt_to_kilowatt;
     powerNightLoss_kW *= m_numInverters * util::watt_to_kilowatt;
     powerTempLoss_kW = tempLoss * m_numInverters * util::watt_to_kilowatt;
-    powerLossTotal_kW = powerDC_kW - powerAC_kW;
+    if (powerDC_kW < 0.0)
+        powerLossTotal_kW = fabs(powerDC_kW) - fabs(powerAC_kW); // DC connected grid charging
```
sjanzou, line 379: Are the charging and discharging losses all accounted for in the difference here?

ssc/cmod_battwatts.cpp
```diff
@@ -191,6 +191,13 @@ battwatts_create(size_t n_recs, size_t n_years, int chem, int meter_pos, double
     batt_vars->batt_voltage_choice = voltage_params::MODEL;
     batt_vars->batt_voltage_matrix = util::matrix_t<double>();
 
+    // Power converters and topology
+    batt_vars->batt_topology = ChargeController::AC_CONNECTED;
+    batt_vars->batt_ac_dc_efficiency = 96;
```
sjanzou, line 196: Is this really stating that changing and discharging efficiency is the same?

sjanzou (changes requested):
I tested with your file posted and went through some other cases and found some interesting plots and the annoying desired duration in hours bug in the parametrics, kWh can be used with additional calculations
image

with some results that @n8blair might be interested in
image

@cpaulgilman may have an issue with the limits on the Battery page in the fuel cell and the dispatch page in the battery configuration (I do not see a way around that) 
image

Hmmm.... Fuel cell is failing with an 8760 weather file and an 8760 load:
image

File for testing with failing fuel cell configuration
[test_grid_outage_2_fuelcell.zip](https://github.com/NREL/ssc/files/7459903/test_grid_outage_2_fuelcell.zip)

shared/lib_shared_inverter.cpp
brtietz: This is the fix for https://github.com/NREL/ssc/issues/631

The more detailed loss accounting is in lib_battery_power_dispatch.cpp, scattered throughout the calculation functions. There's also code in cmod_battery that does updates for the actual AC losses in pvsamv1. Many of the changes in those three files are related to loss computation.

brtietz:
The issue with fuel cell and generic system was that battstor was overwriting the values of crit_load before cmod_battery had a chance to read them. I've fixed that issue, but a few others have cropped up in the ssc tests. I'll fix those tomorrow, and it'll hopefully be ready for another round of review in the afternoon.

dguittet:
> The file I've been using to test this is attached here: [test_grid_outage_2.zip](https://github.com/NREL/ssc/files/7456215/test_grid_outage_2.zip)

Thanks for adding. When I opened this, there were no critical loads specified or outage time periods (the 8760 was all 0), is that expected?

dguittet:
I built on `dispatch_for_grid_outage` for both sam & ssc, but when I save the file and re-open it, the critical loads & grid outage timeseries inputs are back to the defaults. So something in the saving and reopening process isn't working for me. Is it just me?

I'm adding some high-level feedback in the comments before digging into the code:

1. image
It'll be nice to have these e-16 numbers rounded to 0. Probably by the ssc losses and/or powerflow code

2. image
I think the inverter is supposed to be off at the start of the year when there's no power production, but at the start of the second year, this behavior seems to revert to inverter nighttime losses again.

3. image
This label should be "power" not "energy"

brtietz:
@dguittet Great catch on all of the above. Those should be fixed in the latest commit.

test/shared_test/lib_battery_dispatch_automatic_btm_test.cpp
```diff
@@ -753,3 +751,505 @@ TEST_F(AutoBTMTest_lib_battery_dispatch, DispatchAutoBTMCustomWExcessPV) {
...
+            pv_prediction.push_back(0);
+        }
+
+        if (h == 6) {
+            load_prediction.push_back(600);
+        }
+        else if (h > 16) {
+            load_prediction.push_back(700);
+        }
+        else if (h > 18) {
```
dguittet, line 853: this h>18 condition never comes on because anything > 18 is > 16 so will be set to 700

dguittet (approved):
Really well done on this large piece of work! And thanks for splitting up the grid outage work into separate PRs.

shared/lib_resilience.cpp
```diff
@@ -51,48 +51,22 @@ dispatch_resilience::~dispatch_resilience() {
...
-    double max_discharge_kwdc = _Battery->calculate_max_discharge_kw();
-    double max_discharge_kwac = max_discharge_kwdc * m_batteryPower->singlePointEfficiencyDCToDC;
-    double max_charge_kwdc = _Battery->calculate_max_charge_kw();
-
-    double met_load;
-
-    if (pv_kwac > crit_load_kwac){
-        double remaining_kwdc = -(pv_kwac - crit_load_kwac) * m_batteryPower->singlePointEfficiencyACToDC;
-        remaining_kwdc = fmax(remaining_kwdc, max_charge_kwdc);
-        dispatch_kw(remaining_kwdc);
```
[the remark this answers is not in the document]
brtietz: Yes, good catch!

test/shared_test/lib_battery_dispatch_automatic_btm_test.cpp
brtietz: Right, thanks. Removing this doesn't change the test result, and these tests still show what I want them to show (ability to resume normal dispatch after an outage), so I've just removed this else if block.

brtietz:
Merged in 9cf27a1bb29f9488eb1b13f95a3ce35315d3ea9b

PR #633, opened 2021-10-29
Wave Resource API calls

mjprilliman:
-Cpp widget for Advanced wave resource download widget
-Updates to wave model outputs (more changes to come in separate PR)


cpaulgilman (changes requested):
Looks good. A couple of comments:

1. Should the default download folder be the device folder or the weather folder, or do we need a separate folder for wave resource files? If weather file reader can distinguish between SAM CSV format for solar and wave resource file format, we could just use the default SAM Downloaded Weather Files folder as the default for both solar and wave resource files.

2. Please remove the NSRDB attribute list at line 459 of wavetoolkit.cpp.

3. The instructions in the download window mention typing a location name like "eureka, ca" but inputs are only available for latitude and longitude.

I was not able to download a file to test -- this looks like an API issue rather than a problem with the code (for testing, I set API keys in private.h and email address in webapis.conf):

image

cpaulgilman (approved):

ssc/cmod_mhk_wave.cpp
```diff
@@ -524,7 +524,7 @@ class cm_mhk_wave : public compute_module
                 throw exec_error("mhk_wave", "Wave height and Energy period arrays of equal length must be assigned");
 
             ssc_number_t* energy_hourly = allocate("hourly_energy", number_records);
-            ssc_number_t* energy_hourly_gen = allocate("gen", number_records);
+            ssc_number_t* energy_hourly_gen = allocate("gen", number_records * 3);
```
sjanzou, line 527: So, energy_hourly - energy_hourly_gen[i*3] ? Is energy_hourly_gen[number_records*3) used for plotting or other outputs?

ssc/cmod_mhk_wave.cpp
```diff
@@ -564,25 +564,55 @@ class cm_mhk_wave : public compute_module
...
+                if (sig_wave_height_index == 0 || energy_period_index == 0) {
+                    throw exec_error("mhk_wave", "The wave conditions at timestep" + to_string(i) + "were not able to be located in the power matrix. Please check your resource file");
+                }
+                
+
                 //n-hour energy based on wave power matrix value at height and period best matching the time series inputs * number devices * size multiplier
                 energy_hourly[i] = (ssc_number_t)(wave_power_matrix.at(size_t(sig_wave_height_index), size_t(energy_period_index))) * hour_step * (1 - total_loss / 100) * number_devices;
                 p_annual_energy_dist[size_t(sig_wave_height_index) * 22 + size_t(energy_period_index)] += energy_hourly[i]; //Add energy for given time step to height x period distribution matrix at specified grid point
-                energy_hourly_gen[i] = energy_hourly[i]; //Store in gen to use in heatmap output (probably don't need two variables)
+                energy_hourly_gen[i*3] = energy_hourly[i]; //Store in gen to use in heatmap output (probably don't need two variables)
```
sjanzou, line 612: I see the allocation above and the for loop values... is there a check that size_t(sig_wave_height_index) is within the row size of wave_matrix_power and  size_t(energy_period_index)) is within the column size?

sjanzou (approved):
Please check that the wave_matrix row and column are within bounds per comments. Thanks!

ssc/cmod_mhk_wave.cpp
mjprilliman: I need to make some changes here. To have the wave energy outputs be compatible with the Combine Cases functionality there either needs to be a 1-hour timestep output or unique data handling in the compare cases tool. Gen needs to be the kW output at each 3-hour timestamp to be consistent with the rest of the models and allow for the wave model to be used properly in tools such as reV.

PR #634, opened 2021-10-31
Delete cf_batt_replacement_cost duplicate

dguittet:
`cf_batt_replacement_cost` is a duplicate of `cf_battery_replacement_cost` in common.cpp


mjprilliman (approved):
Thanks for the catch! I need to do more cleanup on the LCOS code but it's good to get rid of the duplication here.

PR #646, opened 2021-11-08
Fixed utilityrate to check for hourly buy rates 

mjprilliman:
-ur_calc_timestep updates tier_charge within hourly buy rates loop to avoid falling back on TOU rates
-Closes  https://github.com/NREL/SAM/issues/736


brtietz (approved):
Nice catch!

PR #648, opened 2021-11-09
Sloped Terrain inputs for backtracking algorithm, shadeFraction

mjprilliman:
-new cross_axis_slope function to calculate cross axis slope from terrain tilt and azimuth
-PVsamv1 allows for terrain effects on shadeFraction for non-zero terrain inputs
-Additional inputs to shadeFraction function, given 0 values for flat terrain for all non-pvsamv1 instances
-Added test for non-zero terrain inputs
-Reference: https://www.nrel.gov/docs/fy20osti/76626.pdf


ssc/cmod_irradproc.cpp
```diff
@@ -59,6 +59,8 @@ static var_info _cm_vtab_irradproc[] = {
         { SSC_INPUT,        SSC_NUMBER,      "rotlim",                     "Rotational limit on tracker",    "deg",    "",                      "Irradiance Processor",      "?=45",                    "MIN=0,MAX=90",                             "" },
         { SSC_INPUT,        SSC_NUMBER,      "backtrack",                  "Enable backtracking",            "0/1",    "",                      "Irradiance Processor",      "?=0",                    "BOOLEAN",                                   "" },
         { SSC_INPUT,        SSC_NUMBER,      "gcr",                        "Ground coverage ratio",          "0..1",   "",                      "Irradiance Processor",      "backtrack=1",              "MIN=0,MAX=1",                             "" },
+        { SSC_INPUT,        SSC_NUMBER,      "cross_axis_slope",                        "Cross axis slope",          "deg",   "",                      "Irradiance Processor",      "backtrack=1",              "MIN=0,MAX=1",                             "" },
```
janinefreeman, line 62: I think we don't want this input to be required, but rather to be set by default to zero (?=0).

janinefreeman (approved):
Other than the one comment about changing the input to irradproc, looks good to me!

shared/lib_irradproc.cpp
```diff
@@ -1548,7 +1548,8 @@ void incidence(int mode, double tilt, double sazm, double rlim, double zen,
             }
             else if (en_backtrack) {
                 // TODO: add cross-axis slope angle parameter
-                double backtracking_rotation = backtrack(truetracking_rotation * 180 / M_PI, gcr);
+                double axis_slope = cross_axis_slope(slope_tilt, azm, slope_azm);
```
kanderso-nrel, line 1551: `axis_slope` has potential to be misinterpreted as axis tilt if not reading carefully.  Maybe rename the function to `calc_cross_axis_slope` and then the variable can be named `cross_axis_slope`?  Just an idea :)

Also, I think the comment in the line above can be removed now.  Similar comments appear in this file in a couple other places.

shared/lib_irradproc.h
```diff
@@ -861,7 +863,7 @@ void ModifiedDISC(const double kt[3], const double kt1[3], const double g[3], co
 * \param[in] rotation tracking axis rotation angle in degrees
 * \return fraction shaded (0-1) if system is shaded (0 for unshaded)
 */
-double shadeFraction1x(double solar_azimuth, double solar_zenith, double axis_tilt, double axis_azimuth, double gcr, double rotation);
+double shadeFraction1x(double solar_azimuth, double solar_zenith, double axis_tilt, double axis_azimuth, double gcr, double rotation, double slope_tilt, double slope_azimuth);
```
kanderso-nrel, line 866: Do these new parameters need to be listed in the docs comment above?

shared/lib_irradproc.h
```diff
@@ -881,8 +883,18 @@ double truetrack(double solar_azimuth, double solar_zenith, double axis_tilt, do
 * \param[in] gcr ground coverage ratio (0-1) of array
 * \return updated rotation angle in degrees after backtracking
 */
-double backtrack(double truetracking_rotation, double gcr);
+double backtrack(double truetracking_rotation, double gcr, double axis_slope);
```
kanderso-nrel, line 886: Same comment re: listing new param in the doc comment

ssc/cmod_pvsamv1.cpp
```diff
@@ -129,6 +131,8 @@ static var_info _cm_vtab_pvsamv1[] = {
...
         {SSC_INPUT, SSC_NUMBER,   "subarray2_gcr",                        "Sub-array 2 Ground coverage ratio",                   "0..1",   "",                                                                                                                                                                                      "System Design",                                         "?=0.3",                              "MIN=0.01,MAX=0.99",   "" },
+        {SSC_INPUT, SSC_NUMBER,   "subarray2_slope_tilt",           "Sub-array 1 Cross axis slope",                        "deg",   "",                                                                                                                                                                                      "System Design",                                         "?=0.0",                                                "",   "" },
```
kanderso-nrel, line 134: This is for subarray2 but says `"Sub-array 1 Cross axis slope"`.  Same for subarrays 3 and 4

ssc/cmod_pvsamv1.cpp
```diff
@@ -69,6 +69,8 @@ static var_info _cm_vtab_pvsamv1[] = {
...
+        {SSC_INPUT, SSC_NUMBER,   "subarray1_slope_tilt",           "Sub-array 1 Cross axis slope",                        "deg",   "",                                                                                                                                                                                      "System Design",                                         "?=0.0",                                                "",   "" },
+        {SSC_INPUT, SSC_NUMBER,   "subarray1_slope_azm",           "Sub-array 1 Cross axis slope",                        "deg",   "",                                                                                                                                                                                      "System Design",                                         "?=0.0",                                                "",   "" },
```
kanderso-nrel, line 73: I think there are two issues in the description here -- first, `Sub-array 1 Cross axis slope` is used for both `slope_tilt` and `slope_azm` inputs.  Second, I don't think either of these is actually a cross-axis slope :)

And same for subarrays 2-4

kanderso-nrel (commented):
Hi folks, thanks again for finishing this up!  Here are some comments, maybe too little too late at this point.

@mikofski fyi

mjprilliman:
Thanks Kevin! I'll add your edits in a separate commit as that will help eliminate a lot of confusion for anyone looking at this code in ssc. @janinefreeman these are all just text changes so I think I'll just commit them to develop

PR #649, opened 2021-11-10
Fix broken CentOS7 build

sjanzou:


tyneises (changes requested):
I committed a slightly different version to your branch. If you don't like that, I'm fine using your fix.

sjanzou:
11/11/2021 2am build and run on CentOS7 
image

tyneises (approved):

brtietz (approved):

PR #652, opened 2021-11-10
Time series energy and demand charges for LCOE calcs

brtietz:
Doing an inter-branch pull request to test the repurposing of a few outputs. Changes:

- Instead of reporting the entire energy charge in the last hour of the month for net metering in year1_hourly_ec_with_system, report timestep length changes in the bill
  - Positive means energy charges
  - Negative means net metering credits are being earned
  - True up credits (end of month or annual) are not included
  - Summing over the month matches the monthly value of "energy charges before credits"
  - Net billing already had this feature
-  Instead of reporting the entire demand charge in last hour of the month, report timestep length changes in the demand charge
  -  Each timestep reports an increase vs previous timesteps
  - If billing demand ratchets are used, many months will have a charge representing the ratchet in the first hour
  - This requires computing the demand charge many additional times. This had a roughly 10% increase in runtime for cmod_utility_rate5 (less than 0.1 seconds)

Let me know if you want to account for monthly credits on this branch, or lcoe_lppa_energy_sales


brtietz:
I ran the ssc tests on this branch last night and got the following results:

```
[==========] 536 tests from 91 test suites ran. (1219224 ms total)
[  PASSED  ] 533 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] CMPvsamv1BatteryIntegration_cmod_pvsamv1.PPA_ACBatteryModelIntegration
[  FAILED  ] CMPvsamv1BatteryIntegration_cmod_pvsamv1.ResidentialDCBatteryModelPriceSignalDispatch
[  FAILED  ] CMBattwatts_cmod_battwatts.NoPV
```

@mjprilliman any idea if those are happening on ` lcoe_lppa_energy_sales`? I don't think the changes I made here should have affected those tests, but it's hard to know for sure without knowing the results from that branch.

mjprilliman:
> I ran the ssc tests on this branch last night and got the following results:
> 
> ```
> [==========] 536 tests from 91 test suites ran. (1219224 ms total)
> [  PASSED  ] 533 tests.
> [  FAILED  ] 3 tests, listed below:
> [  FAILED  ] CMPvsamv1BatteryIntegration_cmod_pvsamv1.PPA_ACBatteryModelIntegration
> [  FAILED  ] CMPvsamv1BatteryIntegration_cmod_pvsamv1.ResidentialDCBatteryModelPriceSignalDispatch
> [  FAILED  ] CMBattwatts_cmod_battwatts.NoPV
> ```
> 
> @mjprilliman any idea if those are happening on ` lcoe_lppa_energy_sales`? I don't think the changes I made here should have affected those tests, but it's hard to know for sure without knowing the results from that branch.

image
They are happening on lcoe_lppa_energy_sales, I will look into it now

https://app.travis-ci.com/github/NREL/ssc/jobs/547630959

mjprilliman (commented):
I'm probably not the best person to review the specifics of the financials but the overall structure seems to work for what I need in the parasitic cost calculations in the financial models. 

Quick clarification question: does the financial model need to know which billing option was used in the utility bill calculations or can the hourly time series be used in the same way regardless?

I can account for the monthly credits on this branch

brtietz:
```
Quick clarification question: does the financial model need to know which billing option was used in the utility bill calculations or can the hourly time series be used in the same way regardless?
```
The hourly time series should be useable the same way regardless. My one qualm is the effect on the cost calculations of some of the negative numbers provided by net metering, but these only occur during grid export, so shouldn't affect calculations for the cost of parasitics or grid charging.

Accounting for the monthly credits will depend on the metering option, but since only one credit category has a value it should be possible to just add them all up and not worry about what option was selected. I can add a function that does this in utility_rate_5 if that seems like a better way to do it.

cpaulgilman (approved):
This approach makes sense. You can see how credits accumulate for energy charges and when the demand charge happens. I'll revise descriptions in Help accordingly.

cpaulgilman:
@brtietz I'm finally getting around to documenting these changes for Help. 

The default PVWatts/Commercial case has TOU demand rates with 2 TOU periods each and 2 tiers, so I was expecting to see 4 values per month in the hourly "Demand charge with system" output. But instead, I'm seeing a handful of "Demand charge with system" in $ values per month that do not always correspond to the "Electricity peak from grid" in kW, and I'm seeing negative values in the first day of the month even without billing demand ratchets. Can you help explain what these values represent?

image

brtietz:
The intention with these is to record the incremental demand charge, so each time the demand charge increases it would produce a line proportional to the increase in the "demand charge with system" output. This is to assist with the contribution of grid charging to the LCOE and LCOS for the battery.

The negative lines are definitely a bug. I'm also not sure why there would be a blue line without an orange line, so it looks like there might be two bugs here.

PR #653, opened 2021-11-11
Batt btm to grid and manual clip

brtietz:
Pairs with SAM PR https://github.com/NREL/SAM/pull/754

Update charging and discharging constraints to allow:

- Clip charging for BTM dispatch (manual and price signals) and FOM manual dispatch
- Discharge to grid for BTM dispatch (only affects manual and custom right now)

PVWatts-Battery defaults grid discharging to true.

Peak shaving, grid power targets, and price signals dispatch will need updates to take advantage of grid discharge capabilities. This can be done in a future PR for a patch.

Solves SAM issues https://github.com/NREL/SAM/issues/338 and https://github.com/NREL/SAM/issues/407


brtietz:
Failure matches version of develop used to create this branch. Develop was fixed here: https://github.com/NREL/SAM/runs/4182767103. I can merge in develop and re-run if desired, but refraining for now to save Travis credits.

test/ssc_test/cmod_battwatts_test.cpp
```diff
@@ -100,7 +100,7 @@ TEST_F(CMBattwatts_cmod_battwatts, ResidentialDefaults) {
     ssc_number_t peakKwDischarge = *std::max_element(batt_power_data.begin(), batt_power_data.end());
     ssc_number_t peakKwCharge = *std::min_element(batt_power_data.begin(), batt_power_data.end());
 
-    EXPECT_NEAR(peakKwDischarge, 1.98, 0.1);
+    EXPECT_NEAR(peakKwDischarge, 2.16, 0.1);
```
mjprilliman, line 103: I'm sure I missed it but is this change due to the DischargetoGrid defaulting to true?

mjprilliman (approved):
I didn't notice any issues but I'm not as familiar with this code

test/ssc_test/cmod_battwatts_test.cpp
brtietz: Yes - since the discharge is no longer limited to load, the peak discharge can increase.

PR #654, opened 2021-11-11
Update pvsamv1 and pvwattsv8 to linear DC degradation rather than compounding

### LINKED ISSUE #503: Compounding vs linear module degradation
We may need to revisit the assumptions around compounding vs linear module degradation in the PV models. Tim Silverman and Dirk Jordan have suggested that even though compounding is the actual physical mechanism, linear may be a better representation of the findings in the data.

janinefreeman:
Per NREL experts, updating the detailed pv model and the newest pvwatts version to apply DC degradation linearly instead of compounding. All SSC tests are passing locally. Pairs with NREL/SAM#756 PR to update defaults results and the upgrade script. Fixes #503 .


brtietz (commented):
Two questions on this prior to approving:

(1) Do we need any code protecting the output from values > 100 / analysis period? Previously the pow function would protect us from negative numbers, but we lose that with the linear formulation.

(2) Does this significantly (>0.1%) change the output in test_script_ow.lk? Based on the lack of change in the SSC tests we should be ok, but if one of us has time it might be worth running prior to merging.

brtietz:
Question (2) makes me look like I can't read. Nevermind on that one, I've reviewed SAM 756 as well now.

janinefreeman:
1. Good catch! Looks like in the current release, entering a degradation value > 100% in the UI results in nonsensical degradation numbers that do appear to get used to produce negative energy from a system. I'll add some type of catch for that here next.
2. It changes the LCOE numbers for the PV configs, I re-ran the test script in SAM on the branch but will need to re-run again before merging.

janinefreeman:
@brtietz Updated the logic to address question 1, please review additions to lib_pv_io_manager when you get a chance. Thanks!

brtietz (approved):
The new code and warning flag work great, thanks!

PR #655, opened 2021-11-12
Cent os7 regex replace removal

sjanzou:
Fix on all platforms - replaces approved pull request 649 which did not work on Travis


sjanzou:
ssc tests:

CentOS 7 results
image

Travis-CI results
image

Windows results (SEH exception for non annual data)
image

PR #659, opened 2021-11-15
Standalone battery models in SAM

mjprilliman:
-gen is always zero for standalone battery model cases
-en_standalone_batt variable controls compute module behavior for standalone batteries
-Otherwise battery model runs as normal


ssc/cmod_battery.cpp
```diff
@@ -2107,10 +2108,19 @@ class cm_battery : public compute_module
...
     void exec() override
     {
-        if (as_boolean("en_batt"))
+        if (as_boolean("en_batt") || as_boolean("en_standalone_batt"))
         {
+            std::vector<ssc_number_t> power_input_lifetime;
             // System generation output, which is lifetime (if system_lifetime_output == true);
-            std::vector<ssc_number_t> power_input_lifetime = as_vector_ssc_number_t("gen");
+            if (as_boolean("en_standalone_batt")) {
+                power_input_lifetime.resize(8760.0 * as_integer("analysis_period"), 0.0);
```
brtietz, line 2116: I expect users will complain about only being able to run this on an hourly basis. For behind the meter, we can match the frequency of the load profile. How should we handle setting the time step for front of meter?

mjprilliman: Maybe make an optional timestep input that keys off of the size of the curtailment array?

brtietz: That makes sense to me for this release, at least. The battery needs the curtailment array anyway for powerflow calculations, so might as well use it for timestep.

cpaulgilman (approved):

brtietz (approved):

PR #662, opened 2021-11-17
Update pvsamv1 to assume anti-reflective glass

### LINKED ISSUE #323: calculateIrradianceThroughCoverDeSoto should include anti-reflective coatings
The calculateIrradianceThroughCoverDeSoto function (lib_pv_incidence_modifier.cpp) does not take into account anti-reflective (AR) coatings, and should be modified to do so. The code submitted in #315 could be added to this function to correct that.

janinefreeman:
This PR updates pvsamv1 to assume anti-reflective glass by default, and updates the associated tests. Annual energy increases due  to this change. This is consistent with the assumptions we make in pvwattsv7 and pvwattsv8. The SAM test script needs to be re-run for these changes, but at the moment I'm getting geothermal equations errors. Closes #323


sjanzou (approved):
Many tests did change as expected. Looks good and tests completed.

PR #665, opened 2021-11-18
Corrected sign of parastic costs in LCOE BTM calculations

mjprilliman:
-Negative LCOE due to incorrect sign of parasitic cost cash flow discovered in PVWatts Battery
-Sign flipped in ssc LCOE calculations to correct


cpaulgilman (approved):
I couldn't replicate the negative LCOE issue in the previous Develop version.

mjprilliman:
[PVWatts_NegativeLCOE_example.zip](https://github.com/NREL/ssc/files/7565996/PVWatts_NegativeLCOE_example.zip)

Sorry I should've shared a file before. This is from the current state of develop.

PR #672, opened 2021-11-27
address SAM issue 727 pvwatts adjustment factors to dc and ac output

sjanzou:


cpaulgilman:
https://github.com/NREL/sam/issues/727

Issue #675, opened 2021-11-29
CMPvsamv1BatteryIntegration_cmod_pvsamv1.ResidentialDCBatteryModelPriceSignalDispatch faling on Travis and runs fine on Windows


sjanzou:
The one failing test runs fine on Windows
<img width="606" alt="image" src="https://user-images.githubusercontent.com/6498311/143821069-4b899b0c-cdad-4253-8f24-cc65967f9fec.png">

Retesting on Travis running single test     ::testing::GTEST_FLAG(filter) = "CMPvsamv1BatteryIntegration_cmod_pvsamv1.ResidentialDCBatteryModelPriceSignalDispatch";
<img width="1094" alt="image" src="https://user-images.githubusercontent.com/6498311/143823668-7fdd039d-edab-4b67-845d-f8d4a3252b04.png">

batt_DOD_cycle_average output difference - separate issue
27.29 on Windows 
27.52 on Travis

_Originally posted by @sjanzou in https://github.com/NREL/ssc/issues/673#issuecomment-981348314_

dguittet:
It passes on Mac also with value 27.290548

brtietz:
This appears to be a recurrence of issue https://github.com/NREL/ssc/issues/614 - the linux and windows implementations end up with slightly different dispatch patterns. 

With the defaults currently in the GUI, windows schedules dispatch for step 966 but not 967, whereas Linux does the reverse. This causes the SOC to diverge, with greater divergence in the dispatch plan from there. It amounts to about 4 out of 926 kWh in year 1.

Interestingly, the Linux test is displaying the old value: https://github.com/NREL/ssc/commit/442ea5b6a787f75335aece7199efb6f442f943d3 Unfortunately due to the Travis outages I can't see if that commit caused the test failure or if the error was introduced a few days earlier and that's when it was exposed.

Would it be acceptable to increase the tolerance on the test for now and clean this up after the release? I created a branch addtl_lifetime_testing_in_integration_tests that shows that this issue is restricted to price signals dispatch - we could merge that in for additional test coverage of cycling differences between platforms.

brtietz:
I checked out 314cb53e0 (from Nov 17th, before the anti-reflective glass change) and the test passes there. This means that the dispatch changes in https://github.com/NREL/ssc/pull/662 exposed the Linux failure and my fix to https://github.com/NREL/ssc/issues/614 is insufficient.

[...]

Issue #614, opened 2021-09-09
Price Signals Dispatch runs slightly differently on Windows and Linux


brtietz:
This is most obvious in the test CMPvsamv1BatteryIntegration_cmod_pvsamv1.ResidentialDCBatteryModelPriceSignalDispatch, where Windows has an average cycle depth 0.4 percentage points higher than Linux, hence the high tolerance.

A suspicious candidate is the sorting algorithms, as these can be different in different implementations of the standard library. Clean up and reduce the tolerances for the tests.

brtietz:
Closed in https://github.com/NREL/ssc/commit/9cf27a1bb29f9488eb1b13f95a3ce35315d3ea9b

brtietz:
This issue recurred in https://github.com/NREL/ssc/issues/675. See that issue for some additional debugging details. The timestep numbers may drift as defaults or PVSamV1 changes. The general approach is to look for when battery SOC diverges, and then find the dispatch step that causes the divergence.

brtietz:
Plausible recurrence in https://github.com/NREL/ssc/actions/runs/12321954327/job/34394448965

brtietz:
It looks like the unstable sort behavior is due to the tiebreaks in the lowest marginal cost function. Revisiting these tiebreakers and including a tiebreaker based on step order might solve the problem.

brtietz:
See https://github.com/NREL/ssc/commit/2b01ee234169569be5d665d79e5921252a75c825 for details.

PR #677, opened 2021-11-29
Addtl lifetime testing in integration tests

brtietz:
Add additional tests of battery lifetime to pin down issue https://github.com/NREL/ssc/issues/675 on price signals dispatch. Increase the tolerance for the price signals test until we can better nail down cross platform differences.

Will re-open https://github.com/NREL/ssc/issues/614 to solve after the release.


sjanzou (approved):
Looks good

PR #678, opened 2021-11-30
Correctly handle AR glass in calculateIrradianceThroughCoverDeSoto

### LINKED ISSUE #669: RC IAM Gains?
The current default PV-Battery Partnership Flip defaults in the release candidate result in an IAM "gain" rather than loss. This seems suspicious to me, probably related to the change to assume all modules have an AR coating. I'll investigate.

### LINKED ISSUE #674: Detailed PV Model "POA from reference cell" produces incorrect reflection losses
Possibly present in the current release as well, the option to use "POA from reference cell" produces 100% reflection losses in the loss diagram because the incidence angle modifier is never set in the `cec6par_module_t::operator()` function of `lib_cec6par.cpp`. However, the annual energy does seem to be calculated correctly somehow.

janinefreeman:
This pull request updates the `calculateIrradianceThroughCoverDeSoto` function to properly handle anti-reflective coated glass. Previously, the function didn't include the AR coating in the calculation of transmittance at normal angles, but it did in calculation of transmittance at angle of incidence. This caused calculated incidence angle modifiers for beam irradiance to be >1. The only model impacted by this change previously was pvwattsv7 because it used that function with an AR glass assumption, but it had a check to make sure that transmitted poa wasn't greater than incident poa, so effects are minimal. However, with the as-yet unreleased switch to assuming AR glass throughout pvwattsv8 and pvsamv1, this bug became very apparent.

This fix adds the effect of AR glass into the transmittance at normal incidence, and also adds checks to make sure that no component IAM is calculated to be > 1. Many tests were updated accordingly. In the process of finding the bug, I also slightly reorganized and updated comments in incidence angle modifier code, removed the unused zenith angle input from the calculateIrradianceThroughCoverDeSoto function, and fixed a bug where using POA irradiance directly never reported reflection losses.

Fixes #669 and fixes #674

I will make a PR for SAM to update test results after this is approved.


mjprilliman (approved):
Re-running the ssc tests now but barring any unexpected failures it looks good to me. There's no UI element to this is there? The assumption is built into ssc automatically?

janinefreeman:
Correct, no UI changes, just assumptions in SSC. The CEC database doesn't give us information on whether or not a module is AR coated.

PR #679, opened 2021-11-30
Added divide by zero check for BTM battery grid charging cost

mjprilliman:
-Avoid error in LCOS BTM calculations by checking for 0 monthly grid charge before performing parasitic cost calculations
-Fixes https://github.com/NREL/SAM/issues/801


brtietz (approved):
Looks great! I agree that it makes sense that the value of true up payments should be zero if there is no net energy drawn from the grid.

PR #684, opened 2021-12-08
Github actions on Linux

brtietz:
Working with GitHub actions on Linux. WIP will be removed once this is ready to merge

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,11 @@ jobs:
       run: |
         cd ${SSCDIR}/build
         make -j4
+        ls
+        cd ${SSCDIR}/build/test
+        ls
 
     - name: Test
-      working-directory: ${{github.workspace}}/build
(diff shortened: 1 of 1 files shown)


brtietz:
The Geothermal issue failing the tests is real (confirmed in a VM) so I'm inclined to merge this to patch and develop so we can start using the results right away.

Additional GitHub Actions PRs will include cross platform building and artifacts.

sjanzou (approved):
Will the macOS GitHub Actions follow? Same issue with SAM pull request 845 about the Ubuntu 18.04 version of gcc 7.5.0 and libc 2.27 and our build system CentOS7 gcc 4.8.5 and libc 2.17.

brtietz:
Yes, I'll try to get the Mac build done by the end of next week.

I would be curious to poll our Linux users about what they're running on. I understand the need to support CentOS for HPC, but I suspect a decent number of users would appreciate us supporting more recent Linux releases, especially with Ubuntu 16.04's LTS ending.

PR #690, opened 2021-12-10
Power Block Hourly option causes hour of year indexing failure in Geothermal

mjprilliman:
-Hour indexing corrected to go from 0-23 for the day rather than hourly years when the hourly option is selected in the UI


mjprilliman (commented):
Approved

PR #691, opened 2021-12-10
GitHub Actions working on Mac build

brtietz:
Add Mac runner to GitHub actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
 
       
   build-on-mac:
-    runs-on: macos-latest
+    runs-on: macos-11
 
     steps:
     
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,14 +7,9 @@ on:
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
-  GTEST: 
-  SSCDIR: $GITHUB_WORKSPACE/ssc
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -7,18 +7,72 @@ on:
...
+    - name: Test
+      # Turn off fast fail for when the landbosse tests write to cerr
+      shell: bash
+      run: |
+        set -e
+        ${SSCDIR}/build/test/Test
+
+      
+  build-on-mac:
+    runs-on: macos-latest
```
sjanzou, line 65: Do you know the mac runner hardware and macOS version? what is macos-latest?

sjanzou (approved):

.github/workflows/ci.yml
brtietz: `macOS Catalina 10.15`, as per https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

brtietz: Apparently they're in the process of migrating: https://github.com/actions/virtual-environments/issues/4060
It'd probably safer to fix a version here, let me know which is preferred.

sjanzou: For reference, the 2021.12.02 release was built using 11.6.1 on an Intel based Macbook Pro.
The macOS_wx3.1.5 was built using 12.0.1 on an Apple Silicon M1 based Macbook Air for Intel and Apple Silicon (using Rosetta) based Macs.

PR #695, opened 2021-12-14
fix upper level m_dot effect calc

tyneises:
The User Defined Power Cycle model was using an incorrect equation to calculate the interaction effect for the high level of HTF mass flow rate. This bug only affects cases where the power option is set to the UDPC model, and as such does not influence default CSP cases that use the Rankine cycle model.

When applied to the default molten salt power tower case where the cycle model is changed to the default UDPC inputs, the annual energy increase 0.1%. The effect may be more significant if 1) your UDPC data includes more variation relative to the default data at m_dot_HTF > 1 or 2) your case increases the maximum HTF over-design above the default 1.05 value.


tyneises:
fyi @cpaulgilman

PR #699, opened 2021-12-28
Net elec calculations

tyneises:


tyneises:
Moves parasitic reporting to component classes from CSP solver.

PR #701, opened 2021-12-29
macOS wxWidgets 3.1.5 universal binary

sjanzou:


CMakeLists.txt
```diff
@@ -91,7 +91,8 @@ function(set_default_compile_options target)
     else(MSVC)
         SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
         if (APPLE)
-            set(MAIN_CFLAGS "${MAIN_CFLAGS} -arch x86_64 -fno-common -DWX_PRECOMP -D__MACOSX__")
+#            set(MAIN_CFLAGS "${MAIN_CFLAGS} -arch x86_64 -fno-common -DWX_PRECOMP -D__MACOSX__")
+            set(MAIN_CFLAGS "${MAIN_CFLAGS} -fno-common -DWX_PRECOMP -D__MACOSX__")
```
dguittet, line 95: can you remove unused comment?

solarpilot/CMakeLists.txt
```diff
@@ -68,7 +68,8 @@ if(MSVC)
 	endforeach()
 else()
 	if (APPLE)
-		add_compile_options(-arch x86_64 -fno-common)
+#       add_compile_options( -arch x86_64 -arch arm64  -fno-common)
+        add_compile_options( -fno-common)
```
dguittet, line 72: same here

ssc/CMakeLists.txt
```diff
@@ -197,10 +197,10 @@ if(UNIX AND NOT SAMAPI_EXPORT)
...
 		target_link_libraries(dltest -ldl -Wl,--no-as-needed)
 	endif()
-	add_custom_command(
-		TARGET ssc
-		POST_BUILD
-		COMMAND dltest $<TARGET_FILE:ssc>)
+#	add_custom_command(
+#		TARGET ssc
+#		POST_BUILD
+#		COMMAND dltest $<TARGET_FILE:ssc>)
```
dguittet, line 203: why are we no longer running this?

dguittet (changes requested):

CMakeLists.txt
sjanzou: Will do... thanks!

ssc/CMakeLists.txt
sjanzou: There was an issue with the build path, adding back now...

dguittet (approved):

ssc/CMakeLists.txt
```diff
@@ -200,6 +200,7 @@ if(UNIX AND NOT SAMAPI_EXPORT)
 	add_custom_command(
 		TARGET ssc
 		POST_BUILD
+		COMMAND codesign --force --deep --sign - $<TARGET_FILE:ssc>
```
brtietz, line 203: This is causing me trouble on the Linux build - getting "codesign - not found" Is there a new dependency that should be added to the wiki?

brtietz (changes requested):
Building with SAMAPI_EXPORT on works on linux, but off causes an error with the codesign command. (on fails the sam build due to the ssc library rename, which is expected)

sjanzou:
That is an apple only thing.
Removing for Linux...thanks!

ssc/CMakeLists.txt
sjanzou: Fixed in 8c9657eb7dab4d01be4c8faab7bd46ab96bfc4ce

brtietz (approved):
Testing looked good on Linux this morning. Nice work!

janinefreeman (approved):
Works for me on Windows with minimal testing

PR #707, opened 2022-01-04
Remove old csp code

tyneises:
This pull request removes obsolete code from SSC and TCS, especially obsolete or redundant/parallel code modeling the heliostat field and molten salt power receivers. This request includes:
- removing code from mspt receiver classes that was specific to integration with the ISCC technology model
- removing the ISCC compute modules from the SSC project and repo, and ISCC powerblock and parasitic component classes from the TCS project and repo.
- removing the TCS shell for the molten salt power tower (type 222) from the TCS project and repo
- removing the direct steam power tower compute modules from the SSC project and repo
- removing the TCS shell for the heliostat field and one obsolete version of the csp solver heliostat field class from the TCS project and repo
- removing the out-of-date sco2 power cycle csp solver class with guidance to use pre-process routines and user-defined power cycle model


Matthew-Boyd (approved):
It looks like you removed some "_ss" suffixed variables, but that you are keeping both the transient and steady-state receiver models. Let me know if that's not the case (as I assume we're still wanting to keep both).

PR #723, opened 2022-01-19
Fix SAM 857 and 875

brtietz:
Pairs with SAM pull request https://github.com/NREL/SAM/pull/889

- Allows billing demand computations when there are only kWh/kW energy charges (no demand charges)
- Allows for kWh/kW energy charges to only show up in some months

Both of the above can be tested with the file attached to https://github.com/NREL/SAM/issues/857


sjanzou (approved):
Works with psco_tiered_rates.sam from https://github.com/NREL/SAM/issues/857
image

cpaulgilman (approved):

PR #725, opened 2022-01-19
clean up max power eff function for large numbers of inverters

### LINKED ISSUE #722: DC Connected battery and gid outages
Run parametrics in attached SAM project file - one case dc connected and one case ac connected.
No annual output shown for dc case

base case has undersized battery and shows same system output in heat map on summary page but no annual output.

I can’t tell what’s meeting the critical load at night. The inverter efficiency appears to drop to zero at night, and with a two hour battery there shouldn’t be enough power to meet the critical load anyway.
Changing the parameters such as inverter efficiency cutoff and re-sizing the system doesn’t fix the problem.
 
Based on that, the issue is either in the inverter model or in the post-battery loss accounting for DC connected systems. 

[MiniGrid.zip](https://github.com/NREL/ssc/files/7892353/MiniGrid.zip)

image

image

brtietz:
Fix the max efficiency function by multiplying by the number of inverters. Update tests for improved efficiency estimates accordingly.

Fixes #722


sjanzou (approved):
Tested with project file posted in issue #722 and works without issue.

PR #726, opened 2022-01-20
Patch to develop

brtietz:
Includes https://github.com/NREL/ssc/pull/723 and https://github.com/NREL/ssc/pull/725


brtietz:
Have to close due to merge conflict in utility rate 5

PR #728, opened 2022-01-20
Fix ssc 856 xfmr losses

brtietz:
Fixes https://github.com/NREL/SAM/issues/856

The DC connected batteries need an estimate of loss percent for grid outages, but since some of the transformer load losses use *= the calculation was being applied twice. Reset the variables after the DC loss estimation to fix the issue.

Test by specifying a transformer load loss in a default case. An AC connected battery and DC connected battery should have similar transformer loss %s after running.


cpaulgilman (approved):
Looks good. Thanks!

Issue #694, opened 2021-12-14
common.cpp provides price information to FOM battery dispatch in $, not $/kWh


brtietz:
I did a deep dive into the battery’s merchant plant code this morning and yes, the battery does not dispatch because the cleared capacity is zero:

![image](https://user-images.githubusercontent.com/5530592/146055580-15f59d67-c671-48d1-a83c-efbea292be1d.png)

The line in the for loop there is estimating the revenue as (cleared capacity) * (revenue / MWh) – so since all of the cleared capacity is zero all of the revenue is zero. This strikes me as a units error even if the value was non-zero, since this is being put into a $/kWh variable once this function returns to cmod_battery:

![image](https://user-images.githubusercontent.com/5530592/146055601-bcd602a3-9d7d-4dac-978f-bdb4d5d5dd46.png)

One possible solution would be to correct the units in common.cpp by just looking at mp_energy_market_revenue_mat.at(j, 1) and then deal with cleared capacity once we handle https://github.com/NREL/SAM/issues/181 I’d be happy to file the pricing fix as a GitHub issue.

brtietz:
Additional problem, how to handle 100% cleared capacity from this screen:

![image](https://user-images.githubusercontent.com/5530592/146081327-00c0411d-1f50-4dbb-8954-1eb173e8f292.png)

brtietz:
Closed with 729

PR #729, opened 2022-01-21
Fix ssc 694

brtietz:
cmod_merchant_plant_eqns now delivers units of $/kWh to the battery dispatch code. This allows the automated battery controller battery to dispatch based on the price regardless of the cleared capacity. Introducing cleared capacity into the battery dispatch is scheduled for (the often delayed) https://github.com/NREL/sam/issues/181

Additionally, clean up the merchant plant error checking for cases where gen > cleared capacity.

Fixes https://github.com/NREL/ssc/issues/694


sjanzou (approved):
Looks good,

PR #734, opened 2022-01-25
Sam 877

dguittet:
The behavior of this equation has changed, making the if ( omegaT > omega_m ) return early whereas in the past, it added the error but continued onward to make the power curve. I'm honestly not sure we need to keep this early-exit error if it's failing for reasonable inputs. I think instead of making it an error, we can make it just a warning.

closes https://github.com/NREL/SAM/issues/877, https://github.com/NREL/pysam/issues/107


dguittet:
Accidental duplicate

PR #736, opened 2022-01-26
AC capacity factor

janinefreeman:
Added AC capacity factor to pvwattsv8 and updated associated labels in pvsamv1 and pvwattsv8


mjprilliman (approved):
Looks good. Not sure what there is a Mac error on a PVWattsv7 test. Maybe we want to change the summary table to clearly specify that it is DC capacity factor?

image


image

janinefreeman:
Thanks @mjprilliman ! Yes, I am planning to update the Summary table labels as well.

PR #738, opened 2022-01-27
Fix SSC 737 - improve dc outage charging

brtietz:
- Adjust battery charging for updated inverter efficiency and estimate unmet critical load in the dispatch
- Remove redundant dc-dc efficiency adjustments

Results in expected outage parametrics behavior: decreasing critical load unmet w/ increasing battery size:

image


sjanzou (approved):
Now dc results
image

consistent with ac results
image

both consistent with unmet load decreasing with battery size and duration

PR #739, opened 2022-01-28
Change "resiliency" variable labels for clariity

cpaulgilman:
The term "resiliency" means different things in different contexts, so avoid using in labels. In this case use "Battery hours of autonomy during grid outage" instead.

Accompanies https://github.com/NREL/SAM/pull/902

Replacing "resiliency" in variable labels with "hours of battery autonomy during grid outage".

I *did not* change the "resiliency" group name to avoid problems with PySAM.

Alternate suggestions for variable labels welcome!


ssc/common.cpp
```diff
@@ -797,15 +797,15 @@ ssc_number_t forecast_price_signal::operator()(size_t time)
...
-{ SSC_OUTPUT, SSC_ARRAY  , "survival_function"                    , "Survival function of autonomous hours for resilience"                  , ""                                       , "Prob surviving greater than x hours; hrs from min to max" , "Resilience"          , ""               , "MIN=0,MAX=1"           , ""},
-{ SSC_OUTPUT, SSC_NUMBER , "avg_critical_load"                    , "Average critical load met for resilience"                              , "kWh"                                    , ""                                                         , "Resilience"          , ""               , "MIN=0"                 , ""},
+{ SSC_OUTPUT, SSC_ARRAY  , "resilience_hrs"     , "Battery hours of autonomy during grid outage at each timestep"            , "hr"  , ""                                                         , "Resilience" , ""  , ""             , ""},
+{ SSC_OUTPUT, SSC_NUMBER , "resilience_hrs_min" , "Battery hours of autonomy during grid outage minimum"                     , "hr"  , ""                                                         , "Resilience" , ""  , "MIN=0"        , ""},
```
brtietz, line 801: What is our convention for minimum and maximum in variable names? The others I saw when searching are related to current and degradation:

image

image

I like this grouping since they'll be next to each other, but it does sort of run together when I read it. Would "Battery hours of autonomy during grid outage - minimum" work? Also open to a colon instead of a dash.

brtietz (changes requested):
Overall I like the naming and the grouping - just one nitpick regarding min and max in the variable names (which might trigger some additional renaming)

sjanzou (approved):
Looks good... I agree we should, at some point, have a consistent max and min naming scheme

dguittet (approved):

ssc/common.cpp
cpaulgilman: Based on the conversation at our SAM meeting today, I changed "Battery Autonomy" to "Autonomy" because autonomy from the grid can be provided by both the power generator and battery. I'm leaving the labels as is with "minimum," "maximum," "average" at the end of the label. I'm not using a colon or dash to avoid potential problems with parsing of label strings in some contexts.

https://github.com/NREL/ssc/blob/12f38761a822ed3e8db3bdb595c1bba2e5104cdc/ssc/common.cpp#L800

As for where "minimum" and "maximum" should be in labels, I would recommend putting the descriptive part of the label first (as in "Battery charge current") and the qualifier at the end. Then you would see all of the "battery charge" variables listed together, which seems more useful than seeing all of the "battery maximum" variables listed together.

brtietz (approved):

PR #741, opened 2022-01-28
745 zero load during grid outage

brtietz:
- cmod_grid now sets "load" to zero during grid outage steps
- To report inputs in the data tables, full_load now has the load without the outage steps
- Fixes a bug with the shared inverter losses in the dc connected systems
- Increases the tolerance for when a system is considered offline to cover edge cases with degradation

This means if the outage is a full year, the utility bill without system and the utility bill with system will just consist of fixed costs. This is a big change, so I'm happy to wait to merge the PR until we've had a chance to talk through the downstream effects at a meeting.

Closes https://github.com/NREL/SAM/issues/745


sjanzou:
Hmmm... maybe further discussion is necessary. In using the attached project file, the dc connected electricity bill savings are:
image

and ac connected (same configuration) are: (varying seems okay based on differences in annual energy and battery sizing - however, I would not like to explain the differences!)
image

brtietz:
@sjanzou Ah, thanks for testing those additional cases. There were some situations where unmet losses were slipping through and pushing negative numbers into gen - which counted as small demand charges. I've cleaned that up in the latest push, and now see zero for all "electricity bill savings" locally.

sjanzou (approved):
Looks good - should probably confirm this is the desired behavior at the Puerto Rico meeting...
image

brtietz:
Grid outage added as an hourly output, as per discussion at the SAM team meeting.

@cpaulgilman Does the label still make sense given that it's an output?

image

Perhaps it should be "grid outage in this timestep"?

cpaulgilman:
"Grid outage in this time step" is good.

cpaulgilman (approved):

PR #750, opened 2022-02-07
Fix cf energy sales

brtietz:
Fixes issue with redundant variables: "PPA revenue gross" and "PPA revenue net" were in fact using the same calculations now that we transitioned electricity purchases to an O&M cost regardless of payment method. Remove PPA revenue gross, since it wasn't used in any downstream calculations.

Pairs with SAM PR: https://github.com/NREL/SAM/pull/919

### DIFF
--- a/ssc/cmod_communitysolar.cpp
+++ b/ssc/cmod_communitysolar.cpp
@@ -606,8 +606,6 @@ static var_info _cm_vtab_communitysolar[] = {
     /* PPA revenue not applicable to community solar, may need to be restored later
     { SSC_OUTPUT,       SSC_ARRAY,      "cf_ppa_price",                           "PPA price",                     "cents/kWh",      "",                      "Cash Flow Revenues",             "*",                      "LENGTH_EQUAL=cf_length",                             "" },
     { SSC_OUTPUT,       SSC_ARRAY,      "cf_energy_value",                        "PPA revenue net",                     "$",      "",                      "Cash Flow Revenues",             "*",                      "LENGTH_EQUAL=cf_length",                             "" },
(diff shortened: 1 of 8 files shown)


cpaulgilman (commented):
@brtietz I've got related changes in the om_var_labels branch (https://github.com/NREL/ssc/tree/om_var_labels). I can merge fix_cf_energy_sales into om_var_labels and get these changes into patch via that route. OK?

mjprilliman (approved):

ssc/cmod_communitysolar.cpp
```diff
@@ -606,8 +606,6 @@ static var_info _cm_vtab_communitysolar[] = {
     /* PPA revenue not applicable to community solar, may need to be restored later
     { SSC_OUTPUT,       SSC_ARRAY,      "cf_ppa_price",                           "PPA price",                     "cents/kWh",      "",                      "Cash Flow Revenues",             "*",                      "LENGTH_EQUAL=cf_length",                             "" },
     { SSC_OUTPUT,       SSC_ARRAY,      "cf_energy_value",                        "PPA revenue net",                     "$",      "",                      "Cash Flow Revenues",             "*",                      "LENGTH_EQUAL=cf_length",                             "" },
```
sjanzou, line 608: I am confused based on the email chain... are we keeping "net" here or does that go away when @cpaulgilman merges his changes?

brtietz: We can remove net, but this is in commented out code, so it won't change anything. Should I remove net or just remove the whole block?

sjanzou: Good point... we should be consistent and removing commented code - only keep if necessary or informative. Would you say this is clarifying or just a historic artifact? (Useful for debugging in the future?)

sjanzou (approved):
Comment removal at your discretion!

brtietz:
Included in https://github.com/NREL/ssc/pull/753, closing this PR

PR #753, opened 2022-02-09
Use consistent variable labels and group names for O&M and energy variables

cpaulgilman:
Goes with https://github.com/NREL/SAM/pull/927

Use consistent SSC group names for O&M outputs "Cash Flow Operating Expenses" (output group names do not affect PySAM). Note that Sale Leaseback uses "Sale Leaseback" for most output variables.

Remove redundant cf_sales_energy_value.

Add cf_energy_sales to outputs for all financial models.

Change cash flow energy variable labels to use "Electricity" instead of "Energy" to avoid potential confusion between thermal and electric energy.


ssc/cmod_cashloan.cpp
```diff
@@ -91,30 +91,29 @@ static var_info vtab_cashloan[] = {
     { SSC_OUTPUT, SSC_NUMBER, "total_cost", "Total installed cost", "$", "", "Financial Metrics", "*", "", "" },
 		
 
-	{ SSC_OUTPUT,        SSC_ARRAY,      "cf_energy_net",      "Energy",                  "kWh",            "",                      "Cash Flow",      "*",                     "LENGTH_EQUAL=cf_length",                "" },
-	{ SSC_OUTPUT,        SSC_ARRAY,      "cf_energy_value",      "Value of electricity savings",                  "$",            "",                      "Cash Flow",      "*",                     "LENGTH_EQUAL=cf_length",                "" },
+    { SSC_OUTPUT,       SSC_ARRAY,      "cf_energy_net",            "Electricity to grid net",       "kWh", "", "Cash Flow Electricity", "*", "LENGTH_EQUAL=cf_length", "" },
```
brtietz, line 94: These labels don't translate well to cashloan. Here is a PV-Battery residential case with grid charging turned on:

image

First, note that "net" and "to grid" are the same number, despite a sizeable number of kWh from grid. It's unclear how much of this energy actually makes it to the grid, but the monthly powerflow outputs imply roughly half:

image

The metrics chart appears to have the correct net number for year one:

image

Second, the "from grid" label needs to discuss the system somehow, as a sizeable amount of grid power is going to load.

brtietz (changes requested):
Front of meter configurations look good, O&M section looks good, behind the meter configurations need some more work and discussion. I've flagged the issues with some screenshots for cashloan, host-developer appears to have the same problems. Let me know if you want to meet to discuss these.

cpaulgilman:
Good point. 

For BTM configurations, it probably makes sense to report a single energy value in the cash flow. We could revert to what is in SAM 2021.12.02, which is to show a single value, `cf_energy_net`. In this PR, that is currently labeled "Net electricity to grid" instead of the previous label "Energy." 

In both 2021.12.02 and this PR, `annual_energy` in the metrics table is different from `cf_energy_net`, which is defined and calculated both in cmod_annualoutput.cpp here

https://github.com/NREL/ssc/blob/068cd4f83a0c0c8988a6d64797dfde4476bc798f/ssc/cmod_annualoutput.cpp#L192

and in the individual financial model compute modules, such as here for cashloan:

https://github.com/NREL/ssc/blob/068cd4f83a0c0c8988a6d64797dfde4476bc798f/ssc/cmod_cashloan.cpp#L412

Should we use a single definition of `cf_energy_net` that is consistent with `annual_energy`?

brtietz:
Displaying one energy number in the cash flow like the release makes sense to me.

As far as what that number is, it looks like all of the front of meter configurations have a line like this:

https://github.com/NREL/ssc/blob/068cd4f83a0c0c8988a6d64797dfde4476bc798f/ssc/cmod_singleowner.cpp#L2942

whereas the BTM configurations use what comes out of common_financial.

https://github.com/NREL/ssc/blob/068cd4f83a0c0c8988a6d64797dfde4476bc798f/ssc/common_financial.cpp#L3238

This means CF_energy_net is identical to CF_energy_sales for ~1700 lines of code in Single Owner. This is due to the way utility purchases were handled in 2020.11.29. Now that we've cleaned it up for 2021.12.02, I'm wondering if it makes sense to make sure net is always just "sum of gen" and the financial models should use sales (gross) or net appropriately. This has implications for CF_energy_curtailed but not much else.

I must me missing something - to me it looks like cmod_annualoutput is abandoned code that's no longer used. If not I'd be curious about how it handles availability losses vs the financial models.

cpaulgilman:
@brtietz My latest commit shows only cf_energy_net in cash flow for BTM configs:

image

ssc/cmod_cashloan.cpp
sjanzou, line 94: Another potential confusing configuration is the stand alone battery / residential, which has monthly outputs 
image

The cash flow for the "Electricity to grid net (kWh)" term, 1814kWh, shows the "Energy to load from battery (kWh)" monthly sum for year 1 (potentially a sign issue compared to no battery?):
image

sjanzou (commented):
Maybe review the "net" term in storage configurations per the comments?

ssc/cmod_cashloan.cpp
```diff
@@ -91,30 +91,29 @@ static var_info vtab_cashloan[] = {
...
+    { SSC_OUTPUT,       SSC_ARRAY,      "cf_energy_sales",          "Electricity to grid",           "kWh", "", "Cash Flow Electricity", "*", "LENGTH_EQUAL=cf_length", "" },
```
brtietz, line 95: @cpaulgilman How do you feel about these labels for behind the meter projects:

cf_energy_net: "Electricity net generation"
cf_energy_sales: "Electricity generation"

This avoids the issue @sjanzou points out here: https://github.com/NREL/SAM/pull/927#pullrequestreview-878680572 where we're putting values that go to load in a variable with a "to grid" variable.

ssc/cmod_cashloan.cpp
```diff
@@ -91,30 +91,29 @@ static var_info vtab_cashloan[] = {
...
+    { SSC_OUTPUT,       SSC_ARRAY,      "cf_energy_purchases",      "Electricity from grid",         "kWh", "", "Cash Flow Electricity", "*", "LENGTH_EQUAL=cf_length", "" },
```
brtietz, line 96: This might need a new label as well - since this does not include electricity from grid to load. "Electricity from grid to system" would be accurate for BTM systems.

ssc/cmod_cashloan.cpp
cpaulgilman: > cf_energy_net: "Electricity net generation"
> cf_energy_sales: "Electricity generation"

I think those are good labels for BTM.

sjanzou (approved):
With comments and issues referenced, the changes look good to me...

brtietz (approved):

Issue #755, opened 2022-02-10
Change cash flow dispatch variable names from "net" to "sales"


brtietz:
Currently there a bunch of dispatch outputs that include "net" in their label in common_financial:

![image](https://user-images.githubusercontent.com/5530592/153497579-d6215e1a-9762-4cd4-8cba-bc1e68ffaf5a.png)

Due to how upstream variables are handled, these are now "sales" or "gross" variables. Change these names with the next release appropriately.

PR #749, opened 2022-02-07
Fix all test memory leaks and related issues in ssc and tcs

sjanzou:

PR #756, opened 2022-02-10
Confirm net means net

brtietz:
Use consistent values for cf_energy_net throughout the code - this should always just be "sum of gen." The previous code used "positive numbers in gen" for about half the time and then switched it after the PPA calculation loop. Fixing this required updating some values of cf_energy_net to cf_energy_sales. This calculation change and the associated labels changes should address Steve's comments in https://github.com/NREL/SAM/pull/927

The goal of this PR is that no numbers with units of $ should change at the end of the calculation. Changes in values with units of kWh for behind the meter systems are expected.

Note that there are some output names that need to be changed that aren't here. That seems out of scope for this PR, so I made a follow up issue: https://github.com/NREL/ssc/issues/755

### DIFF
--- a/ssc/cmod_cashloan.cpp
+++ b/ssc/cmod_cashloan.cpp
@@ -91,9 +91,9 @@ static var_info vtab_cashloan[] = {
     { SSC_OUTPUT, SSC_NUMBER, "total_cost", "Total installed cost", "$", "", "Financial Metrics", "*", "", "" },
 		
 
-    { SSC_OUTPUT,       SSC_ARRAY,      "cf_energy_net",            "Electricity to grid net",       "kWh", "", "Cash Flow Electricity", "*", "LENGTH_EQUAL=cf_length", "" },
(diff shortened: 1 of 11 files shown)


brtietz:
Test failures are segmentation faults on mac only and should be resolved once https://github.com/NREL/ssc/pull/749 is merged

cpaulgilman:
Here's an LK script I'm using to investigate:

```
vars = ['cf_energy_curtailed','cf_energy_purchases','cf_energy_without_battery','cf_energy_net','cf_energy_sales','batt_annual_discharge_energy','fuelcell_annual_energy_discharged','annual_energy'];

for ( i=0; i<#vars; i++ )
{
	outln(vars[i] + ' = ' + get(vars[i]) );
	outln( varinfo(vars[i]));
	outln();
}

```

cpaulgilman (approved):
This is consistent with the energy variables I need for cash flow fixes in #927.

PR #758, opened 2022-02-14
Implements pv moothing refactoring SAM issue #744

sjanzou:


sjanzou:
Validated with SRCC SoCo file and tests not updated yet.
Details in attached Word document. 
Tests will be updated soon - PV Smoothing tests skipped for now; although, same files are working in SAM
Note the version upgrader sets PV Smoothing to incorrect dispatch option per attached document.
[GitHubIssuesSAM_744.docx](https://github.com/NREL/ssc/files/8060932/GitHubIssuesSAM_744.docx)

mjprilliman (commented):
Looks good, waiting for @brtietz to give final review

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
```diff
@@ -166,43 +166,192 @@ void dispatch_pvsmoothing_front_of_meter_t::dispatch(size_t year,
...
+            forecast_pv_energy = m_batt_dispatch_pvs_nameplate_ac > 0 ? forecast_pv_energy / m_batt_dispatch_pvs_nameplate_ac : forecast_pv_energy;
+
+            m_batt_dispatch_pvs_PV_ramp_interval = pv_power;
+            m_batt_dispatch_pvs_forecast_pv_energy = forecast_pv_energy;
+
+            ssc_number_t  power_to_energy_conversion_factor = m_batt_dispatch_pvs_timestep_multiplier * _dt_hour;
+            ssc_number_t battery_power_terminal = 0;
+            ssc_number_t forecast_power = 0;
+            ssc_number_t previous_power = m_batt_dispatch_pvs_outpower;
+            ssc_number_t battery_soc = m_batt_dispatch_pvs_battsoc;
```
brtietz, line 234: Why is this not `_Battery->SOC()`?

brtietz (changes requested):
Nice work getting the code running each timestep. I think there's still a little work to be done to the algorithm's SOC and the actual SOC to synchronize. See the drift here:

image

hopefully it should be an easy change. If SOC() doesn't work as a function (I spotted one case where I suspect that'd be the wrong units), energy_available(SOC()) should work instead.

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
sjanzou: @brtietz , with battery_soc = _Battery->SOC(), the results are way off:
image


with battery_soc = _Battery->SOC() /100.0, closer but much worse agreement than using m_batt_dispatch_pvs_battsoc:
image

This is why I pursued the implementation in the SAM_744_rte branch of ssc

shared/lib_battery_dispatch_pvsmoothing_fom.h
```diff
@@ -106,35 +106,22 @@ class dispatch_pvsmoothing_front_of_meter_t : public dispatch_automatic_t
 	/// Return intermediate calculations for validation - unscale by nameplate = min(system ac rating, grid interconnection limit)
     double batt_dispatch_pvs_outpower() { return  m_batt_dispatch_pvs_nameplate_ac * m_batt_dispatch_pvs_outpower; };
     double batt_dispatch_pvs_battpower() { return   m_batt_dispatch_pvs_nameplate_ac * m_batt_dispatch_pvs_battpower; };
-    double batt_dispatch_pvs_battsoc() { return   m_batt_dispatch_pvs_nameplate_ac * m_batt_dispatch_pvs_battsoc ; };
+    double batt_dispatch_pvs_battsoc() { return     m_batt_dispatch_pvs_nameplate_ac * m_batt_dispatch_pvs_battsoc ; };
```
brtietz, line 109: I think some of our output woes are coming from this line. It doesn't make sense that an output with units of % would get multiplied by a value in kW. Apologies for not catching this before.

sjanzou: Scaling updated but now the woes are coming from the resting SOC input for the pv smoothing not affecting the algorithm.
Good agreement within the ramp interval with the SOC values:
image

Issues with the low resting SOC  ( about 7% instead of the entered 50% for case shown ) causing the battery not to be able to discharge at critical ramp times and causing anti smoothing effects:
image

There seems to be an issue with the SOC value used in the pv smoothing dispatch calculated in the charge controller which uses the dispatch model as input...maybe we can average the SOC input over the ramp interval instead of taking the instantaneous value - which does not correspond exactly in the data tables, e.g. 
image


The tests have not been updated because the resting soc issue has not been resolved.

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
brtietz: We might want to rename some variables! This was a units issue. The algorithm expects batt_soc to have units of scaled energy (hours, I think) instead of percent. So if battery_energy was 0.2, batt_soc at 50% should be 0.1, not 0.5. I corrected this in cefbcf0a865c377040ed3fbc5b8f12a1f1829316

This changes the number of violations in the tests. I believe the new code is more accurate. Here is an example of the old code in the Phoenix_Validation_alloptions test file:

old_smoothing

In reality, this is a violation, since the SOC goes to zero and the battery does not provide the desired power. However since the PV smoothing algorithm as pre-run, it didn't catch this. The new code does:

new_smoothing

In fact, its two violations!

I'll get the tests cleaned up shortly, but it looks like we'll be able to get this in the patch!

shared/lib_battery_dispatch_pvsmoothing_fom.h
brtietz: The issue here was pv smoothing SOC was reporting smoothing + the next timestep's worth of power. With the changes in cefbcf0a865c377040ed3fbc5b8f12a1f1829316 - the SOC values should line up at the step when the smoothing dispatch was calculated, and then the battery SOC should increase or decrease until the next dispatch step. The second chart in https://github.com/NREL/ssc/pull/758#discussion_r809473365 shows this behavior.

brtietz (approved):

shared/lib_battery_dispatch_pvsmoothing_fom.cpp
sjanzou: Tested and I went back to the Python code - now the scaling is looking good. Great work!

shared/lib_battery_dispatch_pvsmoothing_fom.h
sjanzou: Some places where the smoothed power is a little suspect:
image

Overall, very good agreement and much better - all tests passing on windows.

Reinstating all tests in last commit and the will merge after Github actions pass...

PR #761, opened 2022-02-15
add total land area as cmod output for downstream cmods

tyneises:


tyneises:
@Matthew-Boyd Can you please define "total_land_area" for the Physical trough compute module?

cpaulgilman:
@tyneises @Matthew-Boyd I just fixed the issue in https://github.com/NREL/SAM/pull/936 with total_land_area in the UI for MSPT configs: There is now a UI equation setting total_land_area = csp.pt.sf.total_land_area on both Tower Solar Field and Tower SolarPilot Solar Field forms. 

Let me know when this SSC PR is ready to merge and I will run some tests on a SAM build from SAM PR 936.

tyneises:
@cpaulgilman The trough model doesn't need modification, so this branch should be ready to merge.

Will/should the ETES model include the new land lease operating cost? I set the land area to 0 in the model and let users calculate land costs through the % of installed or absolute $ inputs. So, I don't think any of the UI pages currently include a "total_land_area" variable. I define it (=0) in the etes cmod. Let me know if you want me to make any changes to that model.

cpaulgilman:
@tyneises If we want to add land area and/or land lease costs to ETES, I would recommend doing that separately and creating a SAM issue with either the SAM 2021.12.02 Patch 2 or SAM Fall 2022 Release milestone. I don't have an opinion one way or the other.

PR #763, opened 2022-02-17
NMC/Gr bug fix & Switch to cycle_counts vector

dguittet:
There was a bug in the calculation of the b1 coefficient in #748 that is fixed here. There were also some adaptations to get the Ua and Voc calculations to be more aligned. Many results in tests have changed.

This PR also changes the NMC/Gr life model to use the `cycle_counts` structure in the cycle model, rather than maintain its own `cycle_DOD_range` vector. However, the usage is slightly different from the cycle-calendar model. The latter is tracking a frequency table of cycles at the DODs given by the input cycle matrix for the whole simulation duration. Whereas the NMC/Gr model will be tracking each cycle separately (essentially the count is always 1 and the DODs are ordered by when the cycle is completed) and for one day at a time.

Needing to easily append new rows to the cycle_count matrix led to my switching that structure from a matrix_t to a 2d vector, so I also added some helper functions for translating between the two.


shared/lib_battery_lifetime_calendar_cycle.h
```diff
@@ -70,7 +70,7 @@ struct cycle_state {
     int rainflow_jlt;                       // last index in Peaks, i.e, if Peaks = [0,1], then jlt = 1
     std::vector<double> rainflow_peaks;
 
-    util::matrix_t<double> cycle_counts;   // Cycles, sorted by DOD bins provided in cycling_matrix
+    std::vector<std::vector<double>> cycle_counts;   // Cycles, sorted by DOD bins provided in cycling_matrix
```
brtietz, line 73: This comment is no longer strictly true, since the NMC and LMO/LTO variables don't have a cycling matrix. Can you add some text on how the data here varies by lifetime algorithm?

shared/lib_util.h
```diff
@@ -810,11 +810,38 @@ namespace util
 		}
 	};
 
+    template <class T>
+    std::vector<std::vector<T>> matrix_to_vector(matrix_t<T> mat_in)
```
brtietz, line 814: This is super helpful, thank you!

shared/lib_battery_lifetime_calendar_cycle.h
dguittet: Yes, definitely. Updated the comment and clarified also in the SSC table for cmod_battery_stateful.cpp

test/shared_test/lib_battery_lifetime_nmc_test.cpp
```diff
@@ -693,3 +698,168 @@ TEST_F(lib_battery_lifetime_nmc_test, replaceBatteryTest) {
     EXPECT_NEAR(s.average_range, 0, tol);
     EXPECT_NEAR(s.n_cycles, 0, tol);
 }
+
+TEST_F(lib_battery_lifetime_nmc_test, BTMSTest) {
+    std::ifstream in("/Users/dguittet/Desktop/batt_soc_NMCGraphite_1_0.5hr_20yrs.txt");
```
brtietz, line 703: Need to remove local paths from the tests.

brtietz (changes requested):
This mechanism for cycle counting makes more sense to me, and actually clarifies some of the NMC code.

The major required change before merging is to re-work lib_battery_lifetime_nmc_test, BTMSTest to not have local paths. More comments are nice as you have time - I think the one in the calendar cycle header file will save head scratching later.

test/shared_test/lib_battery_lifetime_nmc_test.cpp
dguittet: Oops this was erroneously committed. Let me remove this

shared/lib_util.h
dguittet: Great!

brtietz:
> I'm happy to add other comments where you think they would be helpful, any suggestions? I removed the BTMS test that was based on a bunch of large local files, which I had committed by accident.

The other one that came to mind was indicating what the DOD indices are in the NMC and LMO/LTO cpp files the first time they are used ex: `DOD[0] // cycle count`. I agree that this is more useful than what I wrote, but there are advantages to enums sometimes. ;)

brtietz (approved):
Approved, conditional on the tests passing.

PR #767, opened 2022-02-18
Remove crit_load input from PVWatts battery ReOpt calls

mjprilliman:
-Callback error for PVWatts Battery ReOpt invoke function stems from crit_load being defined even though critical load is not defined in PVWatts Battery. Added an assignment check before passing critical load to ReOpt (does not seem to be a required input)
-See https://github.com/NREL/SAM/pull/945


dguittet (approved):
Thanks!

PR #770, opened 2022-02-24
Levelized label changes

cpaulgilman:
Change variable labels for LCOE, NPV, and IRR to use consistent style like "LCOE Levelized cost of energy nominal:"

* Abbreviation precedes full name for alphabetic sorting and to facilitate searches
* Full name is capitalized (first letter only)
* Descriptor like "nominal" is at end of label with no parentheses to avoid confusion with units displayed in parentheses
* Avoid non-alphanumeric characters in labels to minimize potential parsing problems when exporting to CSV and other formats on different operating systems

Fixes https://github.com/NREL/SAM/issues/648

Goes with https://github.com/NREL/SAM/pull/954


ssc/cmod_communitysolar.cpp
```diff
@@ -698,8 +698,8 @@ static var_info _cm_vtab_communitysolar[] = {
     { SSC_OUTPUT,       SSC_ARRAY,      "cf_project_return_aftertax_npv",         "After-tax cumulative NPV",  "$", "",                      "Cash Flow Total and Returns",      "*",                     "LENGTH_EQUAL=cf_length",                "" },
 
 	// metrics table
-    { SSC_OUTPUT,       SSC_NUMBER,     "project_return_aftertax_irr",            "Internal rate of return (IRR, after-tax)",       "%",                   "", "Metrics", "*", "", "" },
-    { SSC_OUTPUT,       SSC_NUMBER,     "project_return_aftertax_npv",            "Net present value (NPV, after-tax)",             "$",                   "", "Metrics", "*", "", "" },
+    { SSC_OUTPUT,       SSC_NUMBER,     "project_return_aftertax_irr",            "IRR Internal rate of return",       "%",                   "", "Metrics", "*", "", "" },
```
brtietz, line 701: It's a little odd to me to remove after-tax from this variable label, given that we also present pre-tax. Is it sufficiently obvious in the help that the "default" IRR presentation is after tax?

brtietz (approved):
Thanks for cleaning this up! One minor comment about pre-tax vs after-tax IRR (in line), but I trust your judgement on the solution there.

mjprilliman (approved):

ssc/cmod_communitysolar.cpp
cpaulgilman: @brtietz Thanks for noticing. I see that there is some inconsistency in these IRR and NPV labels throughout the financial cmods. For example, single owner has `analysis_period_irr` and `flip_actual_irr` in common.cpp and `project_return_aftertax_irr` in cmod_singleowner.cpp. There is also a distinction between the cash flow array variables like `cf_project_return_aftertax_irr` and metric number variables like `project_return_aftertax_irr`. I will clean those up.

PR #773, opened 2022-03-02
Improve error messages for wfcheck

cpaulgilman:
Fixes https://github.com/NREL/SAM/issues/968.

With this fix, it is clearer that the problem with the weather file (see SAM Issue 968)) is that the Day column in this weather file with one-minute data is day-of-year (1-365) instead of day-of month (1-31 for Jan, 1-28 for Feb, 1-30 for Mar, etc.).

image


sjanzou (approved):
Better message!

PR #775, opened 2022-03-06
Minor fixes for PySAM

dguittet:
Add splinter as dependency to shared and tcs CMakeLists so that the build order is correct

Add missing return bool to some equations


sjanzou (commented):
Looks good, thanks for updating the dependencies

PR #778, opened 2022-03-09
Archive shared libraries

brtietz:
Add archiving steps for the linux .so file and the Mac .dylib file to the build


sjanzou (approved):
Great to have these available. Thanks!

PR #780, opened 2022-03-10
Fix sam 976 daily losses batteries

brtietz:
Rewrite some of the losses code so that daily AC losses are better accounted for, and don't cause infinite recursion for an AC-connected battery. Clean up the downstream implications so the grid and loss variables all add up. This entailed splitting losses into two categories: those that affect the output of the PV inverter, and those that affect all AC output. The latter category (system availability and daily losses) are a little fuzzy about energy tracking, but this maintains the behavior in the previous release. We are discussing ways to handle this in the next release.

This SAM file has pre-configured DC and AC systems with outages and losses:
[DC_AC_Connected.zip](https://github.com/NREL/ssc/files/8227121/DC_AC_Connected.zip)


brtietz:
Fixes https://github.com/NREL/SAM/issues/976

brtietz:
I was a little sloppy in my branching, and accidentally pushed a fix for https://github.com/NREL/SAM/issues/977 to this pull request as well. Mind reviewing both at once? A SAM file for the 2nd issue is here:
[fuel_cell_issue_977.zip](https://github.com/NREL/ssc/files/8227736/fuel_cell_issue_977.zip)

If not, I can use git cherry-pick to fix things.

cpaulgilman:
> 



> I was a little sloppy in my branching, and accidentally pushed a fix for [NREL/SAM#977](https://github.com/NREL/SAM/issues/977) to this pull request as well. Mind reviewing both at once? A SAM file for the 2nd issue is here: [fuel_cell_issue_977.zip](https://github.com/NREL/ssc/files/8227736/fuel_cell_issue_977.zip)
> 
> If not, I can use git cherry-pick to fix things.

@brtietz I'll review both, no need to cherry pick. Thanks.

cpaulgilman:
I confirmed that this fixes https://github.com/NREL/SAM/issues/977: Electricity to battery from fuel cell is now consistent with results from SAM 2020.12.02 before r1.

brtietz:
Great testing! For the first chart, I went with adding a DC lifetime losses output, which I'll push shortly. The second chart has exposed some additional issues on how the powerflow code handles AC losses, in the 100% AC loss case, a DC connected battery should be unable to meet the load. It looks like I'm going to have to overhaul the AC losses code in the powerflow code (which to be fair, is pretty new), let me confirm a few things to make sure I get this right in the fix.

For a DC connected battery, the AC losses on the losses page should apply to all power - for example, the 1% wiring loss increases power drawn from the grid during grid charging:

image

For an AC connected battery, that sort of loss shows up in the AC/DC conversion efficiency instead, so the wiring loss only applies to the PV inverter:

image

Does that match your understanding of the assumptions for the losses page? The battery code currently treats all of the AC losses (wiring, transformer, availability, and lifetime) the same way, so if I need to break any of those out so they can be treated differently, let me know. Happy to discuss during the SAM meeting or a side meeting.

cpaulgilman:
The unclear role of inputs on the Losses page for PV-Battery systems is because those inputs were originally designed for PV only systems. I think you are correct that the AC loss inputs currently only apply to the output of the PV inverter, and I think it makes sense to preserve that behavior for now.

There is a larger question of how to better manage all of the loss inputs, some of which seem to overlap: For PV-battery systems, the Shading and Layout, Losses, Battery Cell and System, Grid Limits, and Lifetime and Degradation pages all contain various kinds of loss inputs. "Grid Curtailment", "Availability" and "Lifetime Daily Losses" all include AC losses that are applied in different ways and in different parts of the code in addition to the transformer, transmission, and AC losses. I think this is worth discussion at some point and perhaps cleaning up for a future release.

brtietz:
@cpaulgilman Ah, no. It's "Given a 100% loss, is the action in column A allowed?" Also I made some changes since the version posted above, the latest version is here:
[ac_losses_matrix.xlsx](https://github.com/NREL/ssc/files/8384498/ac_losses_matrix.xlsx)

cpaulgilman:
I think the behavior looks correct now.

It's a bit confusing that hourly kWh and lifetime kW "electricity to/from grid" are different in these results for the MiniGrid AC case of the DC_AC_Connected.sam test file (this is true in this branch and SAM 2021.12.02 r1):

ac-daily-losses-with-outage-ac-connected

brtietz:
"Electricity to/from grid (year 1 hourly)" is computed in utility rate 5. I tried to handle the grid outage variables in cmod_grid, but it looks like utility rate 5 might need to reference the grid outage variable as well in order to correct that output.

Any objections to fixing that in a second issue? I feel like this pull request already has enough stuffed into it.

cpaulgilman:
No objections from me. I was hesitant to bring it up here for that reason.

cpaulgilman (approved):

PR #781, opened 2022-03-15
Update translate schedule inputs to handle 36 tou periods

brtietz:
Add support for up to 36 time of use periods in SSC by changing the max number of inputs on the translate schedule function. Add a unit test to confirm support.

Currently, the workaround for using this in the GUI is to save the utility rate to a CSV, modify the periods, and re-load it. Discuss whether we want to add GUI support in the patch or save it for the release.


brtietz:
Partial fix for https://github.com/NREL/SAM/issues/28

brtietz:
Test file with 24 rates and associated weather file available here:
[24_rates_test.zip](https://github.com/NREL/ssc/files/8255563/24_rates_test.zip)

cpaulgilman (approved):
Looks good in SAM UI.

I added a note to https://github.com/NREL/SAM/issues/28 to address issue with column width in "Electricity Rate Data by Tier and Period" table when we fully implement this in the SAM UI.

sjanzou (approved):
Everything in ssc looks good. The UI in SAM needs to be updated per the comment in [NREL/SAM#28](https://github.com/NREL/SAM/issues/28)

PR #787, opened 2022-03-31
Lifetime prices for automated fom dispatch

brtietz:
- Upgrade the forecast_price_signal class to build a lifetime array of prices for both the merchant plant and single owner
- Convert all year 1 indices in the fom battery dispatch code to lifetime indices
- Add additional merchant plant variables to forecast_price_signal to properly account for constant-percentage cleared capacity inputs
- Reinforce the outputs of forecast_price_signal to be in $/kWh to match with how automated dispatch uses the outputs
- Update unit test values given new dispatch patterns

Recommended tests include:
- A PPA rate that starts smaller than the cycle degradation penalty, but escalates to larger than the penalty (battery shouldn't dispatch in the first year(s), but then should dispatch later)
- Merchant plant rates that change drastically year to year (including $0 prices some months/years - the battery shouldn't dispatch during these)
- Different lengths of merchant plant inputs

Fixes https://github.com/NREL/SAM/issues/894


sjanzou (commented):
The code changes look fine... please review the project file below for the first test case (PPA below penalty with high escalation rate) 

Attached file has two cases 
[ssc_787.zip](https://github.com/NREL/ssc/files/8396920/ssc_787.zip)

1. PPA with suspect results with ppa below penalty for first several years but has a couple of suspect power targets in the following screenshot
image

2. MP with year 1 zeroed out - expected behavior
image

brtietz:
Good question!

Discharging is based on: https://github.com/NREL/ssc/blob/f4460db1ab1ea6d3343e8515055f6524f13b60d3/shared/lib_battery_dispatch_automatic_fom.cpp#L235

and: https://github.com/NREL/ssc/blob/f4460db1ab1ea6d3343e8515055f6524f13b60d3/shared/lib_battery_dispatch_automatic_fom.cpp#L301

At the beginning of the PPA simulation, the high value period is worth $0.145 (1.5 TOD factor), so the revenue to discharge equation says discharging is worth $0.0392/kWh. Note that the battery discharges in the first high cost period in the first day.

Subsequently, the battery does not charge. Instead, the lifetime degradation slowly reduces the total capacity. This decreases the denominator for SOC, causing SOC to slowly increase:

image

After enough time, this SOC increase triggers additional discharging (which is even more profitable). Finally the price escalates enough to make revenue to charge > 0, which is when the cycling starts later.

If I further lower the PPA price to the point where revenue to discharge is < 0 ($0.05/kWh), we get the expected delay in discharging:

image

sjanzou (approved):
Thanks for clarifying the dispatch behavior with the low ppa price and battery discharge spikes.

PR #788, opened 2022-04-01
Fix UI crash when custom field HTF selected

tyneises:
Issue: https://github.com/NREL/SAM/issues/964

### DIFF
--- a/ssc/cmod_csp_common_eqns.cpp
+++ b/ssc/cmod_csp_common_eqns.cpp
@@ -604,6 +604,18 @@ double Min_field_flow_velocity(double m_dot_htfmin, double min_inner_diameter,
         min_inner_diameter * min_inner_diameter);
 }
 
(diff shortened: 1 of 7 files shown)


tyneises:
@cpaulgilman Do we want to disable the Edit button when a library HTF is selecting? With the currently problems with invalid custom HTFs, it might be nice to be able to edit the HTF properties before selecting the custom HTF in the drop down.

cpaulgilman:
> @cpaulgilman Do we want to disable the Edit button when a library HTF is selecting? With the currently problems with invalid custom HTFs, it might be nice to be able to edit the HTF properties before selecting the custom HTF in the drop down.

The [SAM custom-htf branch](https://github.com/NREL/SAM/tree/custom-htf) disables the Edit button (MaterialProperties) for all CSP models unless the "user-defined" option is selected for the HTF. I think this makes it clear that you can only edit custom HTF properties. If we don't disable the Edit button, people might think they can edit the properties of the non-custom HTFs in the list.

cpaulgilman (approved):

PR #790, opened 2022-04-04
783 sequential wiring and transmission losses

brtietz:
Apply the wiring loss and _then_ the transmission loss, so that the numbers can't be greater than 100%. Update the loss calculations so this appropriately shows up in the loss diagram. Transmission losses were added in SAM PR [1000](https://github.com/NREL/SAM/pull/1000)

I'm happy to add a unit test for this, but wanted to confirm the loss behavior with you prior to writing it.

### DIFF
--- a/shared/lib_pv_io_manager.cpp
+++ b/shared/lib_pv_io_manager.cpp
@@ -881,10 +881,6 @@ void PVSystem_IO::AllocateOutputs(compute_module* cm)
     }
 
 }
-void PVSystem_IO::AssignOutputs(compute_module* cm)
-{
-    cm->assign("ac_loss", var_data((ssc_number_t)(acLossPercent + transmissionLossPercent)));
-}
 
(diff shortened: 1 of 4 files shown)


janinefreeman:
A little testing looks good, one thing this does bring up is that the warning about the inverter being oversized seems to get triggered downstream of the actual inverter calculation (after AC and transmission losses), which is probably not the intent of that warning. Possibly a separate issue? Thoughts?

brtietz:
> A little testing looks good, one thing this does bring up is that the warning about the inverter being oversized seems to get triggered downstream of the actual inverter calculation (after AC and transmission losses), which is probably not the intent of that warning. Possibly a separate issue? Thoughts?

I'm willing to lump that in here - the tricky part is that there isn't currently a timeseries output that has the data needed to fix that issue. Would adding something like "inverter output AC gross" as a timeseries output be helpful? Preferences on naming?

janinefreeman (approved):

PR #791, opened 2022-04-04
Sam 978 annual vars to lifetime

brtietz:
Extend variables in pvsamv1 and cmod_battery to lifetime variables, as per request in https://github.com/NREL/SAM/issues/978 This fits the scope of a patch change.

Larger changes may occur in https://github.com/NREL/SAM/issues/926


cpaulgilman:
This looks good! I checked Detailed PV and Detailed PV - Battery configurations for BTM and FOM systems at hourly and 15-minute time steps.

One issue is the Data Tables category says "Three Hour Data" instead of "Lifetime Hourly Data" or "Lifetime 15 Minute Data."

image

brtietz:
Hmm, that's new, and I suspect is caused by this recent change in the SAM repository: https://github.com/NREL/SAM/blob/c8ef80585337bebce4753f6d2f654ba3226b4b1b/src/results.cpp#L159 I get the same issue on the patch branch of ssc.

@mjprilliman - can you walk us through why that changed from row_length == 2920? Putting a multiple of 8760 through that code results in an fmod of 0. Would it work to just move that later in the if/else chain?

mjprilliman:
> Hmm, that's new, and I suspect is caused by this recent change in the SAM repository: https://github.com/NREL/SAM/blob/c8ef80585337bebce4753f6d2f654ba3226b4b1b/src/results.cpp#L159 I get the same issue on the patch branch of ssc.
> 
> @mjprilliman - can you walk us through why that changed from row_length == 2920? Putting a multiple of 8760 through that code results in an fmod of 0. Would it work to just move that later in the if/else chain?

That was a remnant of the JPD file generation work that probably should have not been included in the merge or at the very least tested more thoroughly. I think the best course of action is for me to revert that.

cpaulgilman:
> > Hmm, that's new, and I suspect is caused by this recent change in the SAM repository: https://github.com/NREL/SAM/blob/c8ef80585337bebce4753f6d2f654ba3226b4b1b/src/results.cpp#L159 I get the same issue on the patch branch of ssc.
> > @mjprilliman - can you walk us through why that changed from row_length == 2920? Putting a multiple of 8760 through that code results in an fmod of 0. Would it work to just move that later in the if/else chain?
> 
> That was a remnant of the JPD file generation work that probably should have not been included in the merge or at the very least tested more thoroughly. I think the best course of action is for me to revert that.

Resolved in https://github.com/NREL/SAM/pull/1001

cpaulgilman (approved):

Issue #745, opened 2022-02-02
Improve performance of FOM automated dispatch with highly variable costs


brtietz:
See user issue with automated dispatch here: https://sam.nrel.gov/forum/forum-general/2962-battery-automated-dispatch-version-2020-2-29.html?start=6#10536

The issue is that dispatch only discharges at the time of the max price, so the manual dispatch is able to discharge at a larger number of profitable hours when compared to automated dispatch

Try to improve the automated dispatch either by implementing a threshold relative to the max price, or by computing the most profitable number of hours similar to price signals dispatch. Benchmark both the runtime vs the current algorithm, and performance vs the optimal dispatch in HOPP.

PR #792, opened 2022-04-06
Ssc 745 improve fom dispatch

brtietz:
Add additional variables to the FOM dispatch code such that the price to charge and discharge reflects the capacity of the battery, rather than just the maximum price in the forecast.

File for testing and comparison to HOPP:
[autodispatch_improvements_hopp.zip](https://github.com/NREL/ssc/files/8431864/autodispatch_improvements_hopp.zip)
 
Fixes https://github.com/NREL/ssc/issues/745

A note on save_as_JSON_test_run: the NPV in the current defaults goes up by ~2%, so it's not clear to me why this one is going down by <1%.


dguittet (approved):
LGTM

PR #800, opened 2022-04-13
Ssc 561 update for demand ratchets

brtietz:
Update price signals dispatch to support kWh/kW rates without crashing. The key difference here is rate_data.init_energy_rates can now be called for a single month, and peak data by TOU period is now passed into the rate calculation function for use with these rates.

### DIFF
--- a/shared/lib_battery_dispatch_automatic_btm.cpp
+++ b/shared/lib_battery_dispatch_automatic_btm.cpp
@@ -159,18 +159,19 @@ void dispatch_automatic_behind_the_meter_t::setup_rate_forecast()
 {
     if (_mode == dispatch_t::FORECAST)
     {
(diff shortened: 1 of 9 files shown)


brtietz:
File for testing the fix:
[West_Campus_Model_Net_Load_CorrectE35.zip](https://github.com/NREL/ssc/files/8483303/West_Campus_Model_Net_Load_CorrectE35.zip)
(from https://sam.nrel.gov/forum/forum-general/3832)

sjanzou (approved):
Tested with current release and crashes and with changes in this pull request with 2.5% and 0% rate escalation in the following zip file and note year 1 bill values higher for system than following year for both cases...for zero escalation, flip happens in July without system and November with system - interesting setup:
image

[ssc_800.zip](https://github.com/NREL/ssc/files/8488464/ssc_800.zip)

PR #803, opened 2022-04-26
fixed dispatch reporting issue

qualand:
LpSolve presolve was resulting in time periods where the reported dispatch state of charge was zero when it actually was not. This occurred at the beginning of the dispatch horizon sometimes. To fix, now dispatch solution is saved using the original model form and not the subset model after presolve. This results in all variable values being updated.


brtietz:
@qualand - develop has the gtest fix now: https://github.com/NREL/ssc/pull/805

PR #804, opened 2022-04-26
pin the version of gtest to the last one that worked with our current tests

brtietz:
https://github.com/NREL/ssc/runs/6181437239?check_suite_focus=true failed even though the build before it succeeded, the PR succeeded, and the build succeeds locally. It looks like some of our tests are incompatible with the latest version of gtest. Revert to the last version that worked, and fix things in an issue that is to be created.


sjanzou (approved):
Thanks

PR #813, opened 2022-05-07
Address issues 2 and 3 in SAM issue 836

sjanzou:


sjanzou:
Update tests inputs and outputs and all passing on Windows:
image

ssc/cmod_merchantplant.cpp
```diff
@@ -1333,7 +1333,9 @@ class cm_merchantplant : public compute_module
 			cf.at(CF_om_opt_fuel_2_expense,i) *= om_opt_fuel_2_usage;
 		}
 
-        std::vector<double> mp_energy_market_price = lookup("mp_energy_market_price")->arr_vector();
+        std::vector<double> mp_energy_market_price(8760*nyears, 0.0);
+        if (lookup("mp_energy_market_price"))
+            std::vector<double> mp_energy_market_price = lookup("mp_energy_market_price")->arr_vector();
```
brtietz, line 1338: Re-declaration of mp_energy_market_price almost certainly is generating a compiler warning.

ssc/cmod_merchantplant_eqns.cpp
```diff
@@ -153,7 +153,7 @@ bool mp_ancillary_services(ssc_data_t data)
 		std::vector<ssc_number_t> ancillary_services3_revenue(nsteps, 0.0);
 		std::vector<ssc_number_t> ancillary_services4_revenue(nsteps, 0.0);
 
-		if (!ancillary_services_success)
+//		if (!ancillary_services_success)
 		{
```
brtietz, line 157: Not sure why the bracket is still here if the if statement has been commented out.

brtietz (changes requested):
Great work solving the bugs! A few comments on code cleanup that should be addressed.

One other question: is running the merchant plant with no revenue a valid case? It seems like a user would most likely end up in that situation by accident, and at least a warning (if not an exec error) would be helpful.

sjanzou:
> 

With no revenue streams enabled, the merchant plant now runs with (SSC_WARNING) message:
image

ssc/cmod_merchantplant.cpp
sjanzou: Updated, thanks!

ssc/cmod_merchantplant_eqns.cpp
sjanzou: Updated confusing indentation and fixed brackets. Thanks!

cpaulgilman (approved):
Looks good after my revision of the error message:

image

Issue #815, opened 2022-05-16
PV Smoothing implementation does not account for dispatch SOC limits


brtietz:
**Describe the bug**
PV smoothing dispatch can produce dispatch plans that lead to violations that aren't recorded, since the smoothing algorithm things all of the energy in the battery is available, when it may be subject to SOC limits.

The screenshot below has a minimum SOC of 15%:

![image](https://user-images.githubusercontent.com/5530592/168671218-ef820344-8f4d-4b8a-ac2f-6bf9eeee486b.png)

In the red circle (and again the next day) PV smoothing expects continued battery discharge.

**To Reproduce**
1. Run the attached SAM file: 
[pvsmoothing_soc_issues.zip](https://github.com/NREL/ssc/files/8703031/pvsmoothing_soc_issues.zip)

2. Compare PV Smoothing Outpower and Gen for the limited SOC case

**Expected behavior**
PV Smoothing should be aware of dispatch limits

**Desktop (please complete the following information):**
 - Version 2021.12.02r1

PR #816, opened 2022-05-16
add soc limits to pv smoothing

brtietz:
Fixes: https://github.com/NREL/ssc/issues/815 see that issue for test file.

Note that there is some weird behavior where the PV smoothing output power is negative at night. I view that as a tuning or sizing issue, and out of scope for this PR.


sjanzou (approved):
The soc limit is indeed recording another violation, the output power continues to discharge on next step for both cases. I think this is consistent with the original code and the smoothing paper:
image

with full soc - no additional violation
image

PR #818, opened 2022-05-18
Fix cash flow issues

cpaulgilman:
Goes with https://github.com/NREL/SAM/pull/1044

Community Solar: 

* Fix receivables reserve to calculate based on subscriber payment revenue instead of PPA revenue (PPA revenue not available for community solar).

* Comment out PPA-related output variables to avoid outputs with null values.

Sale Leaseback: 

* Add battery and fuel cell operating costs to cash flows.


ssc/cmod_communitysolar.cpp
```diff
@@ -3918,17 +3930,15 @@ class cm_communitysolar : public compute_module
 		if (npv_energy_nom != 0) lcog_nom = lcog_nom * 100.0 / npv_energy_nom;
 		assign("lcog", var_data((ssc_number_t)lcog_nom));
 
-
-
-
 		// dispatch
-		std::vector<double> ppa_cf;
+        /* Community solar not available with storage, need to address price signal dispatch with no PPA if we enable with storage
```
brtietz, line 3934: First we would need to decide whether or not we were including load.

We could always choose to disable the battery options that require prices. With load that would be peak shaving/grid power targets, custom and manual. Without load that would be pv smoothing, manual, and custom.

brtietz (approved):
Thanks for explaining why you commented out code! SSC changes look good, some issues with the SAM half of this PR.

sjanzou (approved):

mjprilliman (approved):

PR #820, opened 2022-05-18
Fix standalone battery capacity factor, add sales capacity factor

### LINKED ISSUE #819: Capacity Factor reporting NaN for standalone battery configs
**Describe the bug**
In cmod_battery, "capacity_factor" defaults to zero when not specified. As standalone battery has no technology module before cmod_battery, this results in a capacity factor of 0 which leads to divide by zero error in the initialization of capacity factor in the battery module. 

**To Reproduce**
1. Go to Standalone Battery Case
2. Click Simulate
3. Search 'capacity' in Data Tables
4. See NaN

**Expected behavior**
Capacity factor based on gen would be negative for a standalone battery. Capacity factor based on gen sales would be positive

mjprilliman:
-Remove 0 initialization for capacity factor to allow for capacity factor calculation
-Add capacity factor based on energy sales within gen for standalone battery, generic battery (should it be there for all technologies?)
-Fixes #819


ssc/cmod_battery.cpp
```diff
@@ -2104,7 +2104,9 @@ static var_info _cm_vtab_battery[] = {
...
     { SSC_INPUT,        SSC_NUMBER,      "run_resiliency_calcs",                       "Enable resilence calculations for every timestep",        "0/1",        "0=DisableCalcs,1=EnableCalcs",                  "Load",    "?=0",                    "",                               "" },
-    { SSC_INOUT,        SSC_NUMBER,      "capacity_factor",                            "Capacity factor",                                         "%",          "",                     "System Output",                             "?=0",                    "",                               "" },
+    { SSC_INOUT,        SSC_NUMBER,      "capacity_factor",                            "Capacity factor",                                         "%",          "",                     "System Output",                             "",                    "",                               "" },
+    { SSC_OUTPUT,       SSC_NUMBER,      "capacity_factor_sales",                      "Capacity factor based on energy sales",                                         "%",          "",                     "System Output",                             "",                    "",                               "" },
```
brtietz, line 2108: I think our preferred language here is "Capacity factor based on energy to grid" - "energy sales" is sort of internal code lingo.

brtietz (approved):
Code makes sense, pending Paul's approval of names. Does there need to be an associated SAM PR to add this to the stand-alone battery metrics table?

cpaulgilman (approved):
Looks good after my label changes per email discussion.

PR #821, opened 2022-05-19
Generalize UDPC method

jmctigue:
The proposed code generalizes the UDPC interpolation methods so that each variable can use more than three "levels" thereby enabling more input data to be provided which can lead to reduced errors in the interpolation.

The new method is included in files that are based on existing files. The new files are suffixed *_jdm.  Furthermore, existing classes were used as a template and new code was written in classes which also end *_jdm. The next step will be to streamline these files and classes in such a way that the original methods are maintained whilst also adding the generalized method.

Existing files that were modified are:

ssc/cmod_test_ud_power_cycle.cpp: write out lots of data over a range of mass flow rates. The data shows the error in the interpolation method. A similar data output occurs in the generalized method (ssc/cmod_test_ud_power_cycle_jdm.cpp) so that the errors of the two methods can be compared.

tcs/ud_power_cycle.cpp: writing data to log file. This was used to check the new method was behaving consistently with the old method. These changes should be removed.

CMakeLists.txt and sscapi.txt were also updated to include the new files.


tyneises (changes requested):
@jmctigue Can you please remove, comment out, or move into a conditional statement (e.g. if(is_log) where is_log defaults to false) any code that writes to log files?

PR #823, opened 2022-05-26
Fixes SAM issue 1004

sjanzou:
cmod_merchant_plant_eqns updated to handle single value inputs
extrapolate_timeseries updated to handle single value array input


brtietz (changes requested):
Fix works well for a single value, but annual, monthly, and weekly all appear to be broken (no revenue with price set to $10/MWh and 100% cleared capacity). I've updated the test file with those additional cases:
[merchant-single-value-bug.zip](https://github.com/NREL/ssc/files/8780123/merchant-single-value-bug.zip)

sjanzou:
All modes fixed nd tested for all merchant plant revenue streams.

Test project files with 25 year and 2 year analysis period with one and all 5 merchant plant revenue streams enabled
[SAM_1004.zip](https://github.com/NREL/ssc/files/8786241/SAM_1004.zip)
.

brtietz (approved):
Works great, thank you!

Issue #825, opened 2022-05-27
Merchant plant to battery forecast input sums inappropriately


brtietz:
**Describe the bug**
If you provide a price of $20/MWh to just the energy market in the merchant plant, the battery will receive a price for that timestep of $20/MWh. If you provide that same price to all 5 markets, those values will be summed, and the battery will assume it's receiving $100/MWh. While the percentage price differences remain the same between steps, the magnitude matters when comparing to the cost of cycle degradation.

**To Reproduce**
1. Run the attached SAM file: [aligned price series.zip](https://github.com/NREL/ssc/files/8789875/aligned.price.series.zip)
2. Compare the "power price for battery dispatch" between the all energy market case and the all markets case.

**Expected behavior**
If we're only going to provide one price signal to battery dispatch, it should be scaled appropriately to the number of markets active. Discuss whether we can use a simple average or a weighted average, and if this needs to factor into the solution for https://github.com/NREL/SAM/issues/181

brtietz:
Easiest valid approach: weighted average. This has potential issues with cleared capacity as a MW value, where the battery might not decide to dispatch & fail to meet the cleared capacity. See if 181 can address this or if we need to do something else.

Battery price computation in a difference cmod: break this out of the battery code since it potentially contributes to financial model modularity.

PR #826, opened 2022-05-27
Ssc 599 negative cleared capacity

brtietz:
Improve handling of cleared capacity throughout merchant plant code, and some bugfixes for the battery time series prices.

- Only use positive revenue values in revenue row of cashflow table, to match other financial models
- Negative values of revenue are now assigned to cf_utility_bill, with new consumed_cost outputs
  - Note that the previous code had a copy-paste error vs the other financial models, and was off by a factor of 1000 for a kW to MW conversion, hence the large increase in NPV in the two unit tests.
- common.cpp now plays more nicely with cleared capacity as a percent of generation, so that dispatch is the same between single owner and merchant plant.
- Users get a warning (not an error) if the negative cleared capacities don't match generation

The result is that given the same price series for one year, merchant plant and single owner now (once again?) produce the same financial results. Test file here:
[aligned price series.zip](https://github.com/NREL/ssc/files/8789953/aligned.price.series.zip)
 
Differences in the "all markets" case are worth some debate, but I worried they'll be late for the patch, so I created another issue here: https://github.com/NREL/ssc/issues/825


sjanzou (changes requested):
Great that the SO and merchant plant results are aligning!

Two issues:
1. "capacity_factor_sales" is missing from results in Summary tab in SAM, e.g.
image

2. The default results have changed significantly for SAM - please review ssc_599_negative_cleared_capacity branch of SAM and pull request https://github.com/NREL/SAM/pull/1058

sjanzou (approved):

PR #802, opened 2022-04-20
Improve module ceofficient generator error reporting

cpaulgilman:
This PR replaces the generic error message for cmod_6parsolve "could not solve, check inputs" with more informative error messages based on the sanity check performed on the inputs here: https://github.com/NREL/ssc/blob/5ec04426cdfc4f83c397feb7c1ef66753ab9a70e/shared/6par_solve.h#L307.

Addresses https://github.com/NREL/SAM/issues/17

To test, create a Detailed PV / No Financial case, choose the "CEC Performance Model with User Entered Specifications" option on the Module page, try different unrealistic values for the input parameters, and click **Calculate and plot**.

Alternatively, build SDKtool and try modifying parameter values in the attached LK script for SDKtool:
[coeffgen6par_sdktool.zip](https://github.com/NREL/ssc/files/8522995/coeffgen6par_sdktool.zip)

PR #827, opened 2022-05-29
Sixparsolve error reporting

cpaulgilman:
Update #802 to fix uninitialized variable error.

To test:

1. Create a Detailed PV / No Financing case.
2. On the module page, switch to CEC Performance Model with User Entered Specifications.
3. Change short circuit current to 25 Amps
4. Click **Calculate and plot**: "Sanity check failed" error with details should appear.
5. Click **Simulate**: "Could not solve for normalized coefficients" exec fail message should appear.


sjanzou (approved):
Performing test in pull request works as described! Great update
image
image

PR #829, opened 2022-05-31
Update module entry to reflect actual open EI compatability (version 8)

brtietz:
Noticed that the PySAM documentation was out of date, requires fixing this definition in SSC.


cpaulgilman (changes requested):
I think "revenue" in that context is misleading. I would use "Electricity bill calculator." or possibly "Electricity bill calculator based on OpenEI URDB rate data format."

cpaulgilman (approved):

Issue #832, opened 2022-06-02
Voltage table can restore charge without properly accounting for powerflow


brtietz:
**Describe the bug**
Due to https://github.com/NREL/ssc/issues/569, the voltage table sometimes ends up outside of the state of charge bounds specified on the dispatch page. When this happens, the following timestamp gets an increase in charge to the minimum SOC without properly accounting for the battery charging in the powerflow equations. This results in free energy and a round trip efficiency >= 100%.

**To Reproduce**
1. Create a generic-battery commercial case
2. Change the battery chemistry to lead acid - flooded
3. Run the simulation
4. Check the annual energy charged vs discharged
5. See that the annual energy discharged is >= charged

**Expected behavior**
With the defaults, the maximum round trip efficiency should be 92.16% - this would be a 100% DC-DC efficiency. While this is also unrealistically high, we plan to address it in https://github.com/NREL/ssc/issues/831

Fixing https://github.com/NREL/ssc/issues/569 would also assist with this issue, but I view 569 as a problem with the voltage code and this as a problem with the capacity code.

**Desktop (please complete the following information):**
 - Version 2021.12.02 r1 (and possibly earlier)

PR #833, opened 2022-06-02
add function that was enforcing dispatch soc limits for lithium ion batteries

brtietz:
Fix issue #832 by ensuring the check_soc function is applied to both battery types.


cpaulgilman (approved):
This improves the battery roundtrip efficiency reported by the "default" Generic Battery / Commercial configuration from 100% to 91.53%.

Note that there is not a configuration in the defaults for lead acid batteries, so I tested this by creating a default case and changing the battery type without changing any other parameters based on the approach reported by a user here: https://sam.nrel.gov/forum/forum-general/3890.

PR #851, opened 2022-07-13
Fix label for utility escalation rate

mjprilliman:
-Closes https://github.com/NREL/SAM/issues/1085
-Fix label from battery salvage value to utility escalation rate (including inflation)
-Is this output necessary?


sjanzou (approved):

sjanzou:
@mjprilliman, the result is "necessary" insofar as to provide the user (and SAM support) the necessary outputs to fully understand the utility bill calculation.

PR #852, opened 2022-07-15
Fix issue with cmake finding debug paths for SSC tests

cpaulgilman:
* Need to verify this does not break build process for Linux and Mac

* Depends on single environment variable GTDIR pointing to top-level googletest folder as described in revised Windows build instructions: https://github.com/NREL/SAM/wiki/Windows-Build-Instructions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,12 @@ jobs:
     runs-on: ubuntu-18.04
 
     steps:
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v1.12
+      with:
+        cmake-version: '3.24.x'
+    - name: Test cmake version
+      run: cmake --version
     
     - name: Set relative paths
       run: | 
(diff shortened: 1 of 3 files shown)


cpaulgilman:
Related to https://github.com/NREL/SAM/issues/1013

sjanzou:
With VS2019 using develop branches of all other repos, 45/46 projects build without issue. 
image

The debug version of Test fails with 27 unresolved externals:
image

Release version of Test includes gtest.lib:
image

But debug version does not:
image

sjanzou:
Still going through process for VS2022, but the current limiting issue is Microsoft's limitation to not mix debug and release .lib files in a single project build... as mentioned here:
https://docs.microsoft.com/en-us/answers/questions/321574/c-cmake-debug-configuration-visual-studio-2019.html

The configuration currently needs both the debug and release version of googletest and I am updating for VS2022.

Sorry for leading you down the release only path when linking libraries to the ssc test folder. I will update the CMakeLists.txt after both the Debug and Release versions of the Test project are working in both VS2019 and VS2022.

sjanzou:
Also, note that latest stable version of wxWidgets (released 7/7/2022) is 3.2.0 - we should update Build instructions and for Fall 2022 release. 

image

brtietz (approved):

PR #854, opened 2022-07-28
Fixed utility escalation rate application in BTM battery grid charging costs

### LINKED ISSUE #853: BTM Cost of charging battery from grid not calculating correctly for LCOS
**Describe the bug**
image

The cost to charge the battery from grid is returning as zero for cases with non-zero grid charging. This is due to incorrect application of the annual series of utility escalation rate values, which are not passing values from a function call but rather zeros. 

**To Reproduce**
1. Run the attached .sam file on develop or the most recent patch of SAM
2. View annual kWh charged from grid and $ cost to charge from grid in Data Tables tab

**Expected behavior**
The utility escalation rate should be applied as a percentage increase from unity (1). 

**Additional context**

mjprilliman:
-Utility escalation was being applied as 0, eliminating calculations from hourly energy and demand charges
-Closes #853

### DIFF
--- a/ssc/common_financial.cpp
+++ b/ssc/common_financial.cpp
@@ -3355,14 +3355,16 @@ void escal_or_annual(int cf_line, int nyears, const std::string& variable,
 {
     size_t count;
     ssc_number_t* arrp = cm->as_array(variable, &count);
-
+    double check;
     if (as_rate)
     {
         if (count == 1)
         {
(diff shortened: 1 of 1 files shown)


mjprilliman:
This is specific to the cost of charging the battery from the grid used in LCOS calculations and is not relevant to the overall cost of parasitic loads provided by the grid.

cpaulgilman (approved):

PR #856, opened 2022-08-01
O and M costs in BTM and FOM dispatch

sjanzou:


shared/lib_battery_dispatch_automatic_btm.cpp
```diff
@@ -621,7 +622,7 @@ void dispatch_automatic_behind_the_meter_t::cost_based_target_power(size_t idx,
         plans[i].num_cycles = 0;
         plan_dispatch_for_cost(plans[i], idx, E_max, startingEnergy);
         UtilityRateForecast midDispatchForecast(*rate_forecast);
-        plans[i].cost = midDispatchForecast.forecastCost(plans[i].plannedGridUse, year, hour_of_year, 0) + cost_to_cycle() * plans[i].num_cycles - plans[i].kWhRemaining * plans[i].lowestMarginalCost;
+        plans[i].cost = midDispatchForecast.forecastCost(plans[i].plannedGridUse, year, hour_of_year, 0) + cost_to_cycle() * plans[i].num_cycles + plans[i].kWhRemaining * omCost() - plans[i].kWhRemaining * plans[i].lowestMarginalCost;
```
brtietz, line 625: kWhRemaining is not the right variable to multiply here - that represents energy stored in the battery. What needs to happen is the following:

1. Add a new variable kWhDischarged to the plans object
2. Set that variable in plan_dispatch_for_cost, it should be the sum of the positive values of plan.plannedDispatch at the function.
3. Multiply by that value here.

brtietz (changes requested):
Thanks for getting this set up. Front of meter looks good. Behind the meter requires some changes to multiply by the correct kWh value. Details on the appropriate line.

brtietz (approved):
Looks great, thank you!

PR #858, opened 2022-08-12
Financial Model Tests for cashloan

sjanzou:
Added the 23 default configurations from SAM to cmod_cashlon_test.cpp. Prototype for generating tests for other Financial (or any other) Compute Modules. See cmod_cashloan_test.h public Test function for general structure of testing inputs against specific configuration outputs and the lk function simulate_ssc_tests in corresponding SAM pull request.


brtietz (approved):
Excellent work! We should have a larger conversation about which metrics to test for which case. For example, I'd vote for adding "lcos_nom" to any battery cases. Looking forward to expanding on this paradigm for additional models!

PR #859, opened 2022-08-15
Task3 TES modularity

Matthew-Boyd:
Made TES sub-component more modular, made some parameter names more generic, and added a test. Perspective was mostly from the physical trough cmod.

Goals (big picture):
- Create an abstracted container with a uniform interface for all kinds of thermal storage
- Allow straightforward creation and standalone operation
- Allow focused coded development by non-system architects

Purpose:
- Enables the system controller to interact with any kind of thermal storage in the same way
- Enables low-level tests, quicker debugging and easier implementation into new system designs
- Allows more development by a broader team of coders

Caveats:
- Some TES systems have unique requirements (parameters)
	- How does this work in a general abstracted container/class? A common function (e.g., init()) having the unique implementation elements?
- Select parameters depend on other sub-component initialization routines
	- However, I don't think the current subsystems need to depend on other initialization routines.
	- If needed in the future, have a new resize() function in the TES class
		- called by a new integrate_subcomponents() function in the controller
		- resize() takes and returns the same structure of variables
		- the variables are conditions at its boundary with other components (e.g., T_src_in, T_sink_in, mdot_src)
		- the return structure are the other parameters not specified. This facilitates design convergence.
		- all parameters in structure are optional
		- the function does child-class specific functions depending on the parameters passed
 
Conclusions:
- Integration into a system should not be a required step during TES creation
- It should not be required to call a specific, unknown function (e.g., init()) in order to create a valid TES or any other object

Questions:
1. Should the collector receiver and power cycle be refactored in the same way done here for the TES, and add similar tests?
2. Should a calc_design_vals() function be added to the base class (where size_tes_piping_TandP() will be called)?
	- It's the responsibility of the solver to call size_tes_piping_TandP() because it requires info from a different subcomponent routine
	- Maybe add a resize() function later if needed for integration (also not an 'init' functionality)

Note:
size_tes_piping_TandP():
- only called if calc_design_pipe_vals is true
- does not change the pipe diameters
- only calculates the pipe temperatures and pressures at design


tyneises (changes requested):
Importing each design parameter individually to a component constructor requires the user to provide that value. Importing the parameters through a single structure forces something downstream in the class to check if a design parameter is undefined. Then the design parameters would exist as protected member data rather than a structure. In the MSPT code I'm also doing unit conversions when setting the member variable to argument variable. I suppose this is also possible with member structures, but it would probably be a bit more confusing describing the units in the variable comment (i.e. "input" unit versus "converted" unit).

It's subjective, but I prefer to separate constructing the class (and importing design parameters) from performing design point calculations (which currently happens in the init(...)) method. I imagined there might be a case where the init() method of one component might inform design parameters for another component. I think there's one example of that in the csp solver class.

Small point, but I'd prefer the TES to require Q_tes as a design parameter rather than tshours.

I didn't look closely at the pressure drop changes, but generally organizing HTF pumping power calcs into their respective component classes would make sense.

tyneises:
Your comments on the pipe sizing and TES design iteration are interesting. Let's talk more about those.

Matthew-Boyd:
Some potentially better alternatives for passing in parameters ([link](https://stackoverflow.com/questions/40264/how-many-constructor-arguments-is-too-many/40324#40324))
Validating input parameters ([link](https://stackoverflow.com/questions/53659588/how-to-validate-input-parameters-in-c-constructor))
How to handle a constructor that fails ([link](https://isocpp.org/wiki/faq/exceptions#ctors-can-throw))

Matthew-Boyd (commented):
@tyneises I think these address your requested changes. Please review and approve if you agree.

tyneises (approved):

PR #864, opened 2022-08-22
PVWatts albedo soiling checks and outputs

cpaulgilman:
Add albedo and soiling to time series results to facilitate troubleshooting.

Fix albedo logic to correctly assign default 0.2 or 0.6 value depending on weather file albedo validity, snow model enabled, snow depth data.

Goes with https://github.com/NREL/SAM/pull/1136

Weather files for testing (snow/albedo, no albedo, mix of valid and invalid albedo):
[albedo-test-wf.zip](https://github.com/NREL/ssc/files/9397998/albedo-test-wf.zip)


ssc/cmod_pvwattsv8.cpp
```diff
@@ -826,12 +823,28 @@ class cm_pvwattsv8 : public compute_module
                 else if (is_assigned("albedo"))
                     log(util::format("Albedo array was assigned but is not the correct length (1, 12, or %d entries). Using default value.", nrec), SSC_WARNING);
 
-                // if the user hasn't specified an albedo, or has specified to use wf albedo, and the weather file contains hourly albedo, use that instead
-                // but make sure that the weather file albedo is realistic
-                // albedo_len will be zero if the albedo input isn't assigned
-                if (std::isfinite(wf.alb) && wf.alb > 0 && wf.alb < 1 && (albedo_len == 0 || use_wf_albedo))
+                // check albedo is valid value between 0 and 1. if not, set to default value 0.2
+                // if the snow loss model is enabled and there's valid snow depth > 0.5 cm, then set to 0.6
+                if (alb <= 0 || alb >= 1)
```
janinefreeman, line 828: If albedo_len is wrong, pvwatts will never use the snow value because the albedo default set in line 816 won't trigger this if statement. Suggest replacing line 816 with the contents of this if statement.

ssc/cmod_pvwattsv8.cpp
```diff
@@ -826,12 +823,28 @@ class cm_pvwattsv8 : public compute_module
...
+                {
+                    if (std::isfinite(wf.snow) && wf.snow > 0.5 && wf.snow < 999 && en_snowloss)
+                        alb = as_double("albedo_default_snow");
+                    else
+                        alb = as_double("albedo_default");
+                }
+
+                // conditions to use albedo from wf instead of albedo user input:
+                //   * wf.alb exists and is valid and use_wf_albedo = 1
+                //   * wf.alb exists and is valid and use_wf_albedo = 0 and albedo_len = 0
```
janinefreeman, line 838: Is this the desired behavior? You can remind me if we discussed at a meeting and I forgot, but I would think that if use_wf_albedo = 0 and user-specified albedo is not defined, I would be expecting it to use internal defaults (or the defaults that I set via the SDK), even if albedo exists in the weather file.

janinefreeman (dismissed):
Having trouble compiling SAM at the moment, but have a few comments based on just reading the code. Will work on getting SAM building so I can test after you have a chance to look at these comments. Thanks again for putting this PR together!!!

ssc/cmod_pvwattsv8.cpp
cpaulgilman: Fixed in 22acecb6

ssc/cmod_pvwattsv8.cpp
cpaulgilman: I agree with the behavior you describe. Fixed in 22acecb6.

cpaulgilman: Based on discussions at the 9/20/2022 SAM meeting, we agreed to simplify the albedo logic:

If use_wf_albedo is true, and wf albedo is available and valid, use wf albedo, otherwise use default albedo (snow or no snow, depending on snow depth and snow model enabled).

If use_wf_albedo is false, and user albedo is available and valid, use user albedo, otherwise use default albedo as above.

sjanzou (approved):
Nice test addition and I could not break the logic!

PR #867, opened 2022-08-25
138 Financial Model Tests added

sjanzou:


brtietz (approved):
Great work! Out of curiosity, why did the utility rate values and gen just now get added to host developer?

sjanzou:
> Great work! Out of curiosity, why did the utility rate values and gen just now get added to host developer?

The inputs were added to the host developer compute module so that the precheck would pass with no missing required inputs when the cmod_hostdeveloper would run standalone without a technology and utility rate compute module being run beforehand...

Issue #870, opened 2022-09-15
pvwattsv8 inverter low efficiency bug


sjanzou:
pvwatts v8 inverter low efficiency values incorrect per comparison with pvwatts v5 and detailed pv at low efficiencies:

![image](https://user-images.githubusercontent.com/6498311/190346142-0213c938-7cc5-4c99-843e-408f768077d5.png)

sjanzou:
developing a fix in ssc branch ssc_870

PR #872, opened 2022-09-20
Address ssc issue 870

sjanzou:


sjanzou:
Tests added and updated for pvwattsv8, summary document attached along with reference.
[ssc_870.zip](https://github.com/NREL/ssc/files/9606935/ssc_870.zip)

test/ssc_test/cmod_json_comparison_test.h
```diff
@@ -58,11 +58,15 @@ class JSONComparisonTest : public ::testing::Test {
 
 public:
     void Test(const std::string& compute_module, const std::string& file_inputs, const std::string &file_outputs, const std::vector<std::string> &compare_number_variables, const std::vector<std::string> &compare_array_variables) {
+        char solar_resource_path[256];
```
brtietz, line 61: Looks like the error tolerance on lines 92 and 101 is too tight for the linux vs mac differences. Is a blanked increase to 0.01 ok, or do we need to do this on a percentage basis?

brtietz (changes requested):
Looks like this fixes the problem- having the inverter efficiency as a time series output is very helpful. Just need to update some tolerances to the mac tests pass as well.

test/ssc_test/cmod_json_comparison_test.h
sjanzou: tolerance updated for macOS tests. All tests now pass
image

brtietz (approved):

cpaulgilman:
https://github.com/NREL/ssc/issues/870

PR #874, opened 2022-09-26
fix SAM issue 1137

sjanzou:
DC power loss due to snow (%) output changed from fraction to percentage and tests updated


cpaulgilman:
https://github.com/NREL/SAM/issues/1137

cpaulgilman (approved):
PVWatts with snow results now look reasonable.

PR #875, opened 2022-09-26
Fix number of wells flow rate calculations for binary plants

mjprilliman:
-Closes https://github.com/NREL/SAM/issues/1156
-Allows for specifying binary plant as # of wells even though GETEM spreadsheet does not calculate total flow rate as function of # wells


cpaulgilman (approved):
I verify that changing the number of wells now changes the plant output on the Plant and Equipment page for both Binary and Flash plant types.

PR #879, opened 2022-09-29
add ts_hour to ac wiring loss line

brtietz:
To test:

- Run a subhourly simulation
- Confirm that the AC wiring losses match the input (1% is default)

Unchecking the "save all output variables over lifetime" option is recommended if testing with a one minute weather file.

Fixes https://github.com/NREL/SAM/issues/1082


brtietz:
Push build on ubuntu failed due to error connecting to cmake. We can either re-run the tests or trust the results of the mac push and pull request builds.

sjanzou:
> Push build on ubuntu failed due to error connecting to cmake. We can either re-run the tests or trust the results of the mac push and pull request builds.

All tests seem to be passing now!

sjanzou (approved):
Much better for subhourly models! AC wiring loss no longer multiplies with ts_hour!
All tests now passing.

PR #880, opened 2022-09-29
add capability for AC connected battery to charge from fuel cell

brtietz:
Add fuel cell output power to AC outage calculations to fix https://github.com/NREL/SAM/issues/1130

File with Will's test instructions (see issue for details): 
[sam_1130_test_case.zip](https://github.com/NREL/ssc/files/9679325/sam_1130_test_case.zip)

Haven't done DC because there are no SAM configurations with batteries that are DC connected to PV with fuel cells.


mjprilliman (approved):
Looks good. 

Consider adding autonomy autographs for this Fuel Cell configuration: 
image

PR #881, opened 2022-09-30
Fix battery dispatch for retail rate electricity purchases.

### LINKED ISSUE #824: Battery automated dispatch is unaware of retail prices
**Describe the bug**
The automated dispatch method for the battery is supposed to compare the price of charging, revenue to discharge, round trip efficiency, and cost of cycling degradation in order to make automatic dispatch decisions. However, the price of charging given to the algorithm is always based on the market price, whereas the front of meter financial models have the option to draw power from the grid at a (potentially fixed) retail rate.

**To Reproduce**
1. Run the attached SAM file: 
[aligned price series.zip](https://github.com/NREL/ssc/files/8789667/aligned.price.series.zip)
2. Note that battery dispatch is the same in all cases, despite higher costs of electricity when the retail rate is turned on.

**Expected behavior**
Retail rates should be provided to automated dispatch where relevant.

brtietz:
move en_electricity_rates to utility_rates_common, so it's assigned for retail rate dispatch.

Fixes #824. Test file available with that issue. The cases with retail rates should show reduced charging. Increasing the retail rate to $1/kWh should lead to the battery only charging once during the analysis period (when the price goes above $1100/MWh in June).

### DIFF
--- a/ssc/cmod_utilityrate5.cpp
+++ b/ssc/cmod_utilityrate5.cpp
@@ -31,7 +31,6 @@ OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 static var_info vtab_utility_rate5[] = {
 
 /*   VARTYPE           DATATYPE         NAME                         LABEL                                           UNITS     META                      GROUP          REQUIRED_IF                 CONSTRAINTS                      UI_HINTS*/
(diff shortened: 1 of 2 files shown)


sjanzou (approved):
Interesting how the market charge SOC percentage changed for the the merchant plant and not single owner with the change. The retail rates now exhibit the desired behavior per the attached word document using the test file from issue 824
[Ssc pull request 881 for ssc issue 824.docx](https://github.com/NREL/ssc/files/9722485/Ssc.pull.request.881.for.ssc.issue.824.docx)

PR #882, opened 2022-09-30
fix index bug for price signals dispatch

### LINKED ISSUE #843: Monthly TOU periods don't work for price signals forecast
**Describe the bug**
The attached SAM file throws an error because price signals dispatch has trouble interpreting the utility rate, with the absence of period 1 in month 5. However the utility rates code has no trouble with this rate.

**To Reproduce**
1. Run the attached SAM file 
[Hundt.demandraterror.zip](https://github.com/NREL/ssc/files/9023694/Hundt.demandraterror.zip)

2. See error

**Expected behavior**
Price signals dispatch should support all rates supported by the utility rates code.

brtietz:
Correct indexing error in price signals dispatch setup code, which caused error in utility rates where the months utilize different periods. The error only affected price signals dispatch, the utility rates module already properly supported the rate.

Fixes #843. Test file in that issue should run without exceptions.


cpaulgilman (approved):
Looks good. Test file runs with reasonable results.

PR #883, opened 2022-09-30
Align batt_to_grid with MP market revenue for LCOS calculations

mjprilliman:
-Use extrapolate_timeseries function in similar way to merchant plant calculations to align market revenue with battery to grid timeseries arrays
-Closes https://github.com/NREL/SAM/issues/1147

Todo: 
-Add test to evaluate merchant plant revenue size - gen length behavior for future changes.


cpaulgilman (approved):
The test file now runs without crashing. The 2-yr cash flow is not viable, so I can't really evaluate whether the LCOE and LCOS results make sense. The 25-yr case takes a very long time to run at 1-minute resolution, I will follow up if I have anything else to add after it finishes.

mjprilliman:
Will probably need to revisit how to write proper tests for this but through debugging the example file: 

-Hourly grid_to_batt and 5 minute market revenue: stretch the hourly values to 5 minute data, then apply to the market revenue
-1 minute grid_to_batt data and 5 minute market revenue: take the grid_to_batt values at every 5 minutes in grid_to_batt array to make a new array, apply to market revenue

I'm not sure the second example is the ideal way to calculate battery charging costs but I'm also not sure of a better implementation before the release. Thoughts @brtietz @cpaulgilman ? If there are no objections I will merge.

brtietz:
Those approaches both sound valid to me. If you merge this, please make a subsequent issue to add the tests to ssc.

mjprilliman:
Added a simple ssc test using Steve's test generator to at least know if the behavior changes. Testing in more detail requires more outputs of the extrapolated batt_to_grid array, is this something we want to add? (would have to specify it only applies to LCOS)

cpaulgilman:
> I'm not sure the second example is the ideal way to calculate battery charging costs but I'm also not sure of a better implementation before the release. Thoughts @brtietz @cpaulgilman ? If there are no objections I will merge.

I will defer to others on this -- no objections from me.

PR #887, opened 2022-10-11
PV albedo fixes

cpaulgilman:
Goes with https://github.com/NREL/SAM/pull/1180

Clean up albedo variable labels.

Fix var table issue that was causing API file errors.


sjanzou (approved):
Labelling seems much clearer for a confusing multiple entry!

PR #889, opened 2022-10-11
Improve labeling of loss percentage inputs for AC/DC

cpaulgilman:
Goes with https://github.com/NREL/SAM/pull/1183


ssc/cmod_pvsamv1.cpp
```diff
@@ -124,8 +124,8 @@ static var_info _cm_vtab_pvsamv1[] = {
...
+        {SSC_INPUT, SSC_NUMBER,   "acwiring_loss",                        "AC wiring loss",                                      "%",      "percent of inverter AC output",                                                                                                                                                                                      "Losses",                                                "*",                                  "MIN=0,MAX=100",       "" },
+        {SSC_INPUT, SSC_NUMBER,   "transmission_loss",                    "Transmission loss",                                   "%",      "percent of AC output",                                                                                                                                                                                      "Losses",                                                "*",                                  "MIN=0,MAX=100",       "" },
```
brtietz, line 128: Is it worth indicating that the transmission loss is applied after transformer losses?

brtietz (commented):
This is definitely an improvement, but I'm not sure about the transformer_load_loss and the transmission_loss getting the same label. They are applied sequentially (transformer before transmission).

ssc/cmod_pvsamv1.cpp
cpaulgilman: Yes, thanks for that suggestion.

PR #892, opened 2022-10-14
Fix buffer overflow and remove redundant albedo checks

Matthew-Boyd:
The monthly albedo check caused a buffer overflow from an array out-of-bounds error when spatial monthly albedos are used instead of monthly uniform albedos. The other checks are redundant to the checks on lines 185-194 in lib_pv_io_manager.cpp .

Paul, please let me know if I'm misunderstanding anything. Thanks.


cpaulgilman (approved):
Thanks for catching and fixing that.

Matthew-Boyd:
No problem! Thanks for reviewing.

PR #897, opened 2022-10-19
PV temperature updates in ssc

mjprilliman:
-Option to input module temperatures instead of using temperature models (length nrecs per year)
-Option to use lacunarity heat transfer model (University of Utah, Portland State University, NREL) when using the heat transfer model option
-SAM user interface updates to come in patch or future release
-Partially addresses https://github.com/NREL/SAM/issues/106 (on SDK/PySAM side for now)


cpaulgilman (approved):
I did a basic pvsamv1 test with the CEC module model in SDKtool to confirm that setting time series cell temp as input does affect the cell temp output, and to confirm that this option is not accessible from the SAM UI, even via LK.

I thought `subarray1_celltemp` output would be equal to the `measured_temp_array` input, but it isn't -- not sure what to expect.

[pv-cell-temp-as-input.zip](https://github.com/NREL/ssc/files/9825023/pv-cell-temp-as-input.zip)

Issue #784, opened 2022-03-25
Electricity to/from battery is in AC sometimes and DC others


brtietz:
**Describe the bug**
The numbers reported in the "electricity to/from battery" output are sometimes in DC numbers (a DC connected battery charging) and sometimes in AC numbers (AC battery, DC battery discharging)

**To Reproduce**
1. Run a DC connected battery case
2. Compare PV DC gross numbers to battery charging
3. Compare to grid and AC output numbers to battery discharging

**Expected behavior**
Numbers in this output field should be more consistent and easily explainable to users.

Leaving this issue without a milestone while we define conventions.

brtietz:
Decision on 6/16 meeting to add additional outputs and label those that are DC (and one AC number):

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/bmirletz/AppData/Local/Temp/1/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/bmirletz/AppData/Local/Temp/1/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">



Variable Name | AC, Charging | AC, Discharging | DC, charging | DC, discharging
-- | -- | -- | -- | --
Electricity to battery from system (DC) | DC | N/A | DC | N/A
Electricity to/from battery (AC) | AC | AC | AC | AC
Electricity to/from battery (DC) | DC | DC | DC | DC
Electricity to inverter DC from battery | N/A | N/A | DC | DC



</body>

</html>

•	Electricity to battery from system (DC) -> Existing output, change to be consistently DC
•	Electricity to/from battery (AC) -> Same as now for AC connected, be consistent about conversion efficiencies for DC connected
•	Electricity to/from battery (DC) -> Sort of a new output. Report in all cases even if we’re just multiplying by a flat efficiency
•	Electricity to inverter DC from battery -> New output. Sums with PV output to make Inverter DC input power

This achieves consistency within each variable between AC and DC connected systems.

brtietz:
[...]

PR #899, opened 2022-10-23
784 fix ac dc powers

brtietz:
Define additional variables and conventions in the powerflow variables according to https://github.com/NREL/ssc/issues/784. Closes https://github.com/NREL/ssc/issues/784

Conventions for DC to AC conversion for AC connected batteries and DC connected batteries while discharging are preserved. DC connected batteries charging from PV needed new a convention, which is:

Convert to AC power by multiplying by the inverter max efficiency.

This PR also updates some of the losses conventions to ensure "electricity from grid to system" matches the inverter AC "output" when charging. 

Alternative conventions considered included "what would it take to charge from the grid (divide by inverter efficiency)" and "use actual inverter efficiency". Some notes on what didn't work:
* We need AC variables for LCOS to make sense (Cannot report DC only)
* Both "electricity to/from grid" and "from system to battery" need to use the same convention, otherwise "percent charge from system" doesn't work
*  The grid charging convention is problematic because it makes the DC connected battery look less efficient than the AC connected battery
* Using the actual inverter efficiency could push the round-trip efficiency even higher than the current convention

Additional changes:
- Use night time losses when deciding whether or not the inverter is running in the powerflow code, rather than assuming a 5% efficiency
- Modifications to dispatch code to be more consistent about AC vs DC power in the constraints
- Solve a few compiler warnings

Relevant test files:
Ensure grid charging and PV charging numbers add up for a DC connected battery: 
[dc_charging_grid_and_pv.zip](https://github.com/NREL/ssc/files/9847457/dc_charging_grid_and_pv.zip)
Custom dispatch from pvsamv1_battery integration tests that produces low inverter efficiencies: 
[low_power_inverter_test_cases.zip](https://github.com/NREL/ssc/files/9847458/low_power_inverter_test_cases.zip)

Looking for at least two approvals prior to merge.

I expect there will be some changes to the SAM_api and export_config files, but those builds are broken on develop at to time of writing (the relevant person has been emailed). I'll make a SAM PR for that when they're fixed.


cpaulgilman:
@brtietz Thanks for going through this, and for the clear descriptions in your comment above.

The relationship between the DC and AC versions of "Electricity to battery from system" is the inverter's peak nominal efficiency, For a DC-connected battery, the calculated AC value is an approximation for LCOS calculations and not a power quantity that you would measure in a real system. Is that correct?

cpaulgilman (approved):

brtietz:
> @brtietz Thanks for going through this, and for the clear descriptions in your comment above.
> 
> The relationship between the DC and AC versions of "Electricity to battery from system" is the inverter's peak nominal efficiency, For a DC-connected battery, the calculated AC value is an approximation for LCOS calculations and not a power quantity that you would measure in a real system. Is that correct?

For PV charging yes. For grid charging, it is the power draw on the AC side of the inverter.

shared/lib_battery_dispatch.cpp
```diff
@@ -174,7 +174,7 @@ bool dispatch_t::check_constraints(double& I, size_t count)
     // Don't allow grid charging unless explicitly allowed (reduce charging)
     if (!m_batteryPower->canGridCharge && I < 0 && m_batteryPower->powerGridToBattery > tolerance)
     {
-        m_batteryPower->powerBatteryTarget += m_batteryPower->powerGridToBattery;
+        m_batteryPower->powerBatteryTarget += m_batteryPower->powerGridToBattery * m_batteryPower->singlePointEfficiencyACToDC;
```
sjanzou, line 177: Are we going to handle different efficiencies in a future issue or pull request?

sjanzou (approved):
Lots of clarifying outputs and nice use of the code-attic!

shared/lib_battery_dispatch.cpp
brtietz: Yes, if we do some of the DC-DC converter work we talked about with the PVSAMV1 refactor. Some of the DC-DC conversions are currently handled early in the powerflow code, the current setup should work as long as we keep using single point efficiencies.

PR #901, opened 2022-10-24
Updates all compute modules and tests to handle ITC array inputs for SAM 1038

sjanzou:
This is first step to full implementation. Current implementation uses single element arrays input to match previous test. Next step is full ITC implementation over schedule of years.


brtietz (approved):
Thanks for navigating some complex changes!

cpaulgilman (approved):
For the full implementation, I think this will require new `cf_itc_sta` and `cf_itc_fed` variables for consistency with other cash flow outputs.

sjanzou:
> For the full implementation, I think this will require new `cf_itc_sta` and `cf_itc_fed` variables for consistency with other cash flow outputs.

Agreed!

PR #903, opened 2022-10-25
SAM issue 1194 - add output metric checking

sjanzou:
Goes with SAM pull request 1209


cpaulgilman (approved):
Looks good. Financial metric messages working for different financial models.

Issue #906, opened 2022-10-26
Change array inputs that are specified as single number to single value array inputs in ssc precheck


sjanzou:
Per 10/25/2022 meeting and SAM issue https://github.com/NREL/SAM/issues/1038

Change all ssc array inputs that are passed into ssc as single number to a single value array and log message.

brtietz:
In addition to the SSC changes proposed here, we will need to update PySAM, the function here fails when provided a number https://github.com/NREL/pysam/blob/master/files/PySSC.py#L70

Generally all of these functions could use a try-except which gives the name of the variable when there is an error, but that's a separate issue.

PR #907, opened 2022-10-26
Change single number inputs to single element array inputs for SSC_ARRAY required input variables

sjanzou:
works with previous ssc test files and SDKtool with lk generated from previous releases

### DIFF
--- a/ssc/core.cpp
+++ b/ssc/core.cpp
@@ -277,9 +277,17 @@ bool compute_module::verify(const std::string &phase, int check_var_type) {
                         ") required but not assigned");
                     return false;
                 } else if (dat->type != vi->data_type) {
-                    log(phase + ": variable '" + std::string(vi->name) + "' (" + var_data::type_name(dat->type) +
-                        ") of wrong type, " + var_data::type_name(vi->data_type) + " required.");
-                    return false;
+                    // ssc issue 906 - required only inputs!
+                    if ((vi->data_type == SSC_ARRAY) && (dat->type == SSC_NUMBER)) {
+                        log(phase + ": variable '" + std::string(vi->name) + "' input type changed from " + var_data::type_name(dat->type) +
+                            " to single element " + var_data::type_name(vi->data_type) + ".");
+                        dat->type = SSC_ARRAY;
+                    }
+                    else { 
+                        log(phase + ": variable '" + std::string(vi->name) + "' (" + var_data::type_name(dat->type) +
+                            ") of wrong type, " + var_data::type_name(vi->data_type) + " required.");
+                        return false;
+                    }
                 }
 
                 // now check constraints on it


tyneises (commented):
Is it correct that the outcome of this change is that if the cmod variable type is ARRAY but the input variable type is NUMBER, then the new code will convert the number to an array?

If so, I understand that this solves the current issue with the financial model inputs. However, this could make "real" variable definition errors harder to catch/fix. For example, if I incorrectly set an ARRAY as a NUMBER, instead of the pre-checks catching it, I won't know it's an error until whatever code is using that array finds that it isn't the expected length. Oftentimes the error reporting in these scenarios is hard to follow until you step through the code in debug. My opinion (if I understand this change correctly) is that we need a solution specific to the problem of a variable name/definition changing, but I could live with this fix.

janinefreeman:
@tyneises that is a correct understanding. We also discussed updating the precheck routine to issue a warning when this action is triggered, @sjanzou is that in this PR as well or will that be separate? (I understand that warnings in scripts are not as useful as they are in the UI, but that's perhaps a separate conversation). However, we discussed that this change will still trigger any var_table prechecks where the array length is checked, even if the type gets changed, and it obviously also will trigger any exec-level warnings about array length (which should be checked anywhere an array is used).

That said, this is definitely not a perfect solution- it was the best balance between crashing every financial script ever created, giving the user some information, and ease of development given the imminence of the release, that we could think of, and we all agreed a more elegant solution is in order after the release.

cpaulgilman:
> We also discussed updating the precheck routine to issue a warning when this action is triggered, @sjanzou is that in this PR as well or will that be separate? (I understand that warnings in scripts are not as useful as they are in the UI, but that's perhaps a separate conversation).

This PR does include the warnings. For example setting `itc_fed_amount = 300` results in `Notice: precheck input: variable 'itc_fed_amount' input type changed from <number> to single element <array>. time -1.000000`

These simulation warnings are not always obvious in PySAM.

cpaulgilman:
Tested and working in SDKtool: [number-to-array-tests-sdktool.zip](https://github.com/NREL/ssc/files/9871154/number-to-array-tests-sdktool.zip)

This solution does not work for LK in SAM. When a script sets an input to a value of the wrong type, the value of the input does not change and you don't get any feedback: For example, try this script with a PVWatts - Single Owner case:  [number-to-array-test.zip](https://github.com/NREL/ssc/files/9871168/number-to-array-test.zip)

brtietz:
Code looks good to me.

PySAM tests pass (except the REopt test, which hung and had to be cancelled - I think that's unrelated), however PySAM examples fail, and changes to PySSC are required as per: https://github.com/NREL/ssc/issues/906#issuecomment-1292252082 - I'll work on that this afternoon.

sjanzou:
> @tyneises that is a correct understanding. We also discussed updating the precheck routine to issue a warning when this action is triggered, @sjanzou is that in this PR as well or will that be separate? (I understand that warnings in scripts are not as useful as they are in the UI, but that's perhaps a separate conversation). However, we discussed that this change will still trigger any var_table prechecks where the array length is checked, even if the type gets changed, and it obviously also will trigger any exec-level errors about array length (which should be checked anywhere an array is used).
> 
> That said, this is definitely not a perfect solution- it was the best balance between crashing every financial script ever created, giving the user some information, and ease of development given the imminence of the release, that we could think of, and we all agreed a more elegant solution is in order after the release.

@tyneises , @janinefreeman , @cpaulgilman , this pull request is for ssc only and for pysam and scripting against ssc only. The conversion in SAM scripting and updating old project files will be a separate pull request. Thanks for reviewing!

tyneises (approved):

PR #908, opened 2022-10-26
Add weather file header data to outputs Fix #453

### LINKED ISSUE #453: Add time zone as output to solar and wind models
This issue was originally posted for PySAM at https://github.com/NREL/pysam/issues/54:

> The PVWatts7 module includes a [time zone output](https://nrel-pysam.readthedocs.io/en/master/modules/Pvwattsv7.html?highlight=pvwatts#PySAM.Pvwattsv7.Pvwattsv7.Outputs.tz). However, this output does not seem to exist for any of the other modules, including the windpower modules and concentrating solar power modules.
> 
> Especially when working with resources across multiple time zones, it would be nice to have this output so that timeseries outputs can be easily converted to a single time zone.

cpaulgilman:
Add latitude, longitude, elevation, time zone to outputs for pvsamv1.

Change labels and units for lat, lon, elev, tz outputs for pvwattsv8 to be consistent with pvsamv1.

Add latitude, longitude, elevation, year to outputs for windpower. Note that windpower does not read time zone from weather file, so not reporting as an output. Also lat, lon, year do not appear to be used in calculations, but are read in, so I decided to include them in the outputs for completeness.

Use "Location" group name for all.

This PR does not add outputs to CSP models.


ssc/cmod_pvwattsv8.cpp
```diff
@@ -199,9 +199,9 @@ static var_info _cm_vtab_pvwattsv8[] = {
...
-        { SSC_OUTPUT,       SSC_NUMBER,      "lat",                            "Latitude",                                    "deg",       "",                                             "Location",      "*",                       "",                          "" },
-        { SSC_OUTPUT,       SSC_NUMBER,      "lon",                            "Longitude",                                   "deg",       "",                                             "Location",      "*",                       "",                          "" },
-        { SSC_OUTPUT,       SSC_NUMBER,      "tz",                             "Time zone",                                   "hr",        "",                                             "Location",      "*",                       "",                          "" },
+        { SSC_OUTPUT,       SSC_NUMBER,      "lat",                            "Latitude",                                    "DD",       "",                                             "Location",      "*",                       "",                          "" },
```
mjprilliman, line 202: What's the reason for specifying "from weather file" in pvsamv1 but not in pvwattsv8 and windpower (referring to output label).

ssc/cmod_pvsamv1.cpp
```diff
@@ -539,6 +539,12 @@ static var_info _cm_vtab_pvsamv1[] = {
         { SSC_OUTPUT,        SSC_ARRAY,      "df",                                         "Irradiance DHI from weather file",                                     "W/m2",   "",                      "Time Series",       "*",                    "",                              "" },
         { SSC_OUTPUT,        SSC_ARRAY,      "wfpoa",                                      "Irradiance POA from weather file",                                     "W/m2",   "",                      "Time Series",       "",                     "",                              "" },
 
+        // header data from weather file
+        { SSC_OUTPUT,        SSC_NUMBER,      "lat",                                         "Latitude from weather file",                                          "DD",   "",                      "Location",       "",                    "",                              "" },
```
mjprilliman, line 543: Is DD understood by enough of the modeling community to use as a abbreviated unit (probably yes)

ssc/cmod_windpower.cpp
```diff
@@ -616,6 +622,11 @@ void cm_windpower::exec()
 	assign("cutoff_losses", var_data((ssc_number_t)((withoutCutOffLosses - annual) / withoutCutOffLosses)));
 	assign("annual_gross_energy", annual_gross);
 
+    assign("lat", wdprov->lat);
+    assign("lon", wdprov->lon);
+    assign("elev", wdprov->elev);
+    assign("year", wdprov->year);
```
mjprilliman, line 628: The default wind - single owner year value is 1900. Will that be confusing to people?

mjprilliman (approved):
Looks good just some questions about user understanding

ssc/cmod_pvwattsv8.cpp
cpaulgilman: pvsamv1 uses "from weather file" for some outputs because of the POA as input solar resource data option. I decided to use it for these variables as well because they are passed through from the weather file and not really simulation outputs. I did not use "from weather file" for pvwattsv8 or windpower because no other variables in those cmods use it.

ssc/cmod_pvsamv1.cpp
cpaulgilman: I'm not sure how well known it is among the solar modeling community, but I've been using it based on information like this [Wikipedia article](https://en.wikipedia.org/wiki/Decimal_degrees) since there has been confusion in the past about geographic coordinates in SAM, such as using both a negative sign and a "W" for longitude values for locations in the United States. I can change the units to "deg" or "degrees" if that seems clearer than "DD". (The Location and Resource input page uses "DD".)

mjprilliman: I didn't notice that we have been using DD, I would say keep as you've implemented it.

ssc/cmod_windpower.cpp
cpaulgilman: The default cases use "representative files" that have no year value in the header, hence the default 1900. When you download a wind resource file from the WIND Toolkit, the data has a more meaningful year value, which I think might be useful to some users to identify the data (especially SDK/PySAM).

I'm not sure why windpower reads the year value from the wind resource file because it does not seem to be used anywhere in the code. Note that it is required in [Line 111](https://github.com/NREL/ssc/blob/e0297ed5166a90bd46202097ceeedf11ca9b974f/ssc/cmod_windpower.cpp#L111).

ssc/cmod_pvsamv1.cpp
cpaulgilman: Hm. there is this PV Terms project that uses degrees for latitude longitude: https://duramat.github.io/pv-terms/

> latitude [degrees]: Latitude between -90 and 90 degrees. Positive north of the equator.
> longitude [degrees]: Longitude between -180 to 180 degrees. Positive east of the meridian.

PR #912, opened 2022-10-27
Fix annual total loss percent output value

### LINKED ISSUE #911: PV annual total loss percent output metric is inaccurate
The annual_total_loss_percent, or "Total loss from nominal POA to net AC" calculation in the detailed PV model is inaccurately accounting for gain and bifacial terms.

Matthew-Boyd:
This fixes the calculation for both gain terms and bifacial terms and this issue doesn't need to be revisited after the release.

Closes #911

### DIFF
--- a/ssc/cmod_pvsamv1.cpp
+++ b/ssc/cmod_pvsamv1.cpp
@@ -915,8 +915,7 @@ static var_info _cm_vtab_pvsamv1[] = {
             
                 { SSC_OUTPUT, SSC_NUMBER, "annual_dc_optimizer_loss", "DC power optimizer loss", "kWh", "", "Annual (Year 1)", "", "", "" },
 
-                // total loss diagram losses for single year, does not include lifetime losses
-                { SSC_OUTPUT, SSC_NUMBER, "annual_total_loss_percent", "Total loss from nominal POA to net AC", "kWh", "", "Annual (Year 1)", "", "", "" },
+                { SSC_OUTPUT, SSC_NUMBER, "annual_total_loss_percent", "Total loss from nominal POA to net AC", "%", "", "Annual (Year 1)", "", "", "" },
 
                 //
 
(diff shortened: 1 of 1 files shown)


ssc/cmod_pvsamv1.cpp
```diff
@@ -3133,28 +3132,7 @@ void cm_pvsamv1::exec()
...
-                                                 "annual_ac_inv_clip_loss_percent", "annual_ac_inv_pso_loss_percent",
-                                                 "annual_ac_inv_pnt_loss_percent","annual_ac_inv_eff_loss_percent",
-                                                 "annual_ac_wiring_loss_percent", "annual_xfmr_loss_percent",
-                                                 "annual_transmission_loss_percent", "annual_ac_perf_adj_loss_percent"};
-        percent = 1.;
-        for (size_t i = 0; i < loss_components.size(); i++) {
-            percent *= (1. - as_number(loss_components[i]) / 100.);
-        }
-        assign("annual_total_loss_percent", var_data((ssc_number_t)(1. - percent) * 100.));
+        assign("annual_total_loss_percent", var_data((ssc_number_t)((1. - annual_energy / as_number("annual_poa_eff")) * 100.)));
```
brtietz, line 3135: I'm seeing a 84.4% loss with the detailed-PV single owner defaults on this branch, vs 22.7% in 2021.12.02r2. I wonder if you could use DC nominal as the denominator, and add in some of the shading and soiling losses on top of that?

brtietz (changes requested):
Given the downstream use of this in the REopt code, I think the number here is higher than intended. Is there another denominator we can use?

ssc/cmod_pvsamv1.cpp
Matthew-Boyd: What's the expected definition of this term by REopt, assuming it's not "Total loss from nominal POA to net AC"? Was this term originally added for REopt?

Just for background, the 22.7% was inaccurate due to how the bifacial gain term was added or from some other issue with how the individual loss factors were chained together in the prior calculation. Most of the loss is just PV conversion; at STC for the default module having a 19.3% efficiency, this is a 80.7% loss.

brtietz (approved):
Code matches conventions defined in 11/1 SAM meeting. Compatibility with REopt will be handled in https://github.com/NREL/ssc/issues/923

Thanks for fixing and catching the units typo!

Matthew-Boyd:
Thanks for the help Brian!

PR #925, opened 2022-11-02
Ssc 923 use prod factors

### LINKED ISSUE #923: Convert size battery and dispatch REopt call to use prod_factor_series_kw
In the SAM team meeting on 11/1, we determined the best way to resolve the losses issue in https://github.com/NREL/ssc/pull/912 and https://github.com/NREL/SAM/pull/1165 is to convert the REopt call to use prod_factor_series_kw. This allows us to set the losses to zero, so we can define whatever convention is needed for SAM users for the annual_total_loss_percent variable. prod_factor_series_kw also allows more accurate battery dispatch. The only reason it isn't being used currently is it did not exist when the code was written originally.

brtietz:
Use production factors from the actual system instead of lat/lon when getting size and dispatch from REopt. Additional required changes in SAM PR: https://github.com/NREL/SAM/pull/1165 Fixes #923 

Steps to test:
1. Run the attached SAM and weather files in 2021.12.02: 
[2018_weather_file_dispatch.zip](https://github.com/NREL/ssc/files/9923585/2018_weather_file_dispatch.zip)
2. Build SAM from this branch and the SAM branch sam_1092_update_reopt_script
3. See improved sizing and dispatch (via improved NPV) on branches

### DIFF
--- a/ssc/cmod_pvsamv1_eqns.cpp
+++ b/ssc/cmod_pvsamv1_eqns.cpp
@@ -157,7 +157,7 @@ SSCEXPORT bool Reopt_size_battery_params(ssc_data_t data) {
     std::vector<double> kwac_per_kwdc(year_one_values);
     for (size_t i = 0; i < year_one_values; i++) {
         kwac_per_kwdc[i] = gen[i] / system_cap;
-        kwac_per_kwdc[i] = std::fmax(0.0, kwac_per_kwdc[i]);
+        kwac_per_kwdc[i] = std::max(0.0, kwac_per_kwdc[i]);
     }
     reopt_pv.assign("prod_factor_series_kw", kwac_per_kwdc);
     // The above already includes losses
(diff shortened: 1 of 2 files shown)


dguittet (commented):
can you do a check if gen exists before adding the prod timeseries? I'm pretty sure the way it's run in the GUI calls simulate on the system first so gen should definitely exist, but technically this equation doesn't require simulate to be called beforehand.

brtietz:
> can you do a check if gen exists before adding the prod timeseries? I'm pretty sure the way it's run in the GUI calls simulate on the system first so gen should definitely exist, but technically this equation doesn't require simulate to be called beforehand.

Good thinking. What should the code do if gen doesn't exist? Throw an error, or revert to the lat/lon behavior? If the latter, is there a good losses assumption?

dguittet:
Yes I think using the lat/lon, which basically has REopt call PVWatts. Could losses be an optional input?

dguittet (approved):

Issue #909, opened 2022-10-26
Improve utility rates documentation in vartable


brtietz:
The PySAM documentation for some utility rate varables is sparse:

![image](https://user-images.githubusercontent.com/5530592/198092413-f8915a14-4546-4226-9e9f-fb810acc00e8.png)

This is due to very few columns in utility_rates_common being filled out.

Fill out as much detail as possible to allow users to more easily specify things like tier units.

brtietz:
Closed in https://github.com/NREL/ssc/pull/926

PR #926, opened 2022-11-02
add units and metadata for utility rates inputs

brtietz:
Add additional documentation to the vartable to solve (improve)? #909 

Shouldn't change any code behavior.


cpaulgilman (approved):
Very helpful. 

For `ur_ec_tou_mat`, consider revising Column 3 description to "col 3: max usage units (0=kWh, 1=kWh/kW, 2=kWh daily, 3=kWh/kW daily)"

cpaulgilman:
Colons `:` in variable metadata separate attribute name from description, so using them in the variable description causes problems with rendering in PySAM doc generation:

For example, colon in dostrings:

image

Results in PySAM documentation:

image

Replacing colons with equal signs fixes issue in PySAM doc:

image

PR #927, opened 2022-11-09
Update pvsamv1, pvwattsv8 var tables

mjprilliman:
-Fix loss input variables for pvsamv1 subarrays 2 through 4,
-Add integer, min, and max constraints for module_model parameters where appropriate
-align gcr, use_wf_albedo defaults in var table for pvwattsv8 to pvsamv1
-update pvwattsv8 test results affected by change to use weather file albedo


cpaulgilman (approved):
Looks good. Perhaps set the "META" attribute for all of the `[xxx]_is_bifacial` variables to `0=monofacial,1=bifacial` so they are consistent with `cec_is_bifacial`.

PR #931, opened 2022-11-11
Addresses SAM #1223 in all of ssc

sjanzou:


tyneises (approved):

tyneises:
FYI @qualand @Matthew-Boyd 

@qualand can you review the SolarPilot project files that changed?

qualand:
@tyneises The SolarPILOT changes look good. The majority of the changes were fabs() to std::abs(). Do you notice any issues with fabs() or just abs() not using the std namespace?

PR #941, opened 2022-11-17
Fix rollover issues in bill w/o system

brtietz:
Fixes https://github.com/NREL/SAM/issues/1273 See issue for test file. Values of bill w/o system should not change based on whether or not excess $ or kWh rollovers exist.

Establish additional variables so w/system outputs don't affect w/o system calculations. Look at the utility bill in year 2 in the test to check against these issues.


sjanzou (approved):
Looks good and test file from issue now shows no without system change.

PR #952, opened 2022-12-08
fix gen length issue that broke REopt sizing call

brtietz:
Properly account for the fact that PVWatts only generates one year of data when calling the REopt API. Fixes https://github.com/NREL/SAM/issues/1290

Rigorous testing would include multiple timesteps and analysis periods for both PVWatts-Battery and Detailed PV-Battery


cpaulgilman (approved):
Looks good. Tested with hourly data / analysis_period = 25 and 15-minute data / analysis_period = 30.

PR #955, opened 2022-12-22
Update cmod_irradproc.cpp small documentation detail

shirubana:
small update to irradiance input mode 'irrad_mode', as it has option to take info from POA from reference cell or pyranometer expanding options to 0-4

@dguittet


dguittet (approved):

dguittet:
Thanks for the update!

PR #960, opened 2022-12-29
Sandia clearance height

mjprilliman:
-Set clearance height for Sandia module model to use with rear irradiance calculations (no bifacial calculations)
-Fixes https://github.com/NREL/SAM/issues/1296


cpaulgilman (approved):
Tested in Debug mode before and after this fix, and confirm that Sandia module model does not hang after fix. Curious why @mjwagner2 is using the Sandia model?

Issue #947, opened 2022-12-01
Lifetime Electricity Load is Zero for PVWatts


brtietz:
**Describe the bug**
PVWatts  and PVWatts-Battery cases display zero lifetime electricity load for all steps, even though the bill load is >0.

Note that bill load is used for electricity rate calculations, so this doesn't affect any other outputs.

**To Reproduce**
1. Run a PVWatts case
2. Plot bill load and lifetime electricity load
3. See error

**Expected behavior**
Either PVWatts shouldn't display lifetime electricity load, or the results should match bill load

**Screenshots**
![image](https://user-images.githubusercontent.com/5530592/205123503-3993eac1-e645-4d67-90b8-ba1b64a241c4.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 2021.12.02, 2022.11.21

PR #963, opened 2023-01-10
Remove unused lifetime load output for non-lifetime projects

mjprilliman:
-Requirement check in utilityrate5.cpp output variable not working
-Using system_use_lifetime_output check in utilityrate5 instead
-'Lifetime load' doesn't make sense for non-lifetime projects, so removing variable rather than using it for single year outputs

Fixes https://github.com/NREL/ssc/issues/947


brtietz (approved):
Thank you!

cpaulgilman (approved):
Looks good -- I tested for PVWatts and Wind.

PR #969, opened 2023-01-19
Import json as SAM case

tyneises:
Updates to MSPT compute module to enable importing jsons (e.g. generated via Shift + F5) as a SAM case:
- Convert 9 independent TOD values into a single array value for pricing TOD
- New data type for "adjust" inputs

This pull request also includes modifications relevant for non-MSPT technologies:
- New data type for "adjust" inputs, and compute modules updated accordingly

@sjanzou Please add anything I'm missing


cpaulgilman:
It looks like the change to adjustment factors correctly imports inputs from the SAM UI as-is, so we won't need to use the version upgrade script to handle opening files from older versions of SAM. @sjanzou, is that correct?

Not sure what impact this will have on the need for the "AdjustmentFactors" group in PySAM, for example in the Pvwattsv8 documentation: https://nrel-pysam.readthedocs.io/en/master/modules/Pvwattsv8.html#adjustmentfactors-group. We may need to add instructions for converting to the new table format in the PySAM documentation.

ssc/cmod_etes_electric_resistance.cpp
```diff
@@ -149,7 +149,7 @@ static var_info _cm_vtab_etes_electric_resistance[] = {
...
     { SSC_INPUT,  SSC_MATRIX, "dispatch_sched_weekend",        "PPA pricing weekend schedule, 12x24",                           "",             "",                                  "Time of Delivery Factors",                 "ppa_multiplier_model=0&etes_financial_model<5&is_dispatch=1&sim_type=1",      "",              "SIMULATION_PARAMETER"},
-    { SSC_INPUT,  SSC_NUMBER, "dispatch_factor1",              "Dispatch payment factor 1",                                     "",             "",                                  "Time of Delivery Factors",                 "ppa_multiplier_model=0&etes_financial_model<5&is_dispatch=1&sim_type=1",      "",              "SIMULATION_PARAMETER"},
+ /*{SSC_INPUT,  SSC_NUMBER, "dispatch_factor1",              "Dispatch payment factor 1",                                     "",             "",                                  "Time of Delivery Factors",                 "ppa_multiplier_model=0&etes_financial_model<5&is_dispatch=1&sim_type=1",      "",              "SIMULATION_PARAMETER"},
```
brtietz, line 152: Generally I prefer deleting old code rather than commenting it out. Is there a reason this needs to stay as commented out code?

ssc/cmod_etes_electric_resistance.cpp
```diff
@@ -698,18 +701,27 @@ class cm_etes_electric_resistance : public compute_module
 
                     tou_params->mc_pricing.mv_is_diurnal = true;
 
-                    bool are_all_assigned = is_assigned("dispatch_sched_weekday") || is_assigned("dispatch_sched_weekend")
-                        || is_assigned("dispatch_factor1") || is_assigned("dispatch_factor2") || is_assigned("dispatch_factor3")
-                        || is_assigned("dispatch_factor4") || is_assigned("dispatch_factor5") || is_assigned("dispatch_factor6")
-                        || is_assigned("dispatch_factor7") || is_assigned("dispatch_factor8") || is_assigned("dispatch_factor9");
+                    // isn't this any assigned? Shouldn't all be and &&?
```
brtietz, line 704: Agree that this should have been && based on other context. (but please delete commented out code)

ssc/common_financial.cpp
```diff
@@ -385,15 +367,16 @@ bool dispatch_calculations::compute_outputs( std::vector<double>& ppa)
...
-	double dispatch_factor2 = m_cm->as_double("dispatch_factor2");
-	double dispatch_factor3 = m_cm->as_double("dispatch_factor3");
-	double dispatch_factor4 = m_cm->as_double("dispatch_factor4");
-	double dispatch_factor5 = m_cm->as_double("dispatch_factor5");
-	double dispatch_factor6 = m_cm->as_double("dispatch_factor6");
-	double dispatch_factor7 = m_cm->as_double("dispatch_factor7");
-	double dispatch_factor8 = m_cm->as_double("dispatch_factor8");
-	double dispatch_factor9 = m_cm->as_double("dispatch_factor9");
+
+    double dispatch_factor1 = m_dispatch_tod_factors[0];// m_cm->as_double("dispatch_factor1");
```
brtietz, line 371: Is the comment intended to indicate the former code? In that case, say `// Formerly m_cm->as_double("dispatch_factor1");`

ssc/common_financial.cpp
```diff
@@ -1221,29 +1204,43 @@ bool dispatch_calculations::setup()
...
+
+    m_periods.resize(8760, 1);
 	ssc_number_t *ppa_multipliers = m_cm->allocate("ppa_multipliers", 8760);
 
 	for (int i = 0; i < 8760; i++)
 	{
 		m_periods[i] = tod[i];
 
+        if ((tod[i] >= 1) && (tod[i] <= 9)) {
+
```
brtietz, line 1219: Does this generate a warning? If so, please transition to a "not" format. If not, please add `// Nothing to do` so it's clear this was intentional.

test/main.cpp
```diff
@@ -54,8 +54,8 @@ GTEST_API_ int main(int argc, char **argv) {
 
     //    filter to include
     //    ::testing::GTEST_FLAG(filter) = "CmodPVWatts*:CMPvwatts*";
-    //    ::testing::GTEST_FLAG(filter) = "etes_ptes_test.EtesPtesCmod*";
-
+        ::testing::GTEST_FLAG(filter) = "SAM_as_JSON_test.*";
```
brtietz, line 57: This needs to be commented out so we can see all the tests as a part of the PR.

ssc/cmod_swh.cpp
```diff
@@ -89,10 +89,7 @@ static var_info _cm_vtab_swh[] = {
...
-	{ SSC_INPUT,        SSC_MATRIX,      "shading:mxh",           "Month x Hour beam shading loss",       "%",      "",                                  "SWH",              "?",                       "",                                  "" },
-	{ SSC_INPUT,        SSC_MATRIX,      "shading:azal",          "Azimuth x altitude beam shading loss", "%",      "",                                  "SWH",              "?",                       "",                                  "" },
-	{ SSC_INPUT,        SSC_NUMBER,      "shading:diff",          "Diffuse shading loss",                 "%",      "",                                  "SWH",              "?",                       "",                                  "" },
+    { SSC_INPUT,        SSC_TABLE,      "shading",               "Shading loss table",                 "",         "",                                             "SWH",      "?",                        "",                             "" },
```
brtietz, line 92: This might need to have more detail so the auto-generated PySAM documentation is interpret-able.

brtietz (changes requested):
Code seems to work, and this is a great improvement, but there needs to be a lot of clean-up around commented out code, the test runner, and documentation of the new SSC_TABLE variables.

tyneises:
> It looks like the change to adjustment factors correctly imports inputs from the SAM UI as-is, so we won't need to use the version upgrade script to handle opening files from older versions of SAM. @sjanzou, is that correct?
> 
> Not sure what impact this will have on the need for the "AdjustmentFactors" group in PySAM, for example in the Pvwattsv8 documentation: https://nrel-pysam.readthedocs.io/en/master/modules/Pvwattsv8.html#adjustmentfactors-group. We may need to add instructions for converting to the new table format in the PySAM documentation.

@cpaulgilman I also do not see any need to add anything to the Version Upgrade script at this point.

I used Shift+F5 in the SAM 2022.12.22 release to save a LK script of the default MSPT case. Then I tried to run it from this branch with the SDKtool. The error message SSC returns because 'adjust' is in the wrong format is not very helpful - see screenshot below. @sjanzou is it possible to return a more descriptive error message for adjust and sf_adjust from SSC?

image

sjanzou:
> > It looks like the change to adjustment factors correctly imports inputs from the SAM UI as-is, so we won't need to use the version upgrade script to handle opening files from older versions of SAM. @sjanzou, is that correct?
> > Not sure what impact this will have on the need for the "AdjustmentFactors" group in PySAM, for example in the Pvwattsv8 documentation: https://nrel-pysam.readthedocs.io/en/master/modules/Pvwattsv8.html#adjustmentfactors-group. We may need to add instructions for converting to the new table format in the PySAM documentation.
> 
> @cpaulgilman I also do not see any need to add anything to the Version Upgrade script at this point.
> 
> I used Shift+F5 in the SAM 2022.12.22 release to save a LK script of the default MSPT case. Then I tried to run it from this branch with the SDKtool. The error message SSC returns because 'adjust' is in the wrong format is not very helpful - see screenshot below. @sjanzou is it possible to return a more descriptive error message for adjust and sf_adjust from SSC?
> 
> image

@tyneises, that error message is a standard precheck of variable type for compute modules generated from compute_module::verify function in core.cpp:
image

The "adjust" variable is missing because it now is an SSC_TABLE and matches with the VV_TABLE structure in SAM.

Previously, the "adjust" variable from SAM was translated into the following four variables in ssc:
image

I am not sure what message would be most useful in this case and that would apply to all cases where several ssc variables have been replaced with one ssc variable.

Thoughts?

tyneises:
@sjanzou I modified test/main.cpp to include all tests. A lot fail because we need to fix the adjust and dispatch factor inputs...

tyneises:
> @sjanzou I modified test/main.cpp to include all tests. A lot fail because we need to fix the adjust and dispatch factor inputs...

@sjanzou I changed the dispatch_factorX instances in the test inputs to dispatch_tod_factors. 'dispatch_factor1' is still in input in the var table in cmod_ippppa.cpp. Can you update the var table to use the array if necessary?

ssc/common.cpp
```diff
@@ -1009,15 +1021,19 @@ adjustment_factors::adjustment_factors( compute_module *cm, const std::string &p
 //adjustment factors changed from derates to percentages jmf 1/9/15
 bool adjustment_factors::setup(int nsteps, int analysis_period) //nsteps is set to 8760 in this declaration function in common.h
 {
-	ssc_number_t f = m_cm->as_number( m_prefix + ":constant" );
-	f = 1.0 - f / 100.0; //convert from percentage to factor
+    auto& table = m_cm->get_var_table()->lookup(m_prefix)->table;
```
tyneises, line 1024: @sjanzou This adds "is_assigned" so that scripts are only required to provide the constant adjust. What do you think?

tyneises:
> > @sjanzou I modified test/main.cpp to include all tests. A lot fail because we need to fix the adjust and dispatch factor inputs...
> 
> @sjanzou I changed the dispatch_factorX instances in the test inputs to dispatch_tod_factors. 'dispatch_factor1' is still in input in the var table in cmod_ippppa.cpp. Can you update the var table to use the array if necessary?

@sjanzou I tried to update all the tests for adjust tables. There is a test that is crashing the tests, but I'm not sure that is due to adjust. Can you please take a look?

test/ssc_test/cmod_pvsamv1_test.cpp
```diff
@@ -210,7 +210,14 @@ TEST_F(CMPvsamv1PowerIntegration_cmod_pvsamv1, LossAdjustmentLifetime) {
     double timeindex[1];
     for (int i = 0; i < 1; i++)
         timeindex[i] = 100;
-    ssc_data_set_array(data, "adjust:timeindex", (ssc_number_t*)timeindex, 1);
+
+    //ssc_data_set_array(data, "adjust:timeindex", (ssc_number_t*)timeindex, 1);
+    // First, get existing adjust var table
+    var_table* adjust_vt = static_cast<var_table*>(ssc_data_get_table(data, "adjust"));
```
tyneises, line 216: It looks like that we need to add to the adjust table that we're getting from upstream, rather than create a new table? @sjanzou @brtietz Can you please check that the static cast and overall approach in this file makes sense?

sjanzou: Your cast looks okay, the test is failing with the "shading" table input setup in the default case
image

sjanzou: @tyneises, there are many changes that need to be made upstream with updated settings for the test cases...
The "CMPvsamv1PowerIntegration_cmod_pvsamv1.LossAdjustmentLifetime" test is now passing with the latest commit using the updating of the "adjust" factor method that you were using.
image

tyneises: Thanks!

ssc/cmod_pvwattsv7.cpp
```diff
@@ -603,10 +603,12 @@ class cm_pvwattsv7 : public compute_module
 
         // read all the shading input data and calculate the hourly factors for use subsequently
         // timeseries beam shading factors cannot be used with non-annual data
-        auto& table = get_var_table()->lookup("shading")->table;
-        if (table.is_assigned("en_timestep") && table.as_boolean("en_timestep") && !wdprov->annualSimulation())
-//        if (is_assigned("shading:timestep") && !wdprov->annualSimulation())
-            throw exec_error("pvwattsv7", "Timeseries beam shading inputs cannot be used for a simulation period that is not continuous over one or more years.");
+        if (is_assigned("shading")) {
```
tyneises, line 606: @sjanzou Need to check here if shading is assigned, or the table.is_assigned("en_timestep") crashes. This fix seems a little clunky because shad.setup also checks if "shading" is assigned, but it works.

ssc/common_financial.cpp
sjanzou: superfluous comments removed

ssc/cmod_swh.cpp
sjanzou: added single variables back for better external language and ssc processing

ssc/common_financial.cpp
sjanzou: Cleaned up and removed else statement.

test/input_json/FinancialModels/equpartflip/2022.08.08_develop_branch_Geothermal_Power_All_Equity_Partnership_Flip_cmod_equpartflip.json
```diff
@@ -192,6 +192,7 @@
     "dispatch_factor7" : 1,
     "dispatch_factor8" : 1,
     "dispatch_factor9" : 1,
```
brtietz, line 194: Should we be clearing the old dispatch factors out of the test data? Is it best practice to reexport these for this type of change?

sjanzou: Best practices would probably be to tie to latest release and then re-export - what do you think?

sjanzou:
> Code seems to work, and this is a great improvement, but there needs to be a lot of clean-up around commented out code, the test runner, and documentation of the new SSC_TABLE variables.

Comments removed and tests being updated

sjanzou:
> Code seems to work, and this is a great improvement, but there needs to be a lot of clean-up around commented out code, the test runner, and documentation of the new SSC_TABLE variables.

Everything is flattened now, so the SSC_TABLE variables from SAM are non-existent.

sjanzou (approved):
SSC_TABLE removed from inputs from SAM and all requested changes made

brtietz (approved):

PR #970, opened 2023-01-20
Fix broken SDKTool with rapid json in wex/utils

sjanzou:
Wanted to get the ssc changes for rapidJSON in wex into patch to prevent continuously merging patch into SAM_504 branch until wex and SAM are finished.
This should be the only change in ssc necessary for SAM issue 504.


brtietz (approved):
Looks good!

PR #972, opened 2023-01-22
Changing runner to macos-latest to address pvsmoothing test failures

sjanzou:
macos-11 runner seems to have some resource issues. The PVSmoothing tests are one of the most memory intensive tests. The tests run locally on Intel mac and M1 mac. macos-latest runner succeeded with push. Now testing with this pull request.


brtietz (approved):
Looks good. This is also consistent with the SAM runner. Thank you!

PR #979, opened 2023-01-26
Ensure key variables in the powerflow code have a minimum of zero

brtietz:
Fixes https://github.com/NREL/SAM/issues/1324

The specific combination of factors in the test file was driving batt_to_load_ac negative, which lead to a positive batt_to_grid_ac, which caused an infinite loop between the powerflow code and the dispatch code. Fix this by preventing these variables from going negative.


cpaulgilman (approved):
Looks good. Test case simulation finishes with outage loss reported in results.

Issue #948, opened 2022-12-01
Negative losses cause PV+Battery case to freeze


brtietz:
**Describe the bug**
Specifying a negative number in time-series losses causes the PV+Battery case to get stuck. A hunch is that this is an infinite loop in the powerflow code.

**To Reproduce**
1. Open a detailed PV-Battery Case
2. Turn the grid outage on
3. Set the battery losses to time series and specify a negative number
4. Run the case and watch it get stuck 

**Expected behavior**
Any infinite loops in the powerflow code should be fixed. Also throw an error to prevent this.

The stand alone battery case does not have this problem

**Desktop (please complete the following information):**
 - Version 2022.11.21, maybe others

brtietz:
The issue is here: https://github.com/NREL/ssc/blob/abd6cc431802b2d70aa3b65126b06da2ad9fa000/shared/lib_battery_powerflow.cpp#L486 Given a negative system loss, this will always cause infinite recursion. Therefore the only thing that needs to be done is preventing negative system losses, there's nothing that needs to change in the powerflow code to solve this issue.

PR #980, opened 2023-01-26
add exec errors to prevent simulations from running with negative system losses for batteries

brtietz:
Fixes https://github.com/NREL/ssc/issues/948

Test file: 
[losses_test_case.zip](https://github.com/NREL/ssc/files/10513514/losses_test_case.zip)


mjprilliman (approved):
Thanks for fixing.

PR #984, opened 2023-01-28
Add support for WIND Toolkit CSV file formats

cpaulgilman:
Fixes https://github.com/NREL/SAM/issues/962

Also partially addresses https://github.com/NREL/SAM/issues/815 by adding support for CSV files downloaded from [RE Explorer](https://www.re-explorer.org/)

Next step is to update wind data download widget in SAM UI to download from wtk-download instead of wtk-srw-download, and possibly other WIND Toolkit endpoints at https://developer.nrel.gov/docs/wind/wind-toolkit/ (note that not all endpoints use the same CSV file format).

For endpoints that use different file format, consider writing a macro to convert files. For example, compare files from offshore-mid-atlantic-download to wtk-download.

SAM should now work with any of the wind data files here: 
[wind-file-samples.zip](https://github.com/NREL/ssc/files/10524013/wind-file-samples.zip)

To test, run test-wind-files.lk script included in .zip archive.

Should also spot check other wind data files in both WIND Toolkit CSV format and SRW format.

**This requires a new Help topic describing WIND Toolkit CSV format.**


cpaulgilman:
Replace with https://github.com/NREL/ssc/pull/985 to try to fix CI build errors

PR #985, opened 2023-01-28
Support WIND Toolkit CSV format

cpaulgilman:
Replace https://github.com/NREL/ssc/pull/984 due to CI errors.

- New CSV file format for wind.
- Improve wind resource error reporting.
- Remove extra commas in test SRW files.
- Handle "N/A" in data columns from WIND Toolkit files.

Fixes https://github.com/NREL/SAM/issues/962

Also partially addresses https://github.com/NREL/SAM/issues/815 by adding support for CSV files downloaded from [RE Explorer](https://www.re-explorer.org/)

Next step is to update wind data download widget in SAM UI to download from wtk-download instead of wtk-srw-download, and possibly other WIND Toolkit endpoints at https://developer.nrel.gov/docs/wind/wind-toolkit/ (note that not all endpoints use the same CSV file format).

For endpoints that use different file format, consider writing a macro to convert files. For example, compare files from offshore-mid-atlantic-download to wtk-download.

SAM should now work with any of the wind data files here: 
[wind-file-samples.zip](https://github.com/NREL/ssc/files/10524013/wind-file-samples.zip)


brtietz:
@sjanzou Could we get your eyes on this? I can't tell why patch would build, but this would get a linker error on TCS.

cpaulgilman:
When trying to build SAM on Linux, I see build errors in ptesdesignptdialog.cpp caused by functions like GetSelectedMaterial() in the PTESDesignPtDialog class returning `string` instead of `wxString` -- not sure if that is related to the build failure here.

sjanzou:
/Users/runner/work/ssc/ssc/ssc/shared/lib_windfile.cpp:513:127: error:
cannot pass object of non-trivial type 'std::string' (aka
'basic_string<char, char_traits<char>, allocator<char>>') through variadic
function; call will abort at runtime [-Wnon-pod-varargs]
3269
<https://github.com/NREL/ssc/actions/runs/4032614475/jobs/6941522919#step:9:3270>
m_errorMsg = util::format("data must be in local time: data time zone %s
and site time zone %s are not the same", tz_data, tz_site);
3270
<https://github.com/NREL/ssc/actions/runs/4032614475/jobs/6941522919#step:9:3271>
^
3271
<https://github.com/NREL/ssc/actions/runs/4032614475/jobs/6941522919#step:9:3272>/Users/runner/work/ssc/ssc/ssc/shared/lib_windfile.cpp:513:136:
error: cannot pass object of non-trivial type 'std::string' (aka
'basic_string<char, char_traits<char>, allocator<char>>') through variadic
function; call will abort at runtime [-Wnon-pod-varargs]
3272
<https://github.com/NREL/ssc/actions/runs/4032614475/jobs/6941522919#step:9:3273>
m_errorMsg = util::format("data must be in local time: data time zone %s
and site time zone %s are not the same", tz_data, tz_site);

The Mac chokes when using std::string for %s…

Need to use a char* - so, please update to tz_data.c_str() and
tz_site.c_str()

Thanks,
Steve

mjprilliman (commented):
Test files worked in the UI and in included LK script

cpaulgilman:
Thanks @sjanzou for the solution to the mac build issue.

mjprilliman (approved):

PR #989, opened 2023-02-07
Fix Geothermal cases with capacity factor > 1

mjprilliman:
-Capacity factor calculation was comparing different system capacity and output based on different 2nd law efficiency calculations for resource temperature < 150
-Source of this calculation not found within GETEM or documentation
-Removing this < 150 calculation removes erroneous capacity factor outputs


shared/lib_geothermal.cpp
```diff
@@ -1485,7 +1485,8 @@ double CGeothermalAnalyzer::GetPlantBrineEffectiveness(void)
 //	double dAEMaxPossible = (geothermal::IMITATE_GETEM) ? GetAEBinary() -  GetAEBinaryAtTemp(TamphSiO2) : GetAE() - dAE_At_Exit; // watt-hr/lb - [10B.GeoFluid].H54 "maximum possible available energy accounting for the available energy lost due to a silica constraint on outlet temperature"
 
 	mp_geo_out->max_secondlaw = (1 - ((geothermal::IMITATE_GETEM) ? GetAEBinaryAtTemp(TamphSiO2) / GetAEBinary() : dAE_At_Exit / GetAE()) - 0.375);
-	double dMaxBinaryBrineEffectiveness = ((geothermal::IMITATE_GETEM) ? GetAEBinary() : GetAE()) * ((GetTemperaturePlantDesignC() < 150) ? 0.14425 * exp(0.008806 * GetTemperaturePlantDesignC()) : mp_geo_out->max_secondlaw);
+	//double dMaxBinaryBrineEffectiveness = ((geothermal::IMITATE_GETEM) ? GetAEBinary() : GetAE()) * ((GetTemperaturePlantDesignC() < 150) ? 0.14425 * exp(0.008806 * GetTemperaturePlantDesignC()) : mp_geo_out->max_secondlaw);
```
sjanzou, line 1488: Do you want to keep this comment for future reference?

sjanzou (approved):
Looks good for CF > 1

shared/lib_geothermal.cpp
mjprilliman: I prefer to leave it to make it easier to revert if need be.

PR #991, opened 2023-02-10
SAM issue 1348 - price signal dispatch hard crash

sjanzou:


brtietz (approved):
Thanks for cleaning this up!

PR #993, opened 2023-02-13
Addresses ssc issue 992

sjanzou:
No longer have random NaNs in financial model outputs for single year technology model runs with analysis periods greater than one

### DIFF
--- a/ssc/cmod_singleowner.cpp
+++ b/ssc/cmod_singleowner.cpp
@@ -1064,31 +1064,39 @@ public:
 		int add_om_num_types = as_integer("add_om_num_types");
 		ssc_number_t nameplate1 = 0;
 		ssc_number_t nameplate2 = 0;
-        std::vector<double> battery_discharged;
-        std::vector<double> fuelcell_discharged;
-        for (int i = 0; i <= nyears; i++) {
-            battery_discharged.push_back(0);
-            fuelcell_discharged.push_back(0);
-        }
-        //throw exec_error("singleowner", "Checkpoint 1");
-		if (add_om_num_types > 0) //PV Battery
-		{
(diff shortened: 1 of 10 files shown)


ssc/cmod_singleowner.cpp
```diff
@@ -1064,31 +1064,39 @@ class cm_singleowner : public compute_module
...
+            escal_or_annual(CF_om_production2_expense, nyears, "om_fuelcell_variable_cost", inflation_rate, 0.001, false, as_double("om_production_escal")*0.01);
+            escal_or_annual(CF_om_capacity2_expense, nyears, "om_fuelcell_capacity_cost", inflation_rate, 1.0, false, as_double("om_capacity_escal")*0.01);
+            nameplate2 = as_number("om_fuelcell_nameplate");
+            fuelcell_discharged = as_vector_double("fuelcell_annual_energy_discharged");
+        }
+        if (fuelcell_discharged.size()== 1) { // ssc #992
+            double first_val = fuelcell_discharged[0];
+            fuelcell_discharged.resize(nyears, first_val);
+         }
+        if (fuelcell_discharged.size() != nyears)
```
brtietz, line 1098: The ssc test failures appear to be caused by this line. Fuel cell and battery use different length conventions for output variables: battery is nyears, fuel cell is nyears + 1. I believe battery to be correct, but can't find where we wrote down the convention of the financial models pre-pending year 0.

sjanzou: @brtietz , the convention as far as I know is the the technology models start at year 1 and the financial models start at year zero. So, I have no idea about the cmod_fuelcell allocations:
image

I have updated all financial models to handle the fuel cell exception to the technology model rule in the last commit.

brtietz (approved):

PR #996, opened 2023-02-17
Address SAM issue 1232 for Detailed PV and PVWattsv8

sjanzou:


brtietz (approved):
This works well, thank you!

cpaulgilman (approved):
Like this approach. Thanks.

PR #998, opened 2023-02-21
Fix pysam issue 138 - change all weather group names to "weather" case sensitive

sjanzou:


dguittet (commented):
Can these be all capitalized-- since the other group names are also capitalized?

dguittet (approved):

brtietz (approved):
Worked for me after a clean build. Thank you!!

PR #999, opened 2023-03-03
Etes sco2 dev

qualand:
Added numerical integration to calculate average specific heat over a temperature range. This allows modeling of user-defined heat transfer fluids with phase change.


tcs/sco2_cycle_components.cpp
```diff
@@ -1529,7 +1529,7 @@ int C_comp_multi_stage::design_given_outlet_state(int comp_model_code, double T_
 
     double max_calc_tip_speed = mv_c_stages[0]->ms_des_solved.m_tip_ratio;      //[-]
 
-	double tip_speed_limit = 0.85;
+    double tip_speed_limit = 0.8;
```
tyneises, line 1532: What's the reason for this change?

tcs/htf_props.cpp
```diff
@@ -91,6 +91,7 @@ bool HTFProperties::SetUserDefinedFluid( const util::matrix_t<double> &table )
 	// Set class member data
 	m_userTable = table;	
 	m_fluid = User_defined;
+    m_integration_points = 100;
```
tyneises, line 94: Does this default the user-defined HTF to 100 points? Does it make sense to potentially have more integration points than data points in the lookup table?

tcs/csp_solver_pc_Rankine_indirect_224.cpp
```diff
@@ -2286,3 +2277,32 @@ double C_pc_Rankine_indirect_224::Interpolate(int YT, int XT, double X, double Z
 
 } // Interpolate
 
+double C_pc_Rankine_indirect_224::Calculate_T_htf_cold_Converge_Cp(double q_dot_htf /*kWt*/, double T_htf_hot /*K*/, double m_dot_htf /*kg/s*/)
```
tyneises, line 2280: Would this method make more sense in the HTF class so we can use it more broadly?

tcs/csp_solver_pc_heat_sink.cpp
```diff
@@ -112,7 +112,7 @@ void C_pc_heat_sink::init(C_csp_power_cycle::S_solved_params &solved_params)
 	}
 
 	// Calculate the design point HTF mass flow rate
-	double cp_htf_des = mc_pc_htfProps.Cp_ave(ms_params.m_T_htf_cold_des+273.15, ms_params.m_T_htf_hot_des+273.15, 5);	//[kJ/kg-K]
```
tyneises, line 115: Is the '5' parameter represented in your implementation of cp_ave? I think it represents the number of integration points, which defaults to 1 in the new code?

tyneises (changes requested):

tcs/sco2_cycle_components.cpp
qualand: The more restrictive value (i.e., 0.85) was causing the simple cycle off-design calculations to fail due to the stage design breaking this limit. This change got the code working; however, I didn't dig into the implications. I could send you a small example of the code breaking case if you would like.

tcs/htf_props.cpp
qualand: No, it doesn't default the user-defined HTF table to 100 points. This is used in calculating the average Cp value across a arbitrary temperature range. I chose 100 arbitrarily. This could be more robust. This could be set by an optional input. Ideally, we would determined this dynamically depending on the number of points in the user table between the temperature range of the Cp calculation. However, this might take a bit of time to implement.

tcs/csp_solver_pc_Rankine_indirect_224.cpp
qualand: Probably so... However, I don't know how generally robust this convergence method is and if it would be better to switch it to the monotonic equation solver...

tcs/csp_solver_pc_heat_sink.cpp
qualand: The 5 parameter is the number of integration points in this old code. Correct, the new code defaults to 1 if the HTF properties are internal and not a user table. A value of 1 integration points returns the Cp value at the midpoint which was the original implementation throughout most of the system model. It was unclear to me why this particular instance used 5 integration points... it looks like you added it 7 years ago. https://github.com/NREL/ssc/commit/3793ab574c124d8c4ed1d866e4782b34571c15af

tcs/sco2_cycle_components.cpp
qualand: I couldn't find the case that was causing this issue. I have reverted this change.

tyneises (approved):

PR #1004, opened 2023-03-08
Size the collector loop mass flow based on n_collectors

### LINKED ISSUE #1003: Solar Water Heating Mass Flow Incorrect
The collector loop mass flow parameter mdot is fixed at its default value of 0.091055 kg/s regardless of the number of collectors in the loop. This default value appears to be derived from the test flow rate of the default Gobi 408 collector and the default number of collectors (2).

Matthew-Boyd:
Fixes the collector loop solar water heating mass flow as this is not currently dependent on the number of collectors, with the value currently being a constant sized for two collectors.

@cpaulgilman Can you please make sure I'm not misunderstanding what is going on here? I think the UI originally sized this parameter but no longer does.

Closes #1003

### DIFF
--- a/ssc/cmod_swh.cpp
+++ b/ssc/cmod_swh.cpp
@@ -277,7 +277,7 @@ public:
 		double test_flow = as_double("test_flow"); // collector test flow rate (kg/s)
 
 		/* collector properties */
-		double mdot_total = as_double("mdot"); // total system mass flow rate (kg/s)
+        double mdot_total = as_double("test_flow") * as_integer("ncoll"); // total system mass flow rate (kg/s)
 		double area_total = as_double("area_coll") * as_integer("ncoll"); // total solar collector area (m2)
 		double area_coll = as_double("area_coll");
 


cpaulgilman (commented):
@Matthew-Boyd I removed a user input "Total system flow rate" that set the value of the SSC input `mdot` to resolve https://github.com/NREL/SAM/issues/1153.

Looking at that now, it seems like we should restore that input. Removing it was the result of some internal discussions that concluded that SSC was not using the input.

Matthew-Boyd:
@cpaulgilman It's definitely using that input and bad behavior occurs when the number of collectors is increased.

My opinion is that the mass flow through the collector array should either be constant and proportional to the number of collectors and their respective test flow rates, or it should be automatically adjusted within an operating window to control the outlet temperature. Currently in the code it is a constant value irrespective of the number of collectors, which doesn't make sense. I also don't think the user should need to calculate this, nor should it be calculated in a UI equation because it won't translate to scripting or PySAM.

cpaulgilman:
@Matthew-Boyd That makes sense.

cpaulgilman (approved):

PR #1008, opened 2023-04-04
GitHub actions workflow update for deprecated ubuntu-18.04 

sjanzou:


brtietz (approved):
Gets the same test results as develop, so I think it's ok to move ahead with this change.

PR #1011, opened 2023-04-12
Fix cold tank issues in the solar water heating model

Matthew-Boyd:
These fixes limit the tank cold node temperature to a physically realizable temperature and outputs the average instead of the final temperature of the delivered water when the tank drains during a time step.

These fixes were prompted after doing parametric analyses of the hot outlet set temperature.

Before:
image

After:
image


cpaulgilman (approved):
Looks good. 

https://github.com/NREL/SAM/pull/1391 updates SAM test results.

PR #1013, opened 2023-04-18
Sg2 implementation

mjprilliman:
-Attempt to speed up SPA implementation by integrating equations from SG2 algorithm
-Paper: https://www.sciencedirect.com/science/article/pii/S0038092X12002800
-Github: https://github.com/gschwind/sg2/tree/main/src
-Replaces const static 3D arrays of values with single line approximations to speed up runtime
-Does not replace Greenwich mean sidereal time (v0) calculation as I haven't figured out the conversion (in lib_irradproc.cpp): 
image

-ToDo: update test results (changes on the order of 1e-4)
-ToDo: remove unused commented code


sjanzou (commented):
Code looks good....

I sent the script for the DC_AC_ratio tests to regenerate the json files
Tests that are currently failing:
image

mjprilliman:
FYI @kandersolar from our previous discussions we are finding significant speed up from the SG2 modifications to SPA.

kandersolar:
Thanks for the heads up @mjprilliman! 

Of course it would be a much bigger job than merely dropping in a new algorithm under the hood, but rather than replacing the SPA, would it be worthwhile to implement both SPA and SG2 and allow the user to choose which one is used?  (it's easy for me to make suggestions, as someone that doesn't have to put in the work to make it happen 😄)

mjprilliman:
> Looks good. I ran some simulation run time tests for PVWatts, Detailed PV and Physical Trough in both SAM and SDKtool with results here: [simulation-runtime-results.zip](https://github.com/NREL/ssc/files/11288486/simulation-runtime-results.zip)
> 
> pvwattsv8 runs twice as fast with SG2 than SPA on my Windows 11 laptop with average run times of 374 ms for SPA and 174 ms for SG2.

Interesting. When @sjanzou ran comparisons for pvwattsv5 (to compare back to before original SPA was implemented) we were seeing ~3x speed increase. Hopefully 2x speed up still works for pvwattsv8 usage?

[ssc2023.4.14_spa_sg2.zip](https://github.com/NREL/ssc/files/11288556/ssc2023.4.14_spa_sg2.zip)

cpaulgilman:
> > Looks good. I ran some simulation run time tests for PVWatts, Detailed PV and Physical Trough in both SAM and SDKtool with results here: [simulation-runtime-results.zip](https://github.com/NREL/ssc/files/11288486/simulation-runtime-results.zip)
> > pvwattsv8 runs twice as fast with SG2 than SPA on my Windows 11 laptop with average run times of 374 ms for SPA and 174 ms for SG2.
> 
> Interesting. When @sjanzou ran comparisons for pvwattsv5 (to compare back to before original SPA was implemented) we were seeing ~3x speed increase. Hopefully 2x speed up still works for pvwattsv8 usage?
> 
> [ssc2023.4.14_spa_sg2.zip](https://github.com/NREL/ssc/files/11288556/ssc2023.4.14_spa_sg2.zip)

I just ran tests comparing pvwattsv5 in SSC 279 (SPA) and SSC 280 (SG2) and see similar results to Steve's: 297 ms vs 87 ms. SSC 242 (pre-SPA) is 55 ms. The PVWatts API team will probably want to do some tests on pvwattsv8 with SG2 to see how this affects the API run times -- I know they are working on some changes to the API to speed things up as well.

sjanzou (approved):
Looks good and great that @cpaulgilman got similar speed up results and all ssc tests pass.

cpaulgilman (approved):

Issue #831, opened 2022-06-02
Voltage Table does not consider internal resistance


brtietz:
The equations in voltage_table_t do not consider the R parameter, and thus result in overly optimistic DC-DC efficiency with the battery. Determine if there is a reasonable way to add an I*R term to the table equations to model this loss.

Issue #569, opened 2021-04-27
Add voltage functions that better calculate power to reach a given SOC


brtietz:
The function voltage_table_t::calculate_max_discharge_w takes q as an input, but this q is the full amount of charge left in the battery down to 0 SOC, rather than a target SOC specified by dispatch and the capacity model. Create functions in the voltage classes (or at least for the voltage table) that calculate the charge and discharge power required to reach a specific SOC, instead of just to 0 or 100%. Then use these in the dispatch classes to better enforce the SOC constraints.

brtietz:
This issue is a lot less prevalent after https://github.com/NREL/ssc/pull/833. Do some tests and checks in the debugger, it is possible that PR was sufficient to solve this.

brtietz:
A possibly related issue `_Battery->energy_to_fill` utilizes the nominal voltage, which can result in under-estimating the energy required to charge the battery to a given SOC. This frequently results in the battery under-charging during automated dispatch, and then needing to charge at more expensive timesteps.

brtietz:
Given #831 being fixed, this is now solely a dispatch optimization. De-prioritizing until we have more bandwidth for that type of work.

PR #1016, opened 2023-04-21
Add internal resistance to voltage table calculations

brtietz:
A draft implementation of internal resistance for voltage table batteries. This dropped the DC-DC efficiency for lead acid from 99.5% to 95% - if we lower the default internal resistance. Sources including figure 6 of https://batteryuniversity.com/article/how-does-internal-resistance-affect-performance imply that 0.25 Ohms is too high, and that 0.01-0.02 would be more appropriate.

Companion SAM PR: https://github.com/NREL/SAM/pull/1493

Fixes https://github.com/NREL/ssc/issues/831


brtietz:
The goals include both reducing the efficiency (ideally making it more accurate) and harmonizing the assumptions of the voltage table and the thermal model. I'm also wondering if this will assist with some of the issues we have with the voltage table staying within the SOC limits, such as https://github.com/NREL/ssc/issues/569 and the comment on https://github.com/NREL/SAM/issues/1208, but I don't have any evidence that it will so far.

If we don't go this route, we should make it more obvious in the UI that the internal resistance isn't used in the voltage table, just the thermal model.

brtietz:
After my previous post, I remembered that I do have numbers on improved accuracy. Without this code, if you set the AC-DC and DC-AC efficiencies to 100%, the DC-DC efficiency is 99.5%, which is suspiciously high. Additionally, it doesn't change with internal resistance. With this code and the adjustments to lead-acid internal resistance discussed above, the DC-DC efficiency is more in line with the lithium ion models. I'm missing specific numbers for that right now because I don't have the code checked out.

dguittet:
So I actually worry that having the resistance variable fill both a thermal model role and a voltage role is not only incorrect in that they shouldn't be the same value (they are in fact different models with different parameters and weren't "fitted" together), but also causes issues for users to tune the resistance since now you may have overheating of the battery if you want a lower RT efficiencity, or what not.

I wonder if it wouldn't be more user-friendly to allow a "roundtrip efficiency" input for the voltage table.

brtietz:
At least for the NMC data, I fixed the resistance based on Kandler's paper and tuned h and Cp for the thermal model (link to private repo: https://github.com/NREL/sam-analysis/blob/main/battery_validation/thermal_fit.py). It seems like these should at least be correlated: if the resistance in the voltage model increases so should the resistance in the thermal model, and then it's on the battery thermal management system (h) to shed the extra heat.

Are there resistances that the thermal model needs to capture that aren't in the voltage model? If so, should we plan on adding a second parameter?

dguittet:
I would guess the thermal resistance is over the entire battery module whereas the voltage model is on the cell level, so there are all the additional pieces that go into putting many cells together. However my concern is not primarily with correctly representing different types of resistance. It's more that users don't have these parameters and if the point of adding resistance is to represent a lower, more realistic roundtrip efficiency, which actually is meaningful and available to the user, then we should just allow that.

brtietz:
How much do we care about the effect of charging having a higher voltage than discharging? That strikes me as the main thing we lose with the DC-DC efficiency as a percentage approach.

I also did a quick scan of data sheets from the search "battery spec sheet", and internal resistance (or internal impedance) was easier for me to find than DC-DC conversion percent.

dguittet:
My chats with a utility / developer in industry is that they do get battery roundtrip efficiencies from the manufacturer, with a typical range of 80-85% including auxiliaries, converters, losses, etc.

dguittet:
@brtietz Let me know when this is no longer a draft PR and is ready for review.

brtietz:
@dguittet I wanted to get https://github.com/NREL/SAM/pull/1493 ready prior to review. It's ready for review now, thank you!

dguittet (approved):
LGTM

PR #1017, opened 2023-04-25
Allow more flexible SAM CSV for solar weather file format

### LINKED ISSUE #937: Weather file check header rows issue
**Describe the bug**

SAM's weather file reader generates an error if the [first two rows of a SAM CSV file have a different number of columns](https://github.com/NREL/ssc/blob/9bf63eb20df559ee210c5a79f3f858e554102180/shared/lib_weatherfile.cpp#L790), assuming they are header rows with comma-separated values.

This helped us recently identify some of the NSRDB weather files that had missing version numbers, and this new issue with commas in come of the header values. 

Consider making the weather file check more robust so that it continues to work over time as the information in the header rows changes. SAM only needs the latitude, longitude, elevation, and time zone from the header, so ideally should be able to read the file as long as that information is available.

**To Reproduce**

1. Download  and extract sample SAM CSV file with unequal number of columns in first two header rows:  
[33.73_-118.26_33.7323_-118.263_psm3-tmy_60_tmy.zip](https://github.com/NREL/ssc/files/10016792/33.73_-118.26_33.7323_-118.263_psm3-tmy_60_tmy.zip)
2. Start SAM and run  Solar Resource File Checker macro.
6. See error

**Expected behavior**

Error should only be generated if latitude, longitude, elevation and/or time zone is missing from header (or invalid?)

**Additional Notes**

The online PVWatts Calculator and PVWatts API run the weather file checker, so solution needs to work with weather files returned by https://developer.nrel.gov/docs/solar/data-query/v2/ in addition to weather files from the NSRDB (test NSRDB PSM3 tmy-2020 data for Washington, DC addresses).

cpaulgilman:
- Allow files that have minimum required metadata, regardless of number of columns in metadata header rows.
- Add Version to list of metadata columns to support NSRDB version numbers.
- Test with files from NSRDB, PVWatts API, and hand-edited to have different metadata row lengths.

Fixes #937

Weather files for testing on Solar Resource  input page with any solar tech in SAM UI: Should add all of these files to solar resource library and run a simulation, except WIND Toolkit file, which should be ignored: [ssc-937-test-weather-files.zip](https://github.com/NREL/ssc/files/11326254/ssc-937-test-weather-files.zip)

Update SAM CSV for Solar Help topic to add version to table of accepted metadata items. (Done in local branch for Patch 2 release in separate PR.)

### DIFF
--- a/shared/lib_weatherfile.cpp
+++ b/shared/lib_weatherfile.cpp
@@ -792,23 +792,30 @@ bool weatherfile::open(const std::string& file, bool header_only)
         getline(ifs, buf);
         auto cols = split(buf);
         int ncols = (int)cols.size();
+
         getline(ifs, buf1);
         auto cols1 = split(buf1);
         int ncols1 = (int)split(buf1).size();
 
         int hdr_step_sec = -1;
 
-        if (ncols != ncols1)
-        {
-            m_message = "first two header lines must have same number of columns";
-            return false;
-        }
+        // allow metadata rows to have different lengths as long as required data is included
+        ncols = std::min(ncols, ncols1);
+
+        std::string name, value;
 
         for (size_t i = 0; (int)i < ncols; i++)
         {
-            const std::string name = util::lower_case(trimboth(cols[i]));
-            const std::string value = trimboth(cols1[i]);
 
+            name = "";
+            if (!cols[i].empty())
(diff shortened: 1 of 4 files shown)


sjanzou (approved):
Weather files working without notices for extra rows and columns, e.g.:
image

Issue #1007, opened 2023-03-20
Voltage Table can improperly calculate current when two depths of discharge have the same value


brtietz:
**Describe the bug**
At step 202 in the attached SAM file, the battery is charged with 400 kW, when the PV system is only providing 108 kW.

**To Reproduce**
1. Download the attached SAM file: 
[pvBattery_cad.zip](https://github.com/NREL/ssc/files/11023785/pvBattery_cad.zip)
2. Run case cad3_fixedPV
4. Look at power flow, voltage, and current at step 202

**Expected behavior**
The issue is caused by the voltage table having the same voltage for two depths of discharge. SAM should throw an error in this configuration, and users should specify unique points. Also add to the documentation that these points need to be unique.

**Screenshots**
![image](https://user-images.githubusercontent.com/5530592/226486050-9f11f154-d454-4bca-be10-df61fcce0849.png)

![image](https://user-images.githubusercontent.com/5530592/226486122-9d5efb2a-e51b-4298-a708-87b2a3668e24.png)

PR #1020, opened 2023-05-04
Throw an error if duplicate entries in the voltage table produce zero slope

brtietz:
Fixes https://github.com/NREL/ssc/issues/1007

Testing with the case in the issue should produce an error message, testing with the defaults should not.


cpaulgilman:
Updated Help to explain that duplicate voltage values are not allowed in voltage table for Fall 2023 version.

Issue #1015, opened 2023-04-19
Check that load and critical load data have same time step


cpaulgilman:
**Describe the bug**

SAM UI checks added in https://github.com/NREL/SAM/pull/1390 display UI error messages when critical load and load data have different time steps, but SSC does not prevent simulation from running under this condition.

**To Reproduce**

1. Using a build of SAM from the Patch branch that includes https://github.com/NREL/SAM/pull/1390, open the SAM file and load the 15-minute weather file from [here](https://github.com/NREL/SAM/files/11275249/crit-load-timestep-test.zip).
2. Go to the Grid Outage and Electric Load pages and see the error messages.
3. Run a simulation and note no simulation message. 

On Data Tables tab of Results page, note that "Critical load in this timestep" and "Lifetime electricity load" are of correct length: Hourly critical load data is converted to 15-minute data.

Critical load data 60-minute input:
![image](https://user-images.githubusercontent.com/27711933/233139079-88690ae2-09b1-47c4-a3f9-81d3f031edf3.png)

Resulting output, with 60-minute critical load data converted to 15-minute time steps:
![image](https://user-images.githubusercontent.com/27711933/233138792-a412f25b-c6bb-4bde-9170-2ad858e4a114.png)

**Expected behavior**
I would expect a simulation message if time steps do not match, but perhaps the automatic conversion is the behavior we intended?

brtietz:
@cpaulgilman I created an SSC branch that should fix this (ssc_1015_errors_for_load_mismatch), however when using the test file above the errors are not triggered. The issue seems to be that while `load_user_data` has 15 minute data, `load` is still being passed to SSC as an 8760:

![image](https://user-images.githubusercontent.com/5530592/236005242-6693379e-af4c-4b24-b9e4-9d4581ec3546.png)

Do you have time to debug the callbacks and equations to see why that might be happening? I'll create a pull request once we get the expected behavior with the test file.

cpaulgilman:
Yes, I'll look into the SAM UI side of things.

cpaulgilman:
Callbacks should be fixed in https://github.com/NREL/SAM/tree/load_ui_checks.

PR #1021, opened 2023-05-05
Add errors for length mismatch between critical load and electric load

brtietz:
Add checks for critical load length as compared to load length. Fixes https://github.com/NREL/ssc/issues/1015 - SAM branch [load_ui_checks](https://github.com/NREL/SAM/tree/load_ui_checks) required to test. (pull request forthcoming)

Test should get error messages in the file in the issue.


cpaulgilman (approved):
Looks good. I revised some of the error messages for consistency.

UI callback fixed in https://github.com/NREL/SAM/pull/1397

PR #1022, opened 2023-05-09
Batt functions for reopt sizing

brtietz:
Add new function  Reopt_size_standalone_battery_params to assist with dGen PV and battery sizing using the battery compute module. Also update URDB Python call to version 8.

Pairs with SAM PR TBD and PySAM PR TBD


ssc/cmod_battery_eqns.cpp
```diff
@@ -108,3 +109,157 @@ bool Calculate_thermal_params(ssc_data_t data) {
     return true;
 }
 
+bool Reopt_size_standalone_battery_params(ssc_data_t data) {
```
dguittet, line 112: Can you make a test for this?

ssc/cmod_battery_eqns.cpp
```diff
@@ -108,3 +109,157 @@ bool Calculate_thermal_params(ssc_data_t data) {
...
+    reopt_wind.assign("max_kw", 0);
+
+    var_data* vd, * vd2;
+
+    double val1, val2, system_cap;
+    vt_get_number(vt, "system_capacity", &system_cap);
+
+    // financial inputs
+    map_optional_input(vt, "itc_fed_percent", &reopt_batt, "total_itc_pct", 0., true);
+    // TODO: what about reopt vars total_rebate_us_dollars_per_kw?
```
dguittet, line 132: What about this TODO? Are there any incentives?

ssc/cmod_battery_eqns.h
```diff
@@ -69,6 +69,53 @@ static const char* calculate_thermal_params_doc =
...
 
+static const char* Reopt_size_standalone_battery_params_doc =
+"Given battery costs and a load, get the optimal PV and battery sizes. Wind is disabled.\\n"
+"Maps SAM compute module inputs to those of the ReOpt API:\\n"
+"Battery technology paired with Residential, Commercial, Third Party or Host Developer financing.\\n\\n"
+"Optional: if missing, variable will be set to default value if documented, to REopt defaults otherwise.\\n"
+"Conditional: only required if the variable it's meant to replace is missing.\\n"
+"REopt's 'microgrid_upgrade_cost_pct' set to 0 because SAM doesn't apply this cost multiplier.\\n"
+"REopt inputs not used in this function will be REopt's defaults.\\n"
+"Some additional REopt inputs are required, such as lon and lat. See https://nrel.github.io/REopt.jl/dev/reopt/inputs/ for a full set of inputs\\n\\n"
```
dguittet, line 80: Is lat and lon still required?

ssc/cmod_battery_eqns.cpp
brtietz: That TODO has been in there for a while and got copy-pasted over. I'm thinking this TODO might be a better fit for a larger REopt API version upgrade closer to release time, the latest version doesn't include that variable: https://nrel.github.io/REopt.jl/dev/reopt/inputs/#Financial

cpaulgilman (approved):

ssc/cmod_battery_eqns.h
dguittet: I don't see lat and lon in the new battery eqn being passed in?

brtietz: The battery compute module does not have lat and lon, they need to be added to the post after this function call.

dguittet: I see that it's added in the pv equation. So is this standalone battery post a viable submission to REopt in itself?

brtietz: It is not. This line is intended to be a hint that additional inputs are required, is there a better way to phrase it?

dguittet: "The post produced here is not a complete submission. Further additions are required, such as x and y, performed by function z"?

In that case, what is the intended use case in adding it to PySAM's standalone battery module? The user then adds the missing variables?

dguittet (approved):

PR #1028, opened 2023-05-25
Fixes #1027

### LINKED ISSUE #1027: CSP models use different class for adjustment factors
For CSP models, the `sf_adjust` variable uses `sf_adjustment_factors` class and `adjust` variables use `adjustment_factors`. Need to fix so that `sf_adjust` variables work correctly with "constant", "time series" and "custom period" values from LossAdjustment widget in SAM UI.

sjanzou:


tyneises (approved):

cpaulgilman:
Fixes https://github.com/NREL/SAM/issues/1401

cpaulgilman (approved):
This seems to fix the problem of heliostat availability (`sf_adjust`) not being applied correctly for the Power Tower model in either time series or custom periods mode. Would be good for @tyneises to confirm.

PR #1031, opened 2023-05-26
Give price signals dispatch info about real time system power and load for dispatch for apples to apples comparison of forecast error

brtietz:
Provides the same information to price signals dispatch as peak shaving has available at each step: assumes the battery can read current PV and load to make dispatch decisions

NPV of current patch code for forecast case: $8,484,012
NPV of this code: $8,731,707

File to test with: 
[price_signals_forecast_diff.zip](https://github.com/NREL/ssc/files/11578280/price_signals_forecast_diff.zip)


dguittet (approved):
Looks reasonable to me

PR #1043, opened 2023-08-28
Add absorber reflected radiation to heat loss output in evacreceiver

taylorbrown75:
Fix issue #1441 [https://github.com/NREL/SAM/issues/1441](url) 

Add reflective losses to receiver incident thermal power


tyneises (changes requested):
Can you please separate the q_3Reflect term from from the q_34tot term, so that the reporter thermal losses don't include the reflection losses?

tyneises (approved):

Issue #1026, opened 2023-05-22
arm64 failing ssc tests


sjanzou:
Four failing ssc tests on arm64 architecture (all 725 run fine on amd64 Windows, Linux and Mac)

1. lib_battery_test.runTestCycleAt3C

2. csp_common.TesSubcomponentCmod.Default

3. CMPvsamv1BatteryIntegration_cmod_pvsamv1.ResidentialDCBatteryModelPriceSignalDispatch

4. CMBattwatts_cmod_battwatts.NoPV

Detailed value difference for each test below

[0;32m[ RUN      ] [mlib_battery_test.runTestCycleAt3C
/Users/imacuser/Public/Projects/GitHub/NREL/ssc/test/shared_test/lib_battery_capacity_test.h:45: Failure
The difference between tested_state.q0 and expected_state.q0 is 0.025430198576842145, which exceeds tol, where
tested_state.q0 evaluates to 47.115430198576846,
expected_state.q0 evaluates to 47.090000000000003, and
tol evaluates to 0.01.
runTest: 3
/Users/imacuser/Public/Projects/GitHub/NREL/ssc/test/shared_test/lib_battery_test.h:142: Failure
The difference between tested_state.V and expected_state.batt_voltage is 0.02420895181734295, which exceeds 0.01, where
tested_state.V evaluates to 467.11420895181732,
expected_state.batt_voltage evaluates to 467.08999999999997, and
0.01 evaluates to 0.01.
runTest: 3
[0;31m[  FAILED  ] [mlib_battery_test.runTestCycleAt3C (373 ms)
[0;32m[----------] [m1 test from lib_battery_test (373 ms total)



[0;32m[----------] [m1 test from csp_common.TesSubcomponentCmod
[0;32m[ RUN      ] [mcsp_common.TesSubcomponentCmod.Default
/Users/imacuser/Public/Projects/GitHub/NREL/ssc/test/shared_test/lib_csp_tes_test.cpp:411: Failure
The difference between csp_subcomponent.GetOutputVector("T_sink_in")[idx] and T_sink_in_expected[idx] is 0.11852206011218414, which exceeds 0.1, where
csp_subcomponent.GetOutputVector("T_sink_in")[idx] evaluates to 383.69852206011217,
T_sink_in_expected[idx] evaluates to 383.57999999999998, and
0.1 evaluates to 0.10000000000000001.
at index: 18
[...]

PR #1049, opened 2023-09-13
Add new BTM battery dispatch method for self-consumption

janinefreeman:
This pull request and the complementary one in SAM add a new battery dispatch methodology for behind-the-meter PV+Battery and BTM battery hybrid systems that allow the battery to be dispatched to maximize self-consumption, which can also be termed "24/7 carbon-free energy" for larger industrial loads. Behind the scenes, this new dispatch option uses the grid power target option with the entire timeseries set to 0, but it is a new dispatch option to enable this functionality for PySAM users as well. Added a few new tests to go along with the new dispatch method.

Remaining to be done: adding outputs to help users understand how well this dispatch method is achieving its target (zero grid power at each timestep)


brtietz:
Test failure here likely due to https://github.com/NREL/ssc/issues/1026. I propose updating the tolerance for the test for now and then lowering it again once we get to that issue.

shared/lib_battery_dispatch.h
```diff
@@ -57,7 +57,7 @@ class dispatch_t
 public:
 
 	enum FOM_MODES { FOM_AUTOMATED_ECONOMIC, FOM_PV_SMOOTHING, FOM_CUSTOM_DISPATCH, FOM_MANUAL };
-	enum BTM_MODES { PEAK_SHAVING, MAINTAIN_TARGET, CUSTOM_DISPATCH, MANUAL, FORECAST };
+	enum BTM_MODES { PEAK_SHAVING, MAINTAIN_TARGET, CUSTOM_DISPATCH, MANUAL, FORECAST, CARBON_FREE };
```
brtietz, line 60: I hate to slow things down, but having it be CARBON_FREE in the code and self consumption in the UI is going to set us up for confusion later. Mind fixing it to match the UI?

test/ssc_test/cmod_battery_pvsamv1_test.cpp
```diff
@@ -383,7 +388,12 @@ TEST_F(CMPvsamv1BatteryIntegration_cmod_pvsamv1, ResidentialDCBatteryModelIntegr
             pairs["batt_dispatch_wf_forecast_choice"] = 0;
             pairs["batt_dispatch_load_forecast_choice"] = 0;
             break;
-
+        case 3:
+            // Input grid power targets
```
brtietz, line 392: Update comment to match the correct name, please.

brtietz (changes requested):
I think naming consistency is worth the time here.

Line 426 is where the tolerance needs to be updated, maybe go with 1 for now?

shared/lib_battery_dispatch.h
janinefreeman: Ha, no you're absolutely right. I had that on my list, will change now.

test/ssc_test/cmod_battery_pvsamv1_test.cpp
janinefreeman: Done

brtietz (approved):
Thanks for the code cleanup!

PR #1050, opened 2023-09-13
PV subhourly clipping correction

mjprilliman:
-Implements Allen method of correcting PV energy output by underprediction of clipping at hourly intervals. 
-SDK feature for modifying default correction lookup matrix
-Adds Ineichen clearsky model for all irradiance calculations
-Adds tests of subhourly clipping calculation, Ineichen implementation


shared/lib_pv_io_manager.cpp
```diff
@@ -678,6 +678,13 @@ PVSystem_IO::PVSystem_IO(compute_module* cm, std::string cmName, Simulation_IO*
     }
 
     numberOfInverters = cm->as_integer("inverter_count");
+    if (cm->is_assigned("num_inverter_subhourly_clipping")) {
```
janinefreeman, line 681: As annoying as it is, I don't think this should be an input to pvsamv1. I think this value should instead be calculated here if the subhourly clipping model is enabled- you should have sufficient information to do that here, I think?

ssc/cmod_pvsamv1.cpp
```diff
@@ -645,6 +651,10 @@ static var_info _cm_vtab_pvsamv1[] = {
         { SSC_OUTPUT,        SSC_ARRAY,      "subarray1_modeff",                     "Subarray 1 Module efficiency",                                         "%",      "", "Time Series (Subarray 1)",       "*",                    "",                              "" },
         { SSC_OUTPUT,        SSC_ARRAY,      "subarray1_celltemp",                   "Subarray 1 Cell temperature",                                          "C",      "", "Time Series (Subarray 1)",       "*",                    "",                              "" },
         { SSC_OUTPUT,        SSC_ARRAY,      "subarray1_celltempSS",                 "Subarray 1 Cell temperature (steady state)",                           "C",      "", "Time Series (Subarray 1)",       "*",                    "",                              "" },
+        { SSC_OUTPUT,        SSC_ARRAY,      "subarray1_dni_index",                 "Subarray 1 DNI Index",             "",   "", "Time Series (Subarray 1)",       "",                    "",                              "" },
```
janinefreeman, line 654: What is this index representing? Seems like it might not make sense as an output.

ssc/cmod_pvsamv1.cpp
```diff
@@ -645,6 +651,10 @@ static var_info _cm_vtab_pvsamv1[] = {
...
+        { SSC_OUTPUT,        SSC_ARRAY,      "subarray1_poa_beam_front_cs",                 "Subarray 1 beam clearsky irradiance",             "W/m2",   "", "Time Series (Subarray 1)",       "",                    "",                              "" },
```
janinefreeman, line 655: The labels on these three outputs (for each subarray) need to have "POA" inserted for clarity- I got confused why they were attached to Subarray 1 because I thought they were just clearsky irradiance, not POA clearsky. Variable name is clear though!

ssc/cmod_pvsamv1.cpp
```diff
@@ -772,6 +794,7 @@ static var_info _cm_vtab_pvsamv1[] = {
         //SEV: total dc snow loss time series (not a required output)
         { SSC_OUTPUT,        SSC_ARRAY,      "dc_snow_loss",                         "DC power loss due to snow",						 "kW",   "",   "Time Series (Array)",       "",                    "",                              "" },
         { SSC_OUTPUT,        SSC_ARRAY,      "dc_net",                               "Inverter DC input power",                                       "kW",   "",   "Time Series (Array)",       "*",                    "",                              "" },
+        //{ SSC_OUTPUT,        SSC_ARRAY,      "dc_net_clearsky",                               "Inverter DC clearsky input power",                                       "kW",   "",   "Time Series (Array)",       "",                    "",                              "" },
```
janinefreeman, line 797: Can we remove debug output if no longer needed?

ssc/cmod_pvsamv1.cpp
```diff
@@ -792,6 +815,9 @@ static var_info _cm_vtab_pvsamv1[] = {
         { SSC_OUTPUT,        SSC_ARRAY,      "inv_total_loss",                       "Inverter total power loss",                            "kW",   "",   "Time Series (Inverter)",      "*",             "",                   "" },
         { SSC_OUTPUT,        SSC_ARRAY,      "ac_wiring_loss",                       "AC wiring loss",                                       "kW",   "",   "Time Series (Inverter)",      "*",                        "",                   "" },
         { SSC_OUTPUT,        SSC_ARRAY,      "ac_gross",                             "Inverter AC output power",                                       "kW",   "",   "Time Series (Array)",       "*",                    "",                              "" },
+        { SSC_OUTPUT,        SSC_ARRAY,      "clipping_potential",                       "Clipping potential",                                       "",   "",   "Time Series (Inverter)",      "",                        "",                   "" },
```
janinefreeman, line 818: These three outputs should be conditionally assigned, only if the subhourly clipping model is enabled, I think. @cpaulgilman - what do you think?

ssc/cmod_pvsamv1.cpp
```diff
@@ -946,6 +972,8 @@ static var_info _cm_vtab_pvsamv1[] = {
         // annual_ac_gross
         { SSC_OUTPUT, SSC_NUMBER, "annual_ac_inv_eff_loss_percent", "AC inverter efficiency loss", "%", "", "Loss", "", "", "" },
         { SSC_OUTPUT, SSC_NUMBER, "annual_ac_wiring_loss_percent", "AC wiring loss", "%", "", "Loss", "", "", "" },
+        { SSC_OUTPUT, SSC_NUMBER, "annual_subhourly_clipping_loss_percent", "Subhourly clipping correction loss percent", "%", "", "Loss", "", "", "" },
```
janinefreeman, line 975: @cpaulgilman curious on your thoughts if this output should be reported if the subhourly clipping loss model isn't enabled. I don't have a strong opinion one way or the other for the annual results.

ssc/cmod_pvsamv1.cpp
```diff
@@ -969,7 +997,10 @@ static var_info _cm_vtab_pvsamv1[] = {
             { SSC_OUTPUT, SSC_NUMBER, "annual_ac_after_inv_pntloss", "AC output after inverter night tare loss", "kWh", "", "Annual (Year 1)", "", "", "" },
             */
             { SSC_OUTPUT, SSC_NUMBER, "annual_ac_wiring_loss", "AC wiring loss", "kWh", "", "Annual (Year 1)", "", "", "" },
-            { SSC_OUTPUT, SSC_NUMBER, "annual_transmission_loss", "Transmission loss", "kWh", "", "Annual (Year 1)", "", "", "" },
+            { SSC_OUTPUT, SSC_NUMBER, "annual_subhourly_clipping_loss", "Subhourly clipping correction loss", "kWh", "", "Annual (Year 1)", "", "", "" },
+            { SSC_OUTPUT, SSC_NUMBER, "annual_subhourly_clipping_factor", "Subhourly clipping correction loss factor", "", "", "Annual (Year 1)", "", "", "" },
```
janinefreeman, line 1001: @cpaulgilman curious on your thoughts if these two annual outputs should be reported if the subhourly clipping loss model isn't enabled. I don't have a strong opinion one way or the other for the annual results.

ssc/cmod_pvsamv1.cpp
```diff
@@ -3506,4 +3698,30 @@ void cm_pvsamv1::inverter_size_check()
             SSC_WARNING);
 }
 
+const double Subhourly_Clipping_Matrix[21][21] =
```
janinefreeman, line 3701: Looks like this clipping matrix is also present in lib_shared_inverter.cpp? If intentional, needs comments, if unintentional, one can be removed :)

janinefreeman (changes requested):
Awesome job implementing a complex model, that needs to carry through to a bunch of different parts of the code! I do have some comments on a few workflow changes (posted in-line). I think they are mostly smaller changes. Happy to talk through any of my thoughts/comments.

shared/lib_pv_io_manager.cpp
mjprilliman: Updated.

ssc/cmod_pvsamv1.cpp
mjprilliman: Updated.

ssc/cmod_pvsamv1.cpp
mjprilliman: Removed.

ssc/cmod_pvsamv1.cpp
mjprilliman: _factor outputs were for debugging and were removed.

ssc/cmod_pvsamv1.cpp
mjprilliman: Removed.

janinefreeman (approved):

PR #1051, opened 2023-09-14
New Hybrid Configurations

cpaulgilman:
Defaults and UI for new hybrid configurations for Single Owner and Host Developer financial models:

PVWatts Wind Battery Hybrid

PVWatts Wind FuelCell Battery Hybrid

Photovoltaic Wind Battery Hybrid

Generic PVWatts Wind FuelCell Battery Hybrid

Goes with:

* https://github.com/NREL/SAM/pull/1448
* https://github.com/NREL/wex/pull/164


ssc/cmod_hybrid.cpp
```diff
@@ -0,0 +1,762 @@
...
+                pBattGen = ((var_table*)outputs)->lookup(batteries[0])->table.as_array("gen", &battGenLen);
+            }
+
+ 
+/*
+    //Run grid compute module with fuel cell output
+    run grid
+
+    //Run utilityrate if applicable
+    run utilityrate
```
brtietz, line 558: Electricity purchases are not showing up in cash flow for PVWatts-Wind-Battery Single Owner, and maybe others. Unclear if this is the problem or if it is somewhere else. Host-developer looks ok.

SAM file to reproduce: 
[hybrids-electricity-purchases.zip](https://github.com/NREL/ssc/files/12614224/hybrids-electricity-purchases.zip)

brtietz (commented):
@sjanzou would you mind doing a pass for comment cleanup on the hybrid cmod files? It's hard to tell which TODOs are still relevant and which are already done. Similarly some of the pseudocode could likely be deleted at this point.

ssc/cmod_hybrid.cpp
sjanzou: > Electricity purchases are not showing up in cash flow for PVWatts-Wind-Battery Single Owner, and maybe others. Unclear if this is the problem or if it is somewhere else. Host-developer looks ok.
> 
> SAM file to reproduce: [hybrids-electricity-purchases.zip](https://github.com/NREL/ssc/files/12614224/hybrids-electricity-purchases.zip)

Good catch - issue was that the metering option was set to net metering and would not bill for default configurations - metering option updated to buy all sell all like pv battery configuration and working per attached document.
[GitHubPullRequestsssc_1051.docx](https://github.com/NREL/ssc/files/12618853/GitHubPullRequestsssc_1051.docx)

brtietz (approved):
Thanks for the code cleanup!

brtietz:
I don't think this should block the merge, but recording issues here until we decide they go on the main GitHub pile. The two variables on the left here look the same to me:

image

Should the "power generated by renewable resource" be more analogous to the "Power produced by PV pre-curtailment" of the PV+Battery module?

sjanzou (approved):
looks good

PR #1054, opened 2023-09-21
Remove old cmods

janinefreeman:
This PR removes old/unused compute modules `poacalib`, `pvwattsv0`, `pvwattsv1`, `pvwattsv1_1ts`, and `pvwattsv1_poa`. These versions of pvwatts have not been used by the main SAM project or PVWatts websites in many years. They are still available in the code attic, but NREL will no longer be updating them to work with updates made to the libraries used in those files.

I believe I removed them in all the appropriate places, and they were not in any of the SAM or PySAM api files to begin with, but would appreciate a review of that.


dguittet (approved):
I think this shouldn't affect PySAM, but I'll have to check that after I resolve other errors in export_config. I think this should just be merged for now and anything in PySAM can be fixed later

PR #1055, opened 2023-09-22
Address SAM issue 1184

sjanzou:


test/input_cases/general_data/pvsmoothing_Phoenix_Validation_alloptions.json
```diff
@@ -64,8 +64,8 @@
 	"subarray1_nmody" : 2,
 	"subarray1_backtrack" : 0,
 	"subarray2_enable" : 0,
-	"subarray2_modules_per_string" : 0,
-	"subarray2_nstrings" : 0,
+	"subarray2_modules_per_string" : 1,
```
mjprilliman, line 67: Is this going to effect default requirements as well? I.E. will subarray2_modules_per_string have to be set to 1 rather than 0 on case load even though subarray 2 is not enabled? Right now the default Detailed PV cases will fail with subarray 2-4 nstrings and modules_per_string set to 0. Guess it's a logical next step to update those defaults to allow SAM to still run when another subarray is added without extra steps.

mjprilliman (approved):

PR #1057, opened 2023-09-28
Add outputs for battery self-consumption dispatch

janinefreeman:
This PR adds 4 new outputs for the battery self-consumption dispatch method (conditionally defined so they don't show up for other dispatch methods)- number of timesteps that the electric load is met by the system for year 1 and lifetime, and the percentage of timesteps that each of those represent (year 1 and lifetime).


brtietz (approved):
This looks great, thank you!

PR #1060, opened 2023-10-02
Ssc 755 net to sales

brtietz:
Finish the "net" to "sales" conversion we started back in the 2021 release. Closes https://github.com/NREL/ssc/issues/755, for additional context see https://github.com/NREL/ssc/pull/756.

I believe this is an SSC only PR. The only SAM dependencies I could find were in API. My impression is that updates to API are on hold while we work out the export config issues, but if you want me to run the export I can do so.

To test: ensure that the monthly and by TOD period outputs in the FOM financial models are still recording kWh to the grid correctly.


cpaulgilman:
This changes all variables names beginning with "cf_energy_net_" to "cf_energy_sales". For example, `cf_energy_net_jan` changes to `cf_energy_sales_jan'. 

These are all output variables, so should only affect PySAM and LK scripts that are using those output variables.

cpaulgilman (approved):

sjanzou (approved):
Looks good and tests appear fine.

Issue #1058, opened 2023-09-28
Battery can discharge more than critical load during grid outage dispatch


brtietz:
**Describe the bug**
A battery can discharge more power than the critical load during the grid outage dispatch on develop.
While the "battery to load" powerflow is correct, multiplying the current by the voltage gives a full-power discharge during these steps, lowering the SOC prematurely.

**To Reproduce**
1. Open a PV+Battery case
2. Set critical load to 100%
3. Set an outage for the first 6+ hours of the year
4. The battery fully discharges in <2 hours, despite the critical load supporting a 4-5 hour discharge

Attached SAM case demonstrates the error: 
[crit_load_discharge_error.zip](https://github.com/NREL/ssc/files/12753765/crit_load_discharge_error.zip)

**Expected behavior**
Battery voltage times current should match critical load in steps like this. Works fine in the released version.

**Desktop (please complete the following information):**
 - Version Develop

brtietz:
Previous thinking on this topic: 
![image](https://github.com/NREL/ssc/assets/5530592/02d99982-cd96-4eed-94cf-9c85a5db17f9)

brtietz:
Related issue: https://github.com/NREL/SAM/issues/985

PR #1061, opened 2023-10-02
Ssc 1058 ac loss accounting

brtietz:
Improve loss accounting for AC connected batteries. Powerflow code is now more aware of losses when calculating batt_to_load and can increase power to meet the load. Under the previous implementation x% of the power to each location was lost across the board, rather than accounting for the dynamics of whether the load or grid is served first.

Closes https://github.com/NREL/ssc/issues/1058. Test file and expected behavior is available in that issue. DC batteries were unaffected and pass the same new unit tests without changes.

This still treats the loss as a giant resistor in the AC system, which may not be what we want. Changing the definition of the battery availability losses will be handled in the future as a part of https://github.com/NREL/SAM/issues/985


shared/lib_battery_powerflow.cpp
```diff
@@ -483,11 +489,8 @@ void BatteryPowerFlow::calculateACConnected()
             P_fuelcell_to_grid_ac = 0;
         }
 
+        // Priliminary batt to grid for DC losses
```
dguittet, line 492: this looks like a conjunction of prilliman and preliminary

dguittet (approved):
LGTM

shared/lib_battery_powerflow.cpp
brtietz: Thanks, fixed.

PR #1064, opened 2023-10-11
Ssc 1123 price signals to retail rate naming

brtietz:
Rename price signals dispatch to retail rate dispatch, bringing the old "forecast" enum along the way.

Fixes https://github.com/NREL/SAM/issues/1123

Companion SAM PR: 

Should be good to go if the tests pass, but any additional searches for "price signals" as a spot check would be appreciated.


cpaulgilman (approved):
Looks good. Pointing out some additional variables, not sure if these need changing.

* "Price signal" in comment of lib_utility_rate.cpp: https://github.com/NREL/ssc/blob/091de1b2338890b77b6fbfcf2eee88d0211cc3b0/shared/lib_utility_rate.cpp#L357

* Also `forecast_price_series_dollar_per_kwh` in cmod_battery.cpp/h  is for FOM battery, but I think refers to retail rates for electricity purchases, https://github.com/NREL/ssc/blob/091de1b2338890b77b6fbfcf2eee88d0211cc3b0/ssc/cmod_battery.cpp#L621

* `forecast_price_signal` class in common.h

* `vtab_forecast_price-signal` in cmod_pvsamv1.cpp

brtietz:
@cpaulgilman Thanks! I clarified that first comment to "retail rate forecast." The other three variables are for the front of meter automated dispatch, so I'll leave those alone.

PR #1066, opened 2023-10-17
Change cf_energy_net to cf_energy_sales

cpaulgilman:
Goes with https://github.com/NREL/SAM/pull/1501

Calculate O&M production based expense, PBI, PTC, LCOE from cf_energy_sales for all financial models.

Remove return on equity (ROE) from Community Solar model. Was originally implemented specifically for India and then removed from other financial models.


brtietz (approved):
Thanks for cleaning this up!

PR #1067, opened 2023-10-17
Fix MSPT cavity receiver system capacity calculation

tyneises:
Add default pumping power of 2% if not reported from receiver design method


cpaulgilman:
Fixes https://github.com/NREL/SAM/issues/1500

tyneises:
Simple fix to https://github.com/NREL/SAM/issues/1500#event-10680676101

A more detailed fix would add a design-point performance calculation to the cavity model

PR #1069, opened 2023-10-19
Tidal timeseries model

mjprilliman:
-Add tidal file reader module
-Add modeling of tidal energy output based on timeseries velocity data
-Add unit tests for cmod_mhk_tidal, tidal_file_reader
-Add LK functions for scanning tidal file library, viewing tidal velocity data in data viewer


cpaulgilman (changes requested):
Use single <user>/SAM Downloaded Weather File for downloaded solar, wind and marine energy files instead of separate folders.

Set default for tidal resource library selection to avoid pop-up error message.

Set default for tidal resource folder list to empty so "x" does not appear in list by default.

mjprilliman:
> Use single /SAM Downloaded Weather File for downloaded solar, wind and marine energy files instead of separate folders.
> Updated locations, need to test. 
> Set default for tidal resource library selection to avoid pop-up error message.
> Fixed
> Set default for tidal resource folder list to empty so "x" does not appear in list by default.
Believe this was a callbacks issue (wrong UI form title in on_load) more than a defaults issue. Defaults manager for all library_folder_list variables is "x" when I checked in defaults manager

cpaulgilman (approved):

PR #1070, opened 2023-10-19
Sam 1208 revisit safety factor

brtietz:
Better account for round trip efficiency in the peak shaving dispatch calculations, removing the need for the safety factor. Update the tests accordingly.

Based on the test updates, it looks like there was an error in some of the inputs checks for custom dispatch for battwatts. This has been corrected.

Fixes https://github.com/NREL/SAM/issues/1208

Pairs with SAM PR https://github.com/NREL/SAM/pull/1512 to update test results for SAM.


dguittet (approved):
lgtm

PR #1072, opened 2023-10-26
Remove total PTC output from all models

### LINKED ISSUE #644: cf_ptc_total doesn't work
**Describe the bug**
I assume `cf_ptc_total` = `cf_ptc_fed` + `cf_ptc_sta` but the total seems to be 0 even when fed is non-zero.

**To Reproduce**
1. Open default Wind Single Owner case (Federal PTC will be active)
2. Click on 'Simulate'
3. Try out this script: 
image

cpaulgilman:
Fixes #644

Remove cf_ptc_total from cashloan and common. 

Ignoring legacy cmod_ippppa.

### DIFF
--- a/ssc/cmod_cashloan.cpp
+++ b/ssc/cmod_cashloan.cpp
@@ -241,7 +241,6 @@ enum {
 	
 	CF_ptc_fed,
 	CF_ptc_sta,
-	CF_ptc_total,
 	
 	CF_sta_depr_sched,
 	CF_sta_depreciation,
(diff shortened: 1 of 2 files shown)


brtietz (approved):
Looks good!

Is there a reason cmod_ippppa hasn't been moved to the attic? I don't think it needs to happen as a part of this PR, but it would be good to keep all of the code in ssc valid and move what we aren't keeping up to date.

cpaulgilman:
Issue to move old IPP financial model code to attic: https://github.com/NREL/ssc/issues/1074

PR #1078, opened 2023-10-28
Sam 313 reopt v3 for resilience

brtietz:
Several changes to the REopt interface including:

- Update to API version 3 (see https://nrel.github.io/REopt.jl/dev/ for V3 documentation and https://github.com/NREL/REopt-Analysis-Scripts/blob/master/v2_v3_inputs_map.csv for inputs migration guide)
- Add ability to pass grid outages from SAM's grid_outage variable
- Pass through the battery grid charging variable to REopt API instead of requiring it in the GUI
- Reassign ac to dc and dc to ac efficiencies for better alignment between battery models.

Pairs with SAM PR https://github.com/NREL/SAM/pull/1527, see that PR for test instructions.


cpaulgilman (approved):
Implementation looks good in SSC.

May need to revise example code in PySAM documentation: https://nrel-pysam.readthedocs.io/en/main/examples.html?highlight=reopt#generate-inputs-for-a-reopt-api-call

dguittet (approved):
lgtm also

PR #1079, opened 2023-10-29
Fix fuel cell replacement costs for hybrids - SAM issue 1528

sjanzou:
Test with develop branches of other repos using SAM test file 
[fuel-cell-replacement-test.zip](https://github.com/NREL/ssc/files/13197762/fuel-cell-replacement-test.zip)

Works for hybrid configurations with user specified replacement schedules
image

and for percentage fuel cell replacements 
image


brtietz (approved):
Looks good, thank you!

Issue #1006, opened 2023-03-11
All technology outputs for financial models to be analysis period long or single values


sjanzou:
The fuel cell model has fuelcell_dispatched output of length analysis period + 1 and is used for production o and m costs in financial models.

The convention for all technology model outputs that are used in financial models is:
1. analysis period for lifetime mode
2. single values for single year technology compute modules

See discussion at https://github.com/NREL/SAM/issues/659#issuecomment-1464105549 for more details

janinefreeman:
@sjanzou has this been fixed already in the hybrids work, by chance?

sjanzou:
> @sjanzou has this been fixed already in the hybrids work, by chance?

Not yet fixed in hybrid branch 
<img width="972" alt="image" src="https://github.com/NREL/ssc/assets/6498311/005dda00-b5d2-4ef4-9efe-123042e6e864">

PR #1083, opened 2023-10-31
Addresses issue #1006

sjanzou:
All ssc tests passing locally on Windows
SAM defaults running on Windows (No SAM repo changes)


cpaulgilman (commented):
I'm getting a simulation error for the default Fuel Cell / Single Owner configuration:

image

cpaulgilman:
In the results, should the fuel cell annual energy values line up with other technologies, i.e., have length analysis period +1 with zero prepended? It seems like fuel cell is different from other techs:

image

image

cpaulgilman (approved):
Looks good. Fuel cell annual output values consistent with other techs for hybrid configuration, and start in Year 1 for original Fuel cell configuration.

Issue #905, opened 2022-10-25
Enforce charging constraints for PVSmoothing Variables


brtietz:
**Describe the bug**
During testing for https://github.com/NREL/ssc/pull/904, we discovered that the PV Smoothing algorithm assumes power is always available for charging, and so sometimes this can lead to discrepancies between PV Smoothing state of charge and battery state of charge. Add the system and grid charging constraints so that the algorithm is aware of the maximum power at any given timestep.

**To Reproduce**
1. Set up a PV smoothing case
2. Leave system charging on and grid charging off
3. Run a simulation and plot pv smoothing state of charge and battery state of charge

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/5530592/197896398-74c6aaa2-784c-47d2-a149-1687fc15eb74.png)

**Desktop (please complete the following information):**
 - Version: develop, likely the 2022 release

PR #1085, opened 2023-11-01
Ssc 905 pv smoothing pv charge only

brtietz:
Restrict PV power when grid charging isn't allowed. Prevents long term SOC mismatch as shown in the issue.

Fixes https://github.com/NREL/ssc/issues/905


sjanzou (approved):
Limiting looks good!

PR #1086, opened 2023-11-03
Helpful error messages when cell temperature heat transfer method is missing data

mjprilliman:
-Throw error in pvsamv1 when ambient pressure and dew point (or wet bulb) is missing from the weather file
-Make tcell nan when this happens to make sure calculation does not continue as normal
-Fixes https://github.com/NREL/sam/issues/1428

### DIFF
--- a/shared/lib_cec6par.cpp
+++ b/shared/lib_cec6par.cpp
@@ -480,7 +480,7 @@ bool mcsp_celltemp_t::operator() ( pvinput_t &input, pvmodule_t &module, double 
 
 	// convert to kelvin
 	double TA = input.Tdry+273.15;
-    if (isnan(input.Patm)) {
+    if (std::isnan(input.Patm)) {
         m_err = "Atmospheric pressure (millibar) is required for this temperature model.";
         Tcell = std::numeric_limits<double>::quiet_NaN();
         return false;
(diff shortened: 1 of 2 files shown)


mjprilliman:
@cpaulgilman review for error message.
@sjanzou review on if there is a better way of handling this error?

cpaulgilman (approved):
Looks good. Here are weather files I used for testing with heat transfer cell temp model: 
[weather-files-for-heat-transfer-cell-temp-tests.zip](https://github.com/NREL/ssc/files/13254565/weather-files-for-heat-transfer-cell-temp-tests.zip)

shared/lib_cec6par.cpp
```diff
@@ -480,6 +480,11 @@ bool mcsp_celltemp_t::operator() ( pvinput_t &input, pvmodule_t &module, double
 
 	// convert to kelvin
 	double TA = input.Tdry+273.15;
+    if (isnan(input.Patm)) {
```
sjanzou, line 483: Please use std::isnan instead of isnan, see https://github.com/NREL/SAM/issues/1223

shared/lib_cec6par.cpp
```diff
@@ -505,6 +510,11 @@ bool mcsp_celltemp_t::operator() ( pvinput_t &input, pvmodule_t &module, double
 	Fcs        = 1. - Fcg;              // !view factor between top of tilted plate and everything else (sky)
 	Fbs        = Fcg;                   // !view factor bewteen top and ground = bottom and sky
 	Fbg        = Fcs;                   // !view factor bewteen bottom and ground = top and sky
+    if (isnan(input.Tdew)) {
```
sjanzou, line 513: std::isnan()

ssc/cmod_pvsamv1.cpp
```diff
@@ -2188,6 +2188,7 @@ void cm_pvsamv1::exec()
                             tcell = Subarrays[nn]->customCellTempArray[inrec];
                         else {
                             (*Subarrays[nn]->Module->cellTempModel)(in[nn], *Subarrays[nn]->Module->moduleModel, module_voltage, tcell);
+                            if (isnan(tcell)) throw exec_error("pvsamv1", Subarrays[nn]->Module->cellTempModel->error());
```
sjanzou, line 2191: std::isnan please...

sjanzou (changes requested):
Please change isnan to std::isnan throughout changes per https://github.com/NREL/SAM/issues/1223

sjanzou (approved):
Thanks for updating!

PR #1089, opened 2023-11-15
Handle batt_look_ahead_hours <=2 with MIN constraint addresses SAM#1547

sjanzou:
Also, test size_t setting in if statement for valid value.

Goes with https://github.com/NREL/SAM/pull/1575


brtietz:
This fixed the crash, thanks! 

It looks like the unit test failure is due to:

`    error during simulation.Simulation time step is 60 minutes for pvwattsv8.
Simulation time step is 60 minutes for windpower.
exec fail(hybrid): battery_discharged size (30) incorrect
Value of: errors
  Actual: true
Expected: false
`

Any ideas on what might cause that?

sjanzou:
> This fixed the crash, thanks!
> 
> It looks like the unit test failure is due to:
> 
> ` error during simulation.Simulation time step is 60 minutes for pvwattsv8. Simulation time step is 60 minutes for windpower. exec fail(hybrid): battery_discharged size (30) incorrect Value of: errors Actual: true Expected: false`
> 
> Any ideas on what might cause that?

Fixing now - issue with input to the test

brtietz (approved):
Thanks for fixing the tests!

PR #1091, opened 2023-11-16
Update for pysam hybrids

dguittet:
The input technologies to `cmod_hybrid` require some financial variables that aren't listed as required, and aren't checked by the cmod. In PySAM, this means if the user doesn't have these extra financial variables, the cmod will just have a hard exit. Fix this by making a `vtab_hybrid_tech_om_inputs` that isn't added to any of the compute modules, but is added dynamically by `cmod_hybrid` to check that all these variables are present.

Other changes to `cmod_hybrid`:
 - Move check of required input input_tables to beginning of `exec`
 - Use `inflation_rate` and `analysis_period` from the "Hybrid" input_table which includes the financial cmods. The code originally requires each cmod to have its own `inflation_rate` and `analysis_period` which is problematic because not all technology cmods have these as inputs, and it seems to make it possible to have different values. However, this requires there to be a financial cmod to the hybrid cmod, and this is currently not enforced. Should a financial model be required for running a hybrid cmod?
 - Not all technology cmods have `system_use_lifetime_output`, such as `windpower`, so instead of adding a new interface variable to these cmods, I made cmod_hybrid assume False unless the variable is available and True.

Happy to discuss concerns about any of these proposed changes.


sjanzou:
> The input technologies to `cmod_hybrid` require some financial variables that aren't listed as required, and aren't checked by the cmod. In PySAM, this means if the user doesn't have these extra financial variables, the cmod will just have a hard exit. Fix this by making a `vtab_hybrid_tech_om_inputs` that isn't added to any of the compute modules, but is added dynamically by `cmod_hybrid` to check that all these variables are present.
> 
> Other changes to `cmod_hybrid`:
> 
> * Move check of required input input_tables to beginning of `exec`
> * Use `inflation_rate` and `analysis_period` from the "Hybrid" input_table which includes the financial cmods. The code originally requires each cmod to have its own `inflation_rate` and `analysis_period` which is problematic because not all technology cmods have these as inputs, and it seems to make it possible to have different values. However, this requires there to be a financial cmod to the hybrid cmod, and this is currently not enforced. Should a financial model be required for running a hybrid cmod?
> * Not all technology cmods have `system_use_lifetime_output`, such as `windpower`, so instead of adding a new interface variable to these cmods, I made cmod_hybrid assume False unless the variable is available and True.
> 
> Happy to discuss concerns about any of these proposed changes.

A financial compute module is currently required to run a hybrid compute module. We can revisit in the future. Required for the o and m and installation costs for each hybrid component separately in a hybrid system. The technology compute modules do not require the costs to run.

The battery and fuel cell compute modules are run in lifetime mode regardless of the generator compute modules lifetime mode settings.

sjanzou:
@dguittet, are you working on this branch or a different one for patch 1? Thanks!

dguittet:
@sjanzou Yes, this is the one still.

sjanzou (changes requested):
One test failing on Windows - all others are passing!
[  FAILED  ] CmodHybridTest.GenericPVWattsWindFuelCellBatteryHybrid_SingleOwner

ssc/cmod_hybrid.cpp
```diff
@@ -196,7 +214,10 @@ class cm_hybrid : public compute_module
                 else {
                     size_t count_degrad = 0;
                     ssc_number_t* degrad = 0;
-                    degrad = input.as_array("degradation", &count_degrad);
+                    if (compute_module == "generic_system")
+                        degrad = input.as_array("generic_degradation", &count_degrad);
+                    else
+                        degrad = input.as_array("degradation", &count_degrad);
```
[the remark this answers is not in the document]
sjanzou: generic_degradation is input to both cmod_generic_system and cmod_mhk_wave and is in the UI in Degradation AC Lifetime form as follows:
image

So, is used throughout the generic system.

Also, "degradation" is only used for non-lifetime simulation like PVWatts, pvsamv1 applies degradation outside of cmod_hybrid and is not used in cmod_hybrid (true for all generators with system_use_lifetime_output non-zero):
image

The code you showed for cmod_hybrid appears to be out of date... generic_degradation is not used in cmod_hybrid - when did you last merge patch into pysam_hybrids? Did it merge the latest from patch?

The latest patch branch does not have the lines 217 through 220 that you showed
https://github.com/NREL/ssc/blob/patch/ssc/cmod_hybrid.cpp

ssc/sscapi.cpp
```diff
@@ -1300,6 +1305,81 @@ SSCEXPORT ssc_bool_t ssc_module_add_var_info(ssc_module_t p_mod, ssc_info_t v)
     return 1;
 }
 
+SSCEXPORT ssc_bool_t ssc_module_hybridize(ssc_module_t p_mod)
+{
+    compute_module* cmod = static_cast<compute_module*>(p_mod);
+    if (!p_mod)
+        return 0;
+
```
dguittet, line 1313: add comment for why 35

ssc/cmod_fuelcell.cpp
```diff
@@ -120,7 +120,7 @@ cm_fuelcell::cm_fuelcell()
 	add_var_info(vtab_fuelcell_input);
 	add_var_info(vtab_fuelcell_output);
 	add_var_info(vtab_technology_outputs);
-    add_var_info(vtab_hybrid_tech_om);
+    add_var_info(vtab_hybrid_tech_om_outputs);
```
dguittet, line 123: move into hybridize?

ssc/cmod_hybrid.cpp
```diff
@@ -113,34 +117,45 @@ class cm_hybrid : public compute_module
...
-            int len, analysisPeriod = 0;
+            int len = 0;
             std::vector<size_t> genTimestepsPerHour;
             bool ts_adj = false; // keep track of whether time step is adjusted for log messages
 
+            // get financial inputs common to all technologies and copy into each tech's input tables
+            var_data* financial_compute_modules = input_table->table.lookup("hybrid");
+            int analysisPeriod = (int)financial_compute_modules->table.lookup("analysis_period")->num;
+            ssc_number_t inflation_rate = financial_compute_modules->table.lookup("inflation_rate")->num * 0.01;
+            ssc_number_t sales_tax_rate = financial_compute_modules->table.lookup("sales_tax_rate")->num * 0.01;
```
dguittet, line 128: Comment out and check tests

ssc/cmod_hybrid.cpp
```diff
@@ -195,8 +212,9 @@ class cm_hybrid : public compute_module
                 }
                 else {
                     size_t count_degrad = 0;
-                    ssc_number_t* degrad = 0;
-                    degrad = input.as_array("degradation", &count_degrad);
+                    ssc_number_t* degrad = input.as_array("degradation", &count_degrad);
+                    if (compute_module == "generic_system")
+                        input.assign("generic_degradation", *input.lookup("degradation"));
```
dguittet, line 217: @sjanzou Can check this?

ssc/cmod_hybrid.cpp
```diff
@@ -221,7 +239,7 @@ class cm_hybrid : public compute_module
                     escal_or_annual(input, pOMLandLease, analysisPeriod, "om_land_lease", inflation_rate, total_land_area, false, input.as_double("om_land_lease_escal") * 0.01);
                 }
                 // optional fossil fuel costs
-                if (compute_module_inputs->table.lookup("om_fuel_cost")) {
+                if (compute_module_inputs->table.lookup("system_heat_rate")) {
```
dguittet, line 242: trying to prevent seg fault of accessing the following variables

ssc/cmod_hybrid.cpp
```diff
@@ -410,14 +424,28 @@ class cm_hybrid : public compute_module
...
+                if (financial_compute_modules->table.is_assigned("dispatch_tod_factors"))
+                compute_module_inputs->table.assign("dispatch_tod_factors", *financial_compute_modules->table.lookup("dispatch_tod_factors"));
+                if (financial_compute_modules->table.is_assigned("grid_interconnection_limit_kwac"))
+                compute_module_inputs->table.assign("grid_interconnection_limit_kwac", *financial_compute_modules->table.lookup("grid_interconnection_limit_kwac"));
+                if (financial_compute_modules->table.is_assigned("ppa_escalation"))
+                compute_module_inputs->table.assign("ppa_escalation", *financial_compute_modules->table.lookup("ppa_escalation"));
+                if (financial_compute_modules->table.is_assigned("ppa_multiplier_model"))
+                compute_module_inputs->table.assign("ppa_multiplier_model", *financial_compute_modules->table.lookup("ppa_multiplier_model"));
+                if (financial_compute_modules->table.is_assigned("ppa_price_input"))
+                compute_module_inputs->table.assign("ppa_price_input", *financial_compute_modules->table.lookup("ppa_price_input"));
```
dguittet, line 445: reformat, and hopefully this makes sense and aligns with UI logic

dguittet:
@sjanzou Ready for review now that tests are updated from code generator and passing

sjanzou (approved):
Working on windows and results matching with patch branches ssc and SAM.

Great work!

PR #1093, opened 2023-11-16
Fixes to availability losses

### LINKED ISSUE #181: Allow timeseries losses in pvsavm1: dc_adjust and ac_adjust
Rather than only allowing hourly losses for dc_adjust and ac_adjust, allow timeseries (subhourly) losses in these buckets. Need to investigate how this affects all other models that use adjust class.

janinefreeman:
This PR fixes the setup of availability losses to properly determine the loss timestep. It updates the error messages to give more helpful information (and fixes a mislabeled error message in hcpv). It also updates the battery tests for these losses, which had the loss enabled when it wasn't being used. Fixes #181 . This does need one UI update to the widget in the wind model in the SAM repo, which I will wait until @cpaulgilman is finished doing his UI review to complete.

### DIFF
--- a/ssc/cmod_hcpv.cpp
+++ b/ssc/cmod_hcpv.cpp
@@ -403,7 +403,7 @@ public:
 
         adjustment_factors haf(this, "adjust");
         if (!haf.setup())
-            throw exec_error("pvwattsv5", "failed to setup adjustment factors: " + haf.error());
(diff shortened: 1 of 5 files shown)


cpaulgilman:
UI fixes are in https://github.com/NREL/SAM/pull/1580.

mjprilliman (approved):
Looks good when testing in PV.

PR #1097, opened 2023-11-28
Fix SSC 785 hourly variables

### LINKED ISSUE #785: Electricity to/from grid (year 1) does not match lifetime
It's a bit confusing that hourly kWh and lifetime kW "electricity to/from grid" are different in these results for the MiniGrid AC case of the DC_AC_Connected.sam test file -- lifetime hourly is zero for all time steps, as I think it should be due to the grid outage for all time steps (this is true in this branch and SAM 2021.12.02 r1):

ac-daily-losses-with-outage-ac-connected

_Originally posted by @cpaulgilman in https://github.com/NREL/ssc/issues/780#issuecomment-1083691218_

brtietz:
Fixes #785 

Add grid_outage as an optional input to utility_rate 5. When there is a grid outage, set grid use to zero. While cmod_grid handled this effectively for the load, these additional changes are required to preserve the data in gen while not accumulating net metering credits during an outage.

To test:
1. Open a PV+Battery behind the meter case
2. Set up a full year grid outage (file with 8760 ones for testing) with any critical load percentage: 
[_8760_ones.csv](https://github.com/NREL/ssc/files/13491312/_8760_ones.csv)
3. Run the case
4. Hourly to/from grid variables should be zero, net metering credits should not be earned.

Testing other compensation methods would be good rigor.

### DIFF
--- a/ssc/cmod_utilityrate5.cpp
+++ b/ssc/cmod_utilityrate5.cpp
@@ -55,6 +55,10 @@ static var_info vtab_utility_rate5[] = {
 
 	// input from user as kW and output as kW
 	{ SSC_INOUT, SSC_ARRAY, "load", "Electricity load (year 1)", "kW", "", "Load", "", "", "" },
+
+    // Optional input, only provided by battery cases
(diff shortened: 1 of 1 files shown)


cpaulgilman (approved):
Looks good. I checked all excess generation compensation methods.

PR #1100, opened 2023-12-05
Solve crash in retail rates dispatch SAM 1629

brtietz:
Throw an error if a user attempts to run retail rates dispatch without provided utility rate data. Update cmod-hybrid so it automatically gives the hybrid vartable to the battery and fuel cell models for dispatch.

Fixes https://github.com/NREL/SAM/issues/1629, see that PR for test instructions

Pairs with https://github.com/NREL/SAM/pull/1640, pull that PR for required defaults updates to get this to run.


cpaulgilman (approved):
PVWatts-Wind-Battery / TPO Host-Developer case runs with retail rate dispatch option and defaults for all other inputs. Simulation takes a while (long pause at 50%) but eventually finishes. The battery only charges and discharges for the first few days of January.

PR #1102, opened 2023-12-05
Fix CSP commas in var table

dguittet:


taylorbrown75:
Thank you for catching this

taylorbrown75 (approved):

PR #1103, opened 2023-12-06
Improve REopt error messages for invalid dgrules values

cpaulgilman:
Fixes https://github.com/NREL/SAM/issues/1621

Goes with https://github.com/NREL/SAM/pull/1641

Improve error message when SAM `net_metering` option is not available in URDB `dgrules` so that UI callback error in SAM makes more sense.

Rename URDB V7 to V8 in labels.


brtietz (approved):
Thanks for cleaning these up!

PR #1107, opened 2023-12-07
Add ssc_module_exec_with_error per @dguittet with filtering to fix #1105

### LINKED ISSUE #1105: cmod_hybrid needs to propagate errors from lower level cmods
Errors thrown in lower level cmods are ignored, and the error shown to the user is the next arbitrary error from cmod_hybrid.

Check the error status from each cmod called in cmod_hybrid and propagate notices, warnings, and errors to the user.

sjanzou:
Note filtering to SSC_WARNING and SSC_ERROR only. SSC_NOTICE includes several messages that are not useful on simulation results dialog for test file (PVWatts tilt = 95) :
[ssc_1105.zip](https://github.com/NREL/ssc/files/13598044/ssc_1105.zip)

-------------------------------------------------------------------
compute fail(hybrid): pvwattsv8 execution error.
	fail(tilt, max=90): 95

	E=90,S=180,W=270
	total system losses
	0=false,1=true
	0=shadingdb,1=average,2=maximum,3=minimum
	0=false,1=true
	0=false,1=true
	0=false,1=true
	0=false,1=true
	'adjust' and 'constant' separated by _ instead of : after SAM 2022.12.21
	'adjust' and 'en_timeindex' separated by _ instead of : after SAM 2022.12.21
	'adjust' and 'en_periods' separated by _ instead of : after SAM 2022.12.21
Simulation Hybrid failed :Simulation time step is 60 minutes for generic_system.

----------------------------------------------------------------------------------------------

With SSC_WARNING and SSC_ERROR, cmod_hybrid reports the "standard" messages to the simulation dialog (even though "Finished with notices." is somewhat misleading.:
image

The SSC_NOTICE messages should be reported on the Notices tab in SAM.

### DIFF
--- a/ssc/cmod_hybrid.cpp
+++ b/ssc/cmod_hybrid.cpp
@@ -51,6 +51,25 @@ public:
     {
         add_var_info(_cm_vtab_hybrid);
     }
+
+    void ssc_module_exec_with_error(ssc_module_t module, var_table& input, std::string compute_module) {
+        if (!ssc_module_exec(module, static_cast<ssc_data_t>(&input))) {
+            std::string str = std::string(compute_module) + " execution error.\n";
+            int idx = 0;
+            int type = -1;
+            while (const char* msg = ssc_module_log(module, idx++, &type, nullptr))
+            {
+                if (/*/(type == SSC_NOTICE) || */(type == SSC_WARNING) || (type == SSC_ERROR)) {
+                    str += "\t";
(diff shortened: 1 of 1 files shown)


mjprilliman (approved):
This resolved some of the debug issues I was having earlier. 

image

brtietz (approved):
Works for me, thank you!

PR #1113, opened 2023-12-12
fix adjustment factor use for pvwatts

### LINKED ISSUE #1111: Adjustment factors for Non-annual PV simulations
**Describe the bug**
When running a non-annual simulation where the weather data doesn't start at timestep 0, such as in https://github.com/NREL/pysam/blob/main/Examples/NonAnnualSimulation.ipynb

the adjustment factors are not properly aligned and result in a adjustment factor of 0, zeroing out all ac power as in the example above.

The length of the adjustment factors vector is the length of the weather file (810), but when accessing the adjustment factor, the timestep is used (3096) and since this is larger than the length, the default of 0 is returned.

**To Reproduce**
1. Run the above notebook and see that the prediction line at the bottom is 0

**Expected behavior**
Adjustment factor should be aligned

dguittet:
fix #1111


mjprilliman:
Nice catch! This issue is probably more widespread than just pvwatts right? Did you try something similar with pvsamv1 for example?

dguittet:
@mjprilliman I modified (locally) the notebook to use pvsamv1 and didn't run into the same project with adjustment factors zeroing out AC, so they must be handled differently in some way...

janinefreeman:
See my comment on #1111 - I don't think this fixes the root cause of the issue

mjprilliman:
Do we need to do anything here for Patch 1?

janinefreeman (approved):

PR #1114, opened 2023-12-12
Add check for crit load unmet to self consumption metrics

### LINKED ISSUE #1110: Self consumption mode counts all grid outage steps as "system met load"
**Describe the bug**
Turning on a grid outage with self consumption dispatch causes all of the steps in the grid outage to count towards steps where the system met the load. A fully off-grid system would look like it met load 100% of the time with this system, even if there is critical load unmet.

**To Reproduce**
1. Run both cases in the attached SAM file: 
[grid_outage_self_consumption_test.zip](https://github.com/NREL/ssc/files/13641979/grid_outage_self_consumption_test.zip)
2. Note the grid_outage has higher number of steps where the load is met
3. Note the critical load is unmet during the outage.

**Expected behavior**
While these metrics don't have exactly the same definition (i.e. critical load may be < 100% of load), it seems reasonable that steps in which critical load is unmet shouldn't count towards the self-consumption metrics.

Suggested fix is to add a check for critical load unmet (if there's a grid outage) to https://github.com/NREL/ssc/blob/75b4b1778d88234149bbb693d72914c2293ee721/ssc/cmod_battery.cpp#L2114

 - Version 2023.12.17

brtietz:
Fixes #1110, see issue for test instructions.

Due to the differences in battery dispatch, the grid outage case has 2 more steps where the system meets load than the non outage version.

Still open to discussing whether this is on-time for the release or should wait for patch one, hence the lack of milestone tag.

### DIFF
--- a/ssc/cmod_battery.cpp
+++ b/ssc/cmod_battery.cpp
@@ -2105,13 +2105,23 @@ void battstor::calculate_monthly_and_annual_outputs(compute_module& cm)
 
         if (batt_vars->batt_dispatch == dispatch_t::SELF_CONSUMPTION)
         {
+            std::vector<double> crit_load_unmet;
+            if (cm.is_assigned("crit_load_unmet")) {
+                crit_load_unmet = cm.as_vector_double("crit_load_unmet");
+            }
+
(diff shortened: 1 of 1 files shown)


janinefreeman (approved):
Looks good, thank you!

PR #1125, opened 2024-01-25
Fix reporting of ground reflected rear irradiance gains in loss diagram

mjprilliman:
-Individual components of bifacial rear irradiance gain in PV loss diagram were reporting wrong values due to clear sky calls of function calculating rear irradiance being made after original call. 
-Update variable declarations to not overwrite bifacial irradiance outputs
-No changes to bifacial energy calculations, only loss diagram outputs


cpaulgilman (approved):
Looks good now.

Some test comparisons based on Detailed PV / No Financial with bifacial module enabled, "Ground reflected to rear" reported on Losses tab:

SAM 2022.11.21: 6.181%
SAM 2023.12.17: 0.142%
SAM Open Source ssc-1124-fix: 6.399%

PR #1126, opened 2024-01-25
Use calculated DNI for Allen method when there is no DNI in weather file

### LINKED ISSUE #1123: Allen model not working when weather file has no DNI
**Describe the bug**
Allen method is using the weather file DNI in calculation, is giving erroneous results without warning when there is no DNI in the weather file. 

**Expected behavior**
Either use decomposed DNI in calculation when no weather file DNI is provided or error out of the simulation with an appropriate error message. 

**Screenshots**

**Additional context**

mjprilliman:
-Fixes #1123 
-Fixes https://github.com/NREL/SAM/issues/1685
-Check for nan for weather file DNI, use calculated DNI as alternative

### DIFF
--- a/ssc/cmod_pvsamv1.cpp
+++ b/ssc/cmod_pvsamv1.cpp
@@ -1974,6 +1974,10 @@ void cm_pvsamv1::exec()
                         if (!isnan(Irradiance->p_weatherFileDNI[idx])) {
                             PVSystem->p_DNIIndex[nn][idx] = (ssc_number_t)(Irradiance->p_weatherFileDNI[idx] / dni_cs);
                         }
+                        else if (isnan(Irradiance->p_weatherFileDNI[idx]) && (radmode == irrad::DN_DF || radmode == irrad::DN_GH))
+                        {
+                            PVSystem->p_DNIIndex[nn][idx] = 0; //Simulation should exit when looking for DNI but no DNI found
+                        }
                         else {
                             PVSystem->p_DNIIndex[nn][idx] = (ssc_number_t)(Irradiance->p_IrradianceCalculated[2][idx] / dni_cs);
                         }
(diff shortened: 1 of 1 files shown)


cpaulgilman (changes requested):
The simulation runs without crashing now, but the subhourly clipping loss seems very high. This is for the default Detailed PV / No Financial model with the weather file attached below:

image

Weather file used for testing:

[38.13_-86.34_psm3-tmy_60_tmy_no_dni.zip](https://github.com/NREL/ssc/files/14070083/38.13_-86.34_psm3-tmy_60_tmy_no_dni.zip)

mjprilliman:
> The simulation runs without crashing now, but the subhourly clipping loss seems very high. This is for the default Detailed PV / No Financial model with the weather file attached below:
> 
> image
> 
> Weather file used for testing:
> 
> [38.13_-86.34_psm3-tmy_60_tmy_no_dni.zip](https://github.com/NREL/ssc/files/14070083/38.13_-86.34_psm3-tmy_60_tmy_no_dni.zip)

Which Irradiance data option did you use for these results? I'm having trouble replicating results.

mjprilliman:
> I'm not seeing those results now. Perhaps I was in the wrong version of SAM. Please ignore my previous comment.
> 
> However, I am getting an exception in cmod_pvsamv1.cpp when I run a simulation with the default DNI+DHI option with the weather file that has no DNI data. The release version of SAM 2023.12.17 correctly generates an SSC error message, so I think this may have been introduced in this branch:
> 
> image

Paul, I caught that error too, apologies for not updating. I am working on a fix and will push shortly.

cpaulgilman:
That will fix https://github.com/NREL/SAM/issues/1685

cpaulgilman (approved):
Looks good now. I tested weather file with no DNI with and without subhourly clipping. I'll work on the UI change for https://github.com/NREL/SAM/issues/1666 next.

PR #1128, opened 2024-02-13
Expose snow slide coefficient in ssc

mjprilliman:
-Add option to change snow slide coefficient in snow model in SDK
-Previously fixed to 1.97
-Fixes https://github.com/NREL/SAM/issues/1547


janinefreeman (commented):
This looks like it should work, but the Ubuntu build is failing and I can't get the tests to build on my machine to run them right now (my problem, not a problem with this PR)...

mjprilliman:
> This looks like it should work, but the Ubuntu build is failing (I see your latest commit might fix that though) and I can't get the tests to build on my machine to run them right now (my problem, not a problem with this PR)...

The build failed in Windows as well, it just didn't show in the editor until I compiled. I updated it to have the default value in the function call rather than the blank space which should work, I'll let the tests build to see.

janinefreeman (approved):

PR #1130, opened 2024-02-15
Me wave file reader

mjprilliman:
-See https://github.com/NREL/SAM/pull/1701
-Use flexible header keys to fill information from resource file headers


mjprilliman:
-See #1133 
-Updates wave and tidal modules to be flexible for hourly, subhourly resource data (in addition to 3 hour data for wave)

PR #1132, opened 2024-02-19
Address confusion and consistency in powerflow variables

brtietz:
- Use same tolerance in powerflow code and utility rate code, which both generate "electricity from grid to x varaibles"
- Rename gen_purchases so it's clear that it does not include load it btm models

Fixes https://github.com/NREL/SAM/issues/1684, see the issue for test file.


sjanzou (approved):
Tested with files from https://github.com/NREL/SAM/issues/1684

image

PR #1135, opened 2024-02-23
Fix subhourly indexing error in retail rate dispatch forecast function

### LINKED ISSUE #1134: Battery retail rate dispatch with TOU demand charges causes simulation error.
For description and example file, see https://sam.nrel.gov/forum/forum-general/4540

Requires downloading a 15-minute weather file using Advanced Download option.

brtietz:
Fixes #1134 by fixing an indexing error in the forecast setup function. Also adds new tests for subhourly demand charge data.

Test by running the file from the forum and confirming that the simulation completes without errors.


cpaulgilman (approved):
Looks good. Thanks for fixing.

PR #1136, opened 2024-02-23
Update debt fraction limits code to only lend when projects have a possitive ebitda.

brtietz:
Change != to > for debt fraction sizing code. Ensures that projects with negative EBITDA get zero debt, even in IRR target mode. Additionally, ensure minimum debt fraction is non-zero to ensure the relevant code always runs. If users want zero debt, they can specify it in debt percent mode.

My opinion is that if users want negative DSCR or other strange behavior, they can uncheck this box, but happy to discuss. 

Test file available in the issue. Fixes: https://github.com/NREL/SAM/issues/1353


sjanzou (commented):
For consistency, we should update the other occurrences of "if (dscr!=0)" :
image

Also, (maybe a separate issue) we should update cmod_communitysolar to have the same max debt fraction checking and to have the dscr>0 check
image

brtietz:
> For consistency, we should update the other occurrences of "if (dscr!=0)" : image

That seems worth some discussion. If the debt fraction limits box is unchecked, I was under the impression we allowed negative DSCRs. Similarly, if a debt percent is specified, a negative DSCR is a valid outcome of this calculation. I'll add to the SAM team meeting agenda to discuss.

> Also, (maybe a separate issue) we should update cmod_communitysolar to have the same max debt fraction checking and to have the dscr>0 check image

I had filed https://github.com/NREL/SAM/issues/1708 for that, thanks for the reminder on community solar. I had that tagged for the fall release, but happy to discuss which portions are patchable versus requiring a release.

brtietz:
Decision at 2/27 SAM meeting:
- Negative minimum DSCRs should be allowed for the debt % option
- If the maximum debt fraction box is unchecked, give a warning to the user if the debt percent goes negative.
- Create an issue (or see if another issue exists) to revisit the negative debt percent behavior.

cpaulgilman:
Links to related issues now closed:
* https://github.com/NREL/SAM/issues/551
* https://github.com/NREL/SAM/issues/534

brtietz:
@cpaulgilman Thanks, I added debt percent < 0 check to the check functions from https://github.com/NREL/ssc/pull/903/files. This is ready for a second review.

cpaulgilman (approved):

Issue #1047, opened 2023-09-13
IBI and CBI double counting in financial models


sjanzou:
The investment based and capacity based incentives are accounted for in the cost installed (subtracted) value and then two values are considered issuance of equity and purchase of property which are based on the cost installed. These values are used to calculate the investing activities and financial activities which according to the original models should not have the IBI and CBI effects. 

The IBI and CBI reductions are included in the investing activities (correct) and already included in the financial activities in the issuance of equity (and then added again) which gives a pretax project return the includes the IBI and CBI, which is incorrect because the IBI and CBI were already accounted for in the investing activities which leads to a double counting.

The fix to this issue includes:
1. consistent definitions of issuance of equity for all financial models
2. removal of the IBI and CBI from the financing activities 

Additional details and code references included in the attached Word Document.
[FinancialMarketsIssues2023.9.6 IBI.docx](https://github.com/NREL/ssc/files/12594551/FinancialMarketsIssues2023.9.6.IBI.docx)

PR #1141, opened 2024-02-29
Update to address SAM issue 1477 for host developer only

sjanzou:
I think we should review in detail before proceeding with host developer and other markets.

See https://github.com/NREL/SAM/issues/1722


sjanzou:
Test file 
[ITC_basis_reduction.zip](https://github.com/NREL/ssc/files/14447392/ITC_basis_reduction.zip)

Test with no IBI
image

Test with IBI with no basis reduction
image

Test with IBI with federal and state basis reduction
image

Test with federal and state taxable IBI and federal and state basis reduction
image

cpaulgilman (commented):
See my comment https://github.com/NREL/SAM/issues/1477

brtietz:
The ITC amounts now have the behavior that I'd expect, but I'm a little concerned that IBI isn't showing up in purchase of property:

image

Would some of the test processes from https://github.com/NREL/ssc/issues/1047 be relevant here?

sjanzou:
> The ITC amounts now have the behavior that I'd expect, but I'm a little concerned that IBI isn't showing up in purchase of property:
> 
> image
> 
> Would some of the test processes from #1047 be relevant here?

I think much more testing is in order here and for all of the financial markets.

I am leaning towards moving issue https://github.com/NREL/SAM/issues/1477 and https://github.com/NREL/SAM/issues/1722 to patch 2 - any objections?

brtietz:
Closing PR as we're addressing this in https://github.com/NREL/ssc/pull/1178

PR #1149, opened 2024-03-11
Add ssc_data_deep_copy

dguittet:
Minor update to sscapi needed for PySAM Hybrids, to enable deep copy of ssc_data_t.

Also remove filenames from input JSONs


ssc/sscapi.cpp
```diff
@@ -712,6 +712,16 @@ SSCEXPORT ssc_var_t ssc_data_get_data_matrix(ssc_data_t p_data, const char *name
     }
     return dat;
 }
+
+SSCEXPORT ssc_bool_t ssc_data_deep_copy(ssc_data_t source, ssc_data_t dest) {
+    auto source_vt = static_cast<var_table*>(source);
+    if (!source_vt) return 0;
+    auto dest_vt = static_cast<var_table*>(dest);
+    if (!dest_vt) return 0;
+	*dest_vt = *source_vt;  // invokes operator= for deep copy
```
sjanzou, line 721: @dguittet, have you run into any issues with the "=" copy? Do we need a try...catch here?

sjanzou (approved):
Looks good! Maybe the try...catch on the deep copy?

ssc/sscapi.cpp
dguittet: I haven't, as I think the static_cast error checks should handle all the issues.

The `operator=` function doesn't have any exceptions to throw: https://github.com/NREL/ssc/blob/028722445535be16d7f4089b09afeffca07f583e/ssc/vartab.cpp#L220

So if we caught anything, it would be a native C++ error. I think like that error message should be passed as-is, instead of being caught and reinterpreted.

sjanzou: > I haven't, as I think the static_cast error checks should handle all the issues.
> 
> The `operator=` function doesn't have any exceptions to throw:
> 
> https://github.com/NREL/ssc/blob/028722445535be16d7f4089b09afeffca07f583e/ssc/vartab.cpp#L220
> 
> So if we caught anything, it would be a native C++ error. I think like that error message should be passed as-is, instead of being caught and reinterpreted.

Sounds good. Thanks for the detailed explanation.

sjanzou (approved):

PR #1158, opened 2024-04-18
Clear var_data before copy

### LINKED ISSUE #1157: Bug copying `vec` field of var_data
**Describe the bug**
In the function copy for var_data, the vec field is appending entries from the `rhs.vec` instead of replacing.
https://github.com/NREL/ssc/blob/217296f1a92d1a37ccf1600de994664ee216ffa2/ssc/vartab.h#L152

**Expected behavior**
The old vec data should be cleared and the new one copied over.

dguittet:
fix #1157

### DIFF
--- a/ssc/vartab.h
+++ b/ssc/vartab.h
@@ -144,6 +144,7 @@ public:
 
 	var_data &operator=(const var_data &rhs) { copy(rhs); return *this; }
 	void copy( const var_data &rhs ) {
+        clear();
 	    type=rhs.type;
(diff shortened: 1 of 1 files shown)


cpaulgilman (approved):
This fixed the issue I was having with PySAM errors while testing hybrid system models: `Exec fail(hybrid): Only one battery bank allowed at this time` and `UnboundLocalError: cannot access local variable 'unassigned'`.

PR #1159, opened 2024-04-26
C++17 std::byte

dguittet:
After upgrading the code to C++17 for OR-Tools compatibility, an issue shows up in several of the CSP cpp files.

image

According to https://developercommunity.visualstudio.com/t/error-c2872-byte-ambiguous-symbol/93889, this happens even in cases where std::byte isn't used. And it wouldn't be used in SAM code because it was introduced in C++17.

The fix that we can apply for the SAM code for now, until it's fixed in the C++ / VS Code side of things is:

> #define _HAS_STD_BYTE 0
> By doing so, you are effectively disabling the ‘byte’ type that was implemented in VS 2017. Again, this method will only work if you do not use any instances of ‘byte’ throughout your code


sjanzou (changes requested):
Looks like OR-tools was built as an x86_64 library only and the GitHub Actions is failing because the arm64 universal library is missing... Can you rebuild or-tools as a universal (X86_64 and arm64) library?

sjanzou (approved):
Looks good for Apple Silicon...

Would it be possible to setup an Intel Mac runner (in addition to the existing arm64 runner) for future pull requests?

PR #1173, opened 2024-06-05
Add dt_hr options

### LINKED ISSUE #1172: BatteryStateful set_state allow dt_hr input
**Describe the bug**
In the use case where the battery state is being changed back and forth and the timestep is being changed, to reset the battery to some stored state, the `dt_hr` from the battery parameters is required. Allow battery::set_state to take in an optional dt_hr input.

https://github.com/NREL/EnergyPlus/pull/10543

dguittet:
fixes #1172

### DIFF
--- a/shared/lib_battery.cpp
+++ b/shared/lib_battery.cpp
@@ -779,8 +779,10 @@ battery_state battery_t::get_state() { return *state; }
 
 battery_params battery_t::get_params() { return *params; }
 
-void battery_t::set_state(const battery_state& tmp_state) {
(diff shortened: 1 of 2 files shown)


brtietz (commented):
Can you add some documentation as to why a user would want to call this instead of ChangeTimestep? Is this just to reduce function calls?

dguittet:
@brtietz Added, let me know if that comment doesn't explain it

brtietz (approved):
Thanks for adding the comment!

PR #1176, opened 2024-06-20
Address SAM issue #1733 while keeping functionality of SAM issue #1184

sjanzou:
To test:
1. pull SAM_1733 branch of ssc and patch branches of other repos and build
2. create PVWatts/Commercial case
3. Financial parameters page - set straight line depreciation years to 0
4. select straight line depreciation and run - error as expected
5. select different depreciation method and run - no error as expected.

See https://github.com/NREL/SAM/issues/1733 for simulation error in current release.

### DIFF
--- a/ssc/common.cpp
+++ b/ssc/common.cpp
@@ -171,10 +171,10 @@ var_info_invalid
 
 var_info vtab_depreciation[] = {
 { SSC_INPUT,SSC_NUMBER  , "depr_fed_type"                        , "Federal depreciation type"                                      , ""                                       , "0=none,1=macrs_half_year,2=sl,3=custom", "Depreciation"         , "?=0"            , "INTEGER,MIN=0,MAX=3"   , ""},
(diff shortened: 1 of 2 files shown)


cpaulgilman (changes requested):
I'm still seeing the error after cleaning and rebuilding. This is for the default PVWatts / Commercial case switching to 5-yr MACRS after setting straight line depreciation years to zero:

image

mjprilliman (approved):
Worked for me when testing. Where else should SKIP_CONSTRAINT_CHECK be applied (question for a different issue/PR)?

I can't replicate the Windows test failure on my build.

sjanzou (commented):
reviewed and tests with @mjprilliman

sjanzou:
> Worked for me when testing. Where else should SKIP_CONSTRAINT_CHECK be applied (question for a different issue/PR)?
> 
> I can't replicate the Windows test failure on my build.

SKIP_CONSTRAINT_CHECK should be used anywhere for an optional ssc variable that is required only if some other variable is set to a specific value. The example of SAM issue https://github.com/NREL/SAM/issues/1733 is a perfect example in which the optional "years" value for straight line depreciation is required only if the depreciation selection is 2. Then the "years" input for the user should be checked to be positive (the CONSTRAINT field in the vartable).

PR #1177, opened 2024-06-25
Ci runners

dguittet:
Cache GTEST and update versions


sjanzou (approved):
This is great!

Question about the macos runners - should we be using macos-latest-large (intel) and macos-latest-xlarge (arm64) for future runners? 

The macos-latest seems subject to change...

https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners
image

brtietz (approved):

PR #1178, opened 2024-06-25
1477 itc incentive interactions

brtietz:
Transition from using cost_installed to new variables for depreciation and ITC calculations. Feedback from tax experts included:

- Basis should include pre-financing costs, debt fees, and construction interest, but not reserves
- Investment and capacity based incentives should only be subtracted if "reduces basis" box is checked

New output variables in the data tables for PVWatts default:

image

Note: SAM default test results should be updated prior to merge. Decide whether we want to do a SAM PR for just that, or wait for @cpaulgilman to further update the cash flow.

https://github.com/NREL/sam/issues/1477


sjanzou (approved):
Good job updating all the tests...

For the following outputs in SAM PVWatts/Host Developer:
image

Should we update the depreciation and itc tables in SAM to reflect the additional values
image

maybe a separate pull requests with the updated defaults??

Also, are the new depreciation and ITC variables applicable to all the FOM financial markets? i.e. https://github.com/NREL/SAM/issues/1722

brtietz:
@sjanzou and @cpaulgilman 

I'm tempted to remove the debt fees from the basis as well. Per https://www.irs.gov/pub/irs-drop/n-13-70.pdf (question 14) fees cannot be included in the basis. This is referring to the residential section of the code, and while I haven't found as strong of a reference for businesses, I haven't found anything that conflicts and says that they are allowed. It seems unlikely that a fee is considered "solar energy property" for these purposes.

However, users do have the option to adjust for this on the depreciation page, and that page says these fees are included:
image

1. What are your preferences on including or removing these fees from the new basis variables?
2. Do we have a way to notify users that the basis has changed during the patch and they may need to re-check their depreciation allocations? If not, I'm wondering if it makes sense to defer some of these changes to the release.

sjanzou:
> @sjanzou and @cpaulgilman
> 
> I'm tempted to remove the debt fees from the basis as well. Per https://www.irs.gov/pub/irs-drop/n-13-70.pdf (question 14) fees cannot be included in the basis. This is referring to the residential section of the code, and while I haven't found as strong of a reference for businesses (https://zenergyfin.com/solar-itc-table-eligible-and-ineligible-costs/ is the best to date), I haven't found anything that conflicts and says that they are allowed. It seems unlikely that a fee is considered "solar energy property" for these purposes.
> 
> However, users do have the option to adjust for this on the depreciation page, and that page says these fees are included: image
> 
> 1. What are your preferences on including or removing these fees from the new basis variables?
> 2. Do we have a way to notify users that the basis has changed during the patch and they may need to re-check their depreciation allocations? If not, I'm wondering if it makes sense to defer some of these changes to the release.

@brtietz , @cpaulgilman To be totally transparent, a checkbox could be used on whether or not to include the fees...However, this would require waiting until the next release. If we wait until the next release, we can notify the user of the bases changes in the Version upgrade script. What do you think?

brtietz:
> @brtietz , @cpaulgilman To be totally transparent, a checkbox could be used on whether or not to include the fees...However, this would require waiting until the next release. If we wait until the next release, we can notify the user of the bases changes in the Version upgrade script. What do you think?

I like this idea! I'll leave the fees in for the patch in the spirit of making smaller changes, but I still want to remove the reserves at Michele's recommendation which means notifying users (if possible) is still a good idea.

brtietz:
@cpaulgilman This should be in a state where it's worth starting the cash flow and send to excel updates.

I've run out of time to get the tests updated, so those are expected to fail based on the changes to reserves in the basis. I'll get those updated the week of July 8th.

cpaulgilman:
@brtietz On the question of fees, you wrote:

> I'm tempted to remove the debt fees from the basis as well. Per https://www.irs.gov/pub/irs-drop/n-13-70.pdf (question 14) fees cannot be included in the basis. This is referring to the residential section of the code, and while I haven't found as strong of a reference for businesses (https://zenergyfin.com/solar-itc-table-eligible-and-ineligible-costs/ is the best to date), I haven't found anything that conflicts and says that they are allowed. It seems unlikely that a fee is considered "solar energy property" for these purposes.

But also:

> Transition from using cost_installed to new variables for depreciation and ITC calculations. Feedback from tax experts included:
>
> *  Basis should include pre-financing costs, debt fees, and construction interest, but not reserves
> *  Investment and capacity based incentives should only be subtracted if "reduces basis" box is checked

Did we get feedback from tax experts that contradicts the information you found in the IRS and xenergyfin.com documents?

brtietz:
> Did we get feedback from tax experts that contradicts the information you found in the IRS and zenergyfin.com documents?

No, I wrote the paragraph at the top of the pull request with incomplete information. This document also suggests debt fees should be excluded from the basis: https://www.irs.gov/publications/p551

I think the main argument for leaving the fees in for this PR is to make smaller changes in a patch and not force users to redo their depreciation allocations. Happy to do more research and figure out whether the checkbox is necessary or if we should just exclude the fees in the release.

cpaulgilman:
>I think the main argument for leaving the fees in for this PR is to make smaller changes in a patch and not force users to redo their depreciation allocations.

Unless we find documentation or examples where fees are considered property for the purpose of depreciation, I think I would vote to not have check boxes. As discussed above, this will cause results to change, so we will have to manage that via documentation (and version upgrade script if we wait for the fall release).

sjanzou (approved):
The additional test coverage looks good for IBI and CBI. We will have to revisit when SAM PR 1805 is reopened for the Fall Release.

cpaulgilman (approved):
This should be good to go after https://github.com/NREL/SAM/pull/1808 is reviewed and approved.

Issue #1143, opened 2024-03-02
Wind- some annual outputs incorrect for subhourly


janinefreeman:
Some annual outputs in the wind compute module are incorrect for subhourly simulations- annual gross energy, possibly the annual wake loss percent, cutoff losses. The kW values were not being scaled for timestep to get to hourly values, which means the outputs will be correct for hourly simulations but not subhourly. This is likely a holdover from the wind model being hourly originally.

PR #1186, opened 2024-07-21
New Wind Features 2024

janinefreeman:
This pull request adds a variety of new features to the wind model, currently accessible only programmatically.

New features include:
 - additional wake loss outputs, including a timeseries wake loss output (available in the UI as well) <- @cpaulgilman these may need to be documented
  --also separates "internal" wake losses (i.e. calculated by a wake model) from external wake losses (i.e. from other plants)
 - new optional input to the Park wake model to overwrite the calculated coefficient of thrust
 - new optional input to use a different wake decay constant in the Park model
 - an optional wake loss multiplier to proportionally change the total annual wake losses calculated by the compute module
 - optional "icing persistence" input, which allows the user to specify the number of timesteps that an icing shutdown should last (more realistic behavior than assuming the turbine can start/stop as conditions change)

Tests have been added to cover these new features.

Bug fix for #1143 included in this PR.

@dguittet I think you're the best reviewer for this, but note that this is not an urgent PR. :) I did not generate the the new API files for these new inputs, but maybe I can get your help with that when you have some time?


ssc/cmod_windpower.cpp
```diff
@@ -285,6 +303,17 @@ void cm_windpower::exec()
...
 
+    // get optional thrust curve for wind turbine
+    if (is_assigned("wind_turbine_ct_curve"))
+    {
+        size_t ctCurveLength = 0;
+        ssc_number_t* ctc = as_array("wind_turbine_ct_curve", &ctCurveLength);
+        std::vector<double> ct_curve(ctCurveLength);
+        for (size_t i = 0; i < ctCurveLength; i++)
+            ct_curve[i] = ctc[i];
+        wt.setCtCurve(ct_curve);
```
dguittet, line 314: Should there be an error or warning returned to the user if this function returns 0?

ssc/cmod_windpower.cpp
```diff
@@ -377,26 +408,35 @@ void cm_windpower::exec()
...
         wakeModel = std::make_shared<eddyViscosityWakeModel>(eddyViscosityWakeModel(wpc.nTurbines, &wt, as_double("wind_resource_turbulence_coeff")));
     }
-    else if (wakeModelChoice == 3)
+    else if (wakeModelChoice == CONSTANTVALUE)
     {
-        wake_int_loss_percent = as_double("wake_int_loss");
-        wakeModel = std::make_shared<constantWakeModel>(constantWakeModel(wpc.nTurbines, &wt, (100. - wake_int_loss_percent)/100.));
+        annual_wake_int_loss_percent = as_double("wake_int_loss");
+        annual_wake_int_loss_percent *= is_assigned("wake_loss_multiplier") ? as_double("wake_loss_multiplier") : 1.; // add any wake loss multipliers to the constant loss value here
+        wakeModel = std::make_shared<constantWakeModel>(constantWakeModel(wpc.nTurbines, &wt, (100. - annual_wake_int_loss_percent)/100.));
```
dguittet, line 439: Am I understanding that for the constant wake loss model you can provide both the constant and some multiple of the constant? e.g. loss = a * b? 

I think it would make more sense to just provide one value, and make "wake_loss_multiplier" not applicable for wakeModelChoice == CONSTANTVALUE

test/ssc_test/cmod_windpower_test.cpp
```diff
@@ -132,10 +132,90 @@ TEST_F(CMWindPowerIntegration, WakeModelsUsingFile_cmod_windpower) {
...
+    // get new wake loss
+    compute();
+    ssc_data_get_number(data, "annual_wake_loss_internal_percent", &wakeLoss2); //get the wake loss with Ct curve
+
+    ssc_number_t difference = wakeLoss2 - wakeLoss1;
+    EXPECT_NEAR(wakeLoss1, 4.148, 0.01) << "Wake Loss 1";
+    EXPECT_NEAR(wakeLoss2, 5.562, 0.5) << "Wake Loss 2";
+    EXPECT_NEAR(difference, 1.413, 0.5) << "Difference";
+}
+
```
dguittet, line 218: Can you please edit "IcingAndLowTempCutoff_cmod_windpower" test to add the icing persistence feature?

dguittet (changes requested):

test/ssc_test/cmod_windpower_test.cpp
janinefreeman: I had written this test, but hadn't pushed it. Done!

ssc/cmod_windpower.cpp
janinefreeman: You're right, that would be pretty confusing. Updated!

ssc/cmod_windpower.cpp
```diff
@@ -412,8 +464,12 @@ void cm_windpower::exec()
             throw exec_error("windpower", wpc.GetErrorDetails());
         }
 
-        if (wakeModelChoice != 3)
-            wake_int_loss_percent = (1. - farmPower / farmPowerGross) * 100.;
+        if (wakeModelChoice != CONSTANTVALUE) //we already threw an error if a wake loss multiplier was specified, so may just proceed
+        {
+            assign("annual_wake_loss_internal_kWh", var_data((ssc_number_t)(farmPowerGross - farmPower)));
```
dguittet, line 469: shouldn't the loss multiplier be applied here, similar to below?

```
            if (wakeModelChoice != CONSTANTVALUE && is_assigned("wake_loss_multiplier")) //wake loss multiplier isn't available for constant wake loss option
            {
                double wakeLossMultiplier = as_double("wake_loss_multiplier");
                double wakeLossBeforeMultiplier = gross_farmp - farmp;
                double newWakeLoss = wakeLossBeforeMultiplier * wakeLossMultiplier;
                farmp = gross_farmp - newWakeLoss;
            }
```

janinefreeman: Yup, you're right. Good catch!

dguittet (approved):
Thanks!

PR #1187, opened 2024-07-23
Update REopt API

cpaulgilman:
Fixes https://github.com/NREL/SAM/issues/1742

* Set REopt input `net_metering_limit_kw` to system capacity when net metering is enabled.
* Change REopt solver from HiGHS to SCIP to allow large numbers for tier limits (1e38).
* Specify only `max_kw` instead of both `existing_kw` and `max_kw`.

#### To test:

1. Build a) SAM-private or b) SAM open source with NREL developer key and email in private.h.

2. Create a default PV Battery / Residential and PV Battery / Commercial case and click **Get size and dispatch** on Battery Cell and System page. REopt should return non-zero battery size and dispatch array.

Note that REopt API call and optimization takes a while -- I'm not sure if the SCIP solver is slower than HiGHS. This is something we may want to investigate for the Fall 2024 release.


ssc/cmod_battery_eqns.cpp
```diff
@@ -258,9 +258,11 @@ bool Reopt_size_standalone_battery_params(ssc_data_t data) {
             }
         }
         reopt_load.assign("critical_loads_kw", var_data(&vec[0], vec_size));
+        reopt_load.assign("year", 2018); // recent common year starting on Monday
     }
 
     reopt_settings.assign_match_case("time_steps_per_hour", var_data((int)(sim_len / 8760)));
+    reopt_settings.assign("solver_name", var_data("SCIP")); // "HiGHS" option does not work with large numbers like 1e38 for tier max values per https://github.com/NREL/SAM/issues/1742
```
sjanzou, line 266: interesting to note for OR tools inclusion in ssc for the Fall release!

sjanzou (approved):
Looks good... the Windows runners seem to have intermittent issues with the test results. I restarted the failed job.

PR #1189, opened 2024-07-26
Getem cost updates

mjprilliman:
See https://github.com/NREL/SAM/pull/1823


sjanzou (approved):
Thanks for updating the tests!

PR #1191, opened 2024-07-26
Add drilling cost curves to geothermal cost cmod

mjprilliman:
-Add drilling cost curves to cmod for use in reV calculations
-Optional inputs to model, not used by default
-Re-evaluate for release


shared/lib_geothermal.cpp
```diff
@@ -1567,6 +1567,7 @@ double CGeothermalAnalyzer::GetNumberOfWells(void)
         }
 
         if (mp_geo_out->md_NumberOfWellsInj < 0) mp_geo_out->md_NumberOfWellsInj = 0;
+        if (mp_geo_out->md_NumberOfWells < 0) mp_geo_out->md_NumberOfWells = 0;
```
sjanzou, line 1570: Both injection and regular wells drilled - good catch!

sjanzou (approved):
What do we need to re-evaluate before the release?

ssc/cmod_geothermal_costs.cpp
```diff
@@ -75,9 +75,26 @@ static var_info _cm_vtab_geothermal_costs[] = {
...
+        { SSC_INPUT,        SSC_NUMBER,      "resource_depth",                     "Resource Depth",                               "m",              "",             "GeoHourly",        "",                        "",                "" },
+        { SSC_INPUT,        SSC_NUMBER,      "geotherm.cost.prod_wells_drilled",                      "Number of drilled production wells",                      "0/1",             "0=LargerDiameter,1=SmallerDiameter",             "GeoHourly",        "",                        "",                "" },
+        { SSC_INPUT,        SSC_NUMBER,      "geotherm.cost.inj_wells_drilled",                      "Number of drilled injection wells",                      "0/1",             "0=LargerDiameter,1=SmallerDiameter",             "GeoHourly",        "",                        "",                "" },
+        { SSC_INPUT,        SSC_NUMBER,      "calc_drill_costs",                      "Calculate drill costs",                      "0/1",             "0=LargerDiameter,1=SmallerDiameter",             "GeoHourly",        "?=0",                        "",                "" },
```
dguittet, line 88: curious why you are using "." instead of "_"?

ssc/cmod_geothermal_costs.cpp
```diff
@@ -295,6 +312,146 @@ class cm_geothermal_costs : public compute_module
 	void exec() override
 	{
 		SGeothermal_Inputs geo_inputs;
+
+        int calc_drill_costs = as_integer("calc_drill_costs");
+        if (calc_drill_costs == 1) {
+            int inj_cost_curve_welldiam = as_integer("geotherm.cost.inj_cost_curve_welldiam");
```
dguittet, line 318: shouldn't these inputs be required if "calc_drill_costs"==1?

ssc/cmod_geothermal_costs.cpp
mjprilliman: We've just been working off what was there before. It's probably worth changing this in the future.

ssc/cmod_geothermal_costs.cpp
mjprilliman: Yes, I will update this.

cpaulgilman:
https://github.com/NREL/SAM/pull/1824

PR #1192, opened 2024-07-29
Addresses SAM issue 1224

sjanzou:
goes with https://github.com/NREL/wex/pull/175


brtietz:
An example from: https://hub.alfresco.com/t5/alfresco-content-services-blog/tldr-the-lgpl-license-explained/ba-p/314998

> A good way to do that is to add a short text in the readme and in the file headers, something like: "Please notice that this ‘foo’ program is a modified version of program ‘bar’ version x.x.x", with links and details to the original version. It must be easy for the end users to understand. 

I didn't see a readme in lpsolve - is it just V5?

sjanzou:
> The changes themselves look good, but we need some additional changes to be compliant with nlopt and lpsolve's licenses. Specifically the bolded line below:
> 
> > 2. You may modify your copy or copies of the Library or any portion
> >    of it, thus forming a work based on the Library, and copy and
> >    distribute such modifications or work under the terms of Section 1
> >    above, provided that you also meet all of these conditions:
> >    a) The modified work must itself be a software library.
> >    **b) You must cause the files modified to carry prominent notices
> >    stating that you changed the files and the date of any change.**
> >    c) You must cause the whole of the work to be licensed at no
> >    charge to all third parties under the terms of this License.
> >    d) If a facility in the modified Library refers to a function or a
> >    table of data to be supplied by an application program that uses
> >    the facility, other than as an argument passed when the facility
> >    is invoked, then you must make a good faith effort to ensure that,
> >    in the event an application does not supply such function or
> >    table, the facility still operates, and performs whatever part of
> >    its purpose remains meaningful.
> >    (For example, a function in a library to compute square roots has
> >    a purpose that is entirely well-defined independent of the
> >    application.  Therefore, Subsection 2d requires that any
> >    application-supplied function or table used by this function must
> >    be optional: if the application does not supply it, the square
> >    root function must still compute square roots.)
> 
> Can you add painfully obvious comments to the top of each file and the README describing the changes?

Changes added in 

> The changes themselves look good, but we need some additional changes to be compliant with nlopt and lpsolve's licenses. Specifically the bolded line below:
> 
> > 2. You may modify your copy or copies of the Library or any portion
> >    of it, thus forming a work based on the Library, and copy and
> >    distribute such modifications or work under the terms of Section 1
> >    above, provided that you also meet all of these conditions:
> >    a) The modified work must itself be a software library.
> >    **b) You must cause the files modified to carry prominent notices
> >    stating that you changed the files and the date of any change.**
> >    c) You must cause the whole of the work to be licensed at no
> >    charge to all third parties under the terms of this License.
> >    d) If a facility in the modified Library refers to a function or a
> >    table of data to be supplied by an application program that uses
> >    the facility, other than as an argument passed when the facility
> >    is invoked, then you must make a good faith effort to ensure that,
> >    in the event an application does not supply such function or
> >    table, the facility still operates, and performs whatever part of
> >    its purpose remains meaningful.
> >    (For example, a function in a library to compute square roots has
> >    a purpose that is entirely well-defined independent of the
> >    application.  Therefore, Subsection 2d requires that any
> >    application-supplied function or table used by this function must
> >    be optional: if the application does not supply it, the square
> >    root function must still compute square roots.)
> 
> Can you add painfully obvious comments to the top of each file and the README describing the changes?

Header for each file added in https://github.com/NREL/ssc/commit/76d6c00db9125ca4dce352e57e755032e569918f

Readme added with header message for both nlopt and lpsolve

sjanzou:
> An example from: https://hub.alfresco.com/t5/alfresco-content-services-blog/tldr-the-lgpl-license-explained/ba-p/314998
> 
> > A good way to do that is to add a short text in the readme and in the file headers, something like: "Please notice that this ‘foo’ program is a modified version of program ‘bar’ version x.x.x", with links and details to the original version. It must be easy for the end users to understand.
> 
> I didn't see a readme in lpsolve - is it just V5?
> 
> We may need to include a longer history of changes with this using the commit history https://github.com/NREL/ssc/commits/patch/lpsolve

Thanks for noting the licensing issues.

Addressed in https://github.com/NREL/ssc/commit/76d6c00db9125ca4dce352e57e755032e569918f

Review requested.

brtietz (approved):
Thanks for adding the header and the README!

Issue #1190, opened 2024-07-26
Windows test timeout in lib_irradproc.cpp


brtietz:
**Describe the bug**
Windows tests have been timing out due to an initialization error in lib_irradproc.cpp

**To Reproduce**
1. Run CMPvsamv1PowerIntegration_cmod_pvsamv1, NoFinancialModelModuleAndInverterModels on Windows
2. For module_model 4 and inverter_model 0, getGroundShadeFactors has some values with e-67
3. This means that the row to row shading routine never finishes before the tests update

**Expected behavior**
Variables should be properly initialized and the routine should finish on all platforms

Potentially need to initialize ground clearance height as 1.0 in https://github.com/NREL/ssc/blame/724d05aefd7267c66ba8d926bebd8805201ff797/shared/lib_pv_io_manager.cpp#L1321

**Screenshots**
![image](https://github.com/user-attachments/assets/357193e0-95fa-4785-80f7-fd37fe84c265)

![image](https://github.com/user-attachments/assets/33415dd6-1778-4280-8f42-403d47f2fe15)

**Desktop (please complete the following information):**
 - OS: Windows
 - Version Patch as of July 2024

PR #1194, opened 2024-07-29
Initialize ground clearance height for IEC module model

brtietz:
Fixes https://github.com/NREL/ssc/issues/1190

Should be good to go if SSC tests pass


sjanzou (approved):
Thanks for catching this... One of the Windows runners failed - I restarted. This does seem to be a common issue...the windows runner works for the push and then a test fails on the pull request
image

brtietz:
Thanks! I've also seen other tests fail intermittently in lib_utility_rate_test, again only on the windows pull runner. I filed https://github.com/NREL/ssc/issues/1195 and tagged it for the release since intermittent issues are less urgent and take more time.

PR #1203, opened 2024-08-14
Fix REopt API v3 call on MacOS SAM issue #1830

sjanzou:


sjanzou (commented):
Only changes needed for REopt API call to work on MacOS for PV-Batt/Commercial and PV-Batt/Residential

PR #1209, opened 2024-09-26
Update GETEM DT in prod well calculations

mjprilliman:
-Add plant temperature initialization check for ramey wellbore calculation of interface output variables such as brine efficiency
-Update specific heat of water based on reservoir temp at each timestep


shared/lib_geothermal.cpp
```diff
@@ -1160,13 +1160,21 @@ double CGeothermalAnalyzer::RameyWellbore()
...
+        time = mp_geo_out->ElapsedHours * 3600; //elapsed time (s)
+        working_temp = md_WorkingTemperatureC;
+    }
     double utilfactor = 1.0; //capacity factor?
     double avg_gradient = 2 / GetResourceDepthM(); //local average geothermal gradient
     double framey = -1.0 * log(1.1 * (0.3048 * (mo_geo_in.md_DiameterProductionWellInches / (2 * 12)) / sqrt(4.0 * alpharock * time * utilfactor))) - 0.29;
-    double rameyA = productionFlowRate() * md_WorkingTemperatureC * framey / (2 * physics::PI * mo_geo_in.md_EGSThermalConductivity);
-    double ProdTempDrop = -1.0 * ((GetResourceTemperatureC() - md_WorkingTemperatureC) - avg_gradient * (GetResourceDepthM() - rameyA) + (md_WorkingTemperatureC - avg_gradient * rameyA - GetResourceTemperatureC()) * exp(-GetResourceDepthM() / rameyA));
+    double c_w = geothermal::EGSSpecificHeat(working_temp);
+    double rameyA = mo_geo_in.md_ProductionFlowRateKgPerS * c_w * framey / (2 * physics::PI * mo_geo_in.md_EGSThermalConductivity);
```
sjanzou, line 1176: Is there a separate dT for Hydrothermal?

sjanzou (approved):
Just checking in that the updated dT covers all scenarios for Hydrothermal and EGS and binary and flash...

shared/lib_geothermal.cpp
mjprilliman: It works the same for EGS and hydrothermal. I used the existing variable names, I suppose a future item would be to clean up variable names. Thanks for checking!

PR #1212, opened 2024-09-30
PV models speed up

dguittet:
Speed up pvsamv1, pvwattsv7 and pvwattsv8 by speeding up irradproc and sssky_diffuse_table.

irradproc:
- for intra-annual solarpos_spa calculation speed up, introduce a hash table that tracks the intermediate outputs of `calculate_spa` that reduces the number of times that `calculate_spa` is called per timestep
- for lifetime calculation, inter-annual speed up is done by storing the results of `solarpos_spa` in a vector so that the values are not recalculated each year

sssky_diffuse_table:
- change the keys of the hash table to numeric key instead of string since `sprintf` takes a long time
- reduce the number of steps/samples required to compute the self-shading loss in `sssky_diffuse_table::compute`

pvwattsv8 and pvwattsv7:
- move irrad proc outside the time loop

pvsamv1:
- move irrad proc outside the time loop
- change the constructor used for irradproc

tests:
- pvsamv1 test values are changed (shows how the actual value changed)
- pvwattsv8 test tolerances are changed (shows how much the relative error increased in order to use the old expected values). A potential additional change in this PR is to change the expected values to the new values

Speed changes 
----------------
Before is the original code, "Spa Hash" is what is implemented in this PR

image

image


mjprilliman (commented):
This looks great! I want to double check seasonal tilt behavior, so I may need to do a bit more debugging.

shared/lib_irradproc.cpp
```diff
@@ -1767,62 +1813,32 @@ void irrad::setup() {
...
+    clear_spa_table();
 }
 
-irrad::irrad(weather_record wf, weather_header hdr,
+irrad::irrad(weather_header hdr,
              int skyModelIn, int radiationModeIn, int trackModeIn,
-             bool useWeatherFileAlbedo, bool instantaneousWeather, bool backtrackingEnabled, bool forceToStowIn,
-             double dtHour, double tiltDegreesIn, double azimuthDegreesIn, double trackerRotationLimitDegreesIn,
-             double stowAngleDegreesIn,
-             double groundCoverageRatioIn, double slopeTiltIn, double slopeAzmIn, std::vector<double> monthlyTiltDegrees,
```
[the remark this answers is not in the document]
dguittet: Those degrees are input during each timestep: https://github.com/NREL/ssc/blob/221afbc592c352e676ddcb3846412f4b894c59ef/ssc/cmod_pvsamv1.cpp#L1550

The function set_from_weather_record should setup the monthly tilt, but let me check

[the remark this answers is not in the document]
dguittet: The likely issue is that the hash table assumes the tilt is always the same. So let me add a check for that...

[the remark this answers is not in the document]
dguittet: Should backtracking be modifying the tilt too?

shared/lib_irradproc.cpp
```diff
@@ -2086,6 +2145,37 @@ void irrad::set_sun_component(size_t index, double value) {
...
+        albedoSpatial = userSpecifiedSpatialAlbedos->row(month_idx).to_vector();
+        albedo = std::accumulate(albedoSpatial.begin(), albedoSpatial.end(), 0.) / albedoSpatial.size();
+    }
+    else {
+        albedo = userSpecifiedAlbedo[month_idx];
+        albedoSpatial.assign(userSpecifiedSpatialAlbedos->ncols(), albedo);
+    }
+    if (trackingMode == TRACKING::SEASONAL_TILT) {
+        tiltDegrees = monthlyTiltDegrees[month_idx];
+        trackingMode = TRACKING::FIXED_TILT;
```
mjprilliman, line 2174: Because the irrad constructor is now no longer accessing Subarrays[nn]->trackMode in the loop, this line is changing the trackingMode for good, thus locking you out of the if loop check. So I think everything else works as intended in the hash table.

shared/lib_irradproc.cpp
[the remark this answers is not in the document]
dguittet: Gotcha, but yep the tilt isn't used in the solarpos_spa calcs, so that's not the issue. Fixing the tracking mode input!

test/ssc_test/cmod_pvsamv1_test.cpp
```diff
@@ -598,7 +598,7 @@ TEST_F(CMPvsamv1PowerIntegration_cmod_pvsamv1, NoFinancialModelSystemDesign)
     pairs["inverter_count"] = 22;
     pairs["subarray1_track_mode"] = 0;
 
-    std::vector<double> annual_energy_expected = { 185649, 243262, 259014, 217887, 195101 };
```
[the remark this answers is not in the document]
dguittet: Reverted the expected test results. Thanks for catching!

mjprilliman (approved):
Looks good to me. Just one merge conflict from recent merges to address

janinefreeman:
This looks good to me! I think it makes sense to update the pvwattsv8 expected test values to the new values as well. I like scaling the error tolerance by annual energy, but I think I'd like to make the error tolerance itself smaller if we're keeping that approach.

Requesting an additional review from @sjanzou , particularly on the hash table, just because this is a big change for pvwatts and pvsam.

Super excited for these results @dguittet !!! Nice job!!!

janinefreeman (approved):

dguittet:
Need tests for Alaska and South Africa

test/ssc_test/cmod_pvwattsv5_test.cpp
```diff
@@ -91,7 +91,7 @@ TEST_F(CMPvwattsV5Integration_cmod_pvwattsv5, DifferentTechnologyInputs)
 {
 //	std::vector<double> annual_energy_expected = { 6909.79, 7123.32, 7336.478, 6909.79, 6804.376, 8711.946, 8727.704, 9690.735 };
 	// single axis tracking reduction due to pull request 280
-	std::vector<double> annual_energy_expected = { 6908.13, 7121.63, 7334.83, 6908.13, 6802.72, 8633.37, 8721.21, 9687.06 };
+	std::vector<double> annual_energy_expected = { 6908.13, 7121.63, 7334.83, 6908.13, 6802.72, 8632.33, 8721.21, 9687.06 };
```
sjanzou, line 94: Seems like the general trend is for lower annual energy in all the tests and the default configuration results.

shared/lib_irradproc.cpp
```diff
@@ -903,6 +942,11 @@ calculate_spa(double jd, double lat, double lng, double alt, double pressure, do
         azimuth = M_PI;
     }
 
+    std::vector<double> spa_outputs = {needed_values[0], needed_values[1], needed_values[2], needed_values[3], needed_values[4], needed_values[5], needed_values[6], needed_values[7], needed_values[8],
+        ascension_and_declination[0], ascension_and_declination[1]};
+
+    if (use_table)
+        spa_table[spa_key_inputs] = spa_outputs;
```
sjanzou, line 949: Have you run with use_table = false to verify previous results?

shared/lib_irradproc.cpp
```diff
@@ -1849,6 +1865,49 @@ int irrad::check() {
     return 0;
 }
 
+void irrad::setup_solarpos_outputs_for_lifetime(size_t ts_per_year) {
+    solarpos_outputs_for_lifetime.resize(ts_per_year);
+}
+
+bool irrad::getStoredSolarposOutputs() {
+    if (!solarpos_outputs_for_lifetime.size()) return false;
```
sjanzou, line 1873: Is it clearer for maintenance to use if (solarpos_outputs_for_lifetime.size() == 0) return false;

shared/lib_irradproc.h
```diff
@@ -628,9 +628,68 @@ double sun_rise_and_set(double* m_rts, double* h_rts, double* delta_prime, doubl
...
+                && temp == other.temp
+                && ascension == other.ascension
+                && declination == other.declination
+                );
+        }
+
+    spa_table_key(double j, double dt, double p, double t, double a, double d):
+        jd(j), delta_t(dt), ascension(a), declination(d)
+    {
+        int pressure_bucket = 10;
```
sjanzou, line 663: Do you have a sensitivity plot based on pressure and temp bucket size?

shared/lib_irradproc.h
```diff
@@ -628,9 +628,68 @@ double sun_rise_and_set(double* m_rts, double* h_rts, double* delta_prime, doubl
...
+template <>
+struct std::hash<spa_table_key>
+{
+  std::size_t operator()(const spa_table_key& k) const
+  {
+    using std::hash;
+    // Compute individual hash values for first, second, etc and combine them using XOR and bit shifting:
+    return 
+    ((((
+            ((((hash<double>()(k.jd)
```
sjanzou, line 681: Have you tried hash_combine in C++17 for more robustness (i.e. less collisions)?

image

shared/lib_pvshade.cpp
```diff
@@ -291,27 +291,28 @@ void selfshade_xs_horstr(bool landscape,
 
 // Accessor for sky diffuse derates for the given surface_tilt. If the value doesn't exist in the table, it is computed.
 double sssky_diffuse_table::lookup(double surface_tilt) {
-    char buf[8];
-    sprintf(buf, "%.3f", surface_tilt);
-    if (derates_table.find(buf) != derates_table.end())
-        return derates_table[buf];
+    int surface_tilt_key = int(surface_tilt * derates_table_digits_multiplier);
```
sjanzou, line 294: This seems like it would be faster - nice approach!

shared/lib_pvshade.cpp
```diff
@@ -291,27 +291,28 @@ void selfshade_xs_horstr(bool landscape,
...
+        return derates_table[surface_tilt_key];
     return compute(surface_tilt);
 }
 
 double sssky_diffuse_table::compute(double surface_tilt) {
     if (gcr == 0)
         throw std::runtime_error("sssky_diffuse_table::compute error: gcr required in initialization");
     // sky diffuse reduction
-    double step = 1.0 / 1000.0;
+	const size_t n_steps = 250;
```
sjanzou, line 305: Do you have a sensitivity analysis on some different n_steps values?

sjanzou (commented):
Looks good overall, just a few comments and questions on some of the implementation values and sensitivities. 

Thanks for tackling this!

dguittet:
> Need tests for Alaska and South Africa

@janinefreeman 

Ran Pvwattsv8 tests for Alaska and tip of South America and the changes in this PR (both self shading and solarpos), and checked changes in annual energy and monthly energy.

The max change over those 26 values was < 0.02%.

shared/lib_irradproc.cpp
dguittet: Yes. I'm planning to remove this code once we're done examining sensitivities-- thoughts on deleting this?

shared/lib_irradproc.h
dguittet: Once we've upgraded to C++17, this would be great!

shared/lib_irradproc.cpp
dguittet: Done

shared/lib_irradproc.h
dguittet: Sure, the sensitivity is super low to pressure and temperature:

pressure bucket | temp bucket | time (ms) | result change
-- | -- | -- | --
10 | 5 | 3250 | na
2 | 5 | 3443 | na
2 | 2 | 3490 | na
10 | 2 | 3411 | na
1 | 1 | 3630 | na
20 | 10 | 3238 | na

sjanzou (approved):
Thanks for tackling this!

Issue #1197, opened 2024-08-02
Update mhk equations for PySAM integration


mjprilliman:
**Describe the bug**
-Update cmod linkage for mhk ssc equations in ssc_equations.h for integration in next major release of PySAM.

mjprilliman:
Closing this for now as a wishlist item.

PR #1217, opened 2024-10-08
Update mhk equation docstrings

mjprilliman:
-Closes https://github.com/NREL/ssc/issues/1197
-See https://github.com/NREL/ssc/issues/1198 for discussion from patch


dguittet:
Are the inputs / outputs up to date for the docstrings, e.g. me_array_cable_length_doc?

I see 5 inputs in the doc but 10 in the function itself

ssc/ssc_equations.h
```diff
@@ -84,22 +84,21 @@ struct ssc_equation_entry{
 static ssc_equation_entry ssc_equation_table [] = {
         // Marine energy
         
-		{"me_array_cable_length", me_array_cable_length,
-            "Marine Energy", me_array_cable_length_doc,
+        {"me_array_cable_length", me_array_cable_length,
+            "MhkWave", me_array_cable_length_doc,
```
dguittet, line 88: I think this should be attached to the MhkCost module, since the outputs of these equations are inputs to MhkCost, not MhkWave or MhkTidal. Thoughts?

mjprilliman:
Updated equation docstrings. No longer linking mhk equations to PySAM modules at this time.

PR #1218, opened 2024-10-08
Add capacity factor input to tidal turbine ssc_eqn

mjprilliman:
-Use target capacity factor to calculate power curve for tidal turbine design feature
-Input was previously hardcoded
-See https://github.com/NREL/SAM/pull/1807


cpaulgilman (commented):
@mjprilliman Variable naming is confusing to me because I expect `target_cf` to be an input and `capacity_factor` to be an output, but here `target_cf` is being set to `capacity_factor`.

cpaulgilman (approved):
Thanks for changing the variable names

PR #1221, opened 2024-10-16
Add wind data check for missing data

cpaulgilman:
Fix https://github.com/NREL/SAM/issues/1880

Add check for missing data to wind file reader.


sjanzou (approved):
With the 2023.12.17 release, I get the results using the Wind_data_csv_format.csv file from SAM issue 1880
https://github.com/NREL/SAM/issues/1880
image

After changing hub height from 80m default to 145m, I get the informative (stof string to float error message):
image

In the sam-1880-missing-wind-data branch, I receive at default 80m hub height
image

and the more informative message at 145m hub height
image

PR #638, opened 2021-11-04
New community solar financial model

cpaulgilman:
Goes with https://github.com/NREL/SAM/pull/730

PR #1223, opened 2024-10-18
Add IRR to cashloan and migrate functions to lib_financial.h

cpaulgilman:
Fix #638 

## 1. Add IRR output to cashloan

* IRR output available for Residential and Commercial financial models.
* Add nominal discount rate output to help with identifying nominal discount rate that results in NPV = 0 (useful for parametrics on real discount rate)
*  Note that default Residential and Commercial cases use 100% debt, so IRR is NaN for default configurations because initial investment is zero
* Not modifying SAM UI to add IRR to Metrics table on Summary tab, so it is only available in Data Tables because IRR metric is a bit ambiguous for projects that do not earn revenue

## 2. Move financial model functions to lib_financial for all financial models

| Task                                       | SO      | CS      | TPO HD  | MP      | EPF     | LPF     | SLB     | TPO H    | CashLoan |
|  ---                                       | ---     | ---     | ---     | ---     | ---     | ---     | ---     | ---      | ---      | 
| Compare functions to SO before moving      | NA      | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | NA       | NA       |
| Check `depreciation_sched_custom()`        | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | NA       | NA       |
| Delete `using namespace libfin` from cmod  | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745;  | &#9745;  |
| Delete libfin functions from cmod          | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745;  | &#9745;  |
| Convert npv()to libfin::                   | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745;  | &#9745;  |
| Convert irr()to libfin::                   | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745;  | &#9745;  |
| Convert min()to libfin::                   | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | NA       | &#9745;  |
| Convert max()to libfin::                   | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | &#9745; | NA       | &#9745;  |
| Convert ppmt()to libfin::                  | &#9745; | &#9745; | &#9745; | &#9745; | NA      | &#9745; | &#9745; | NA       | &#9745;  |
| Convert payback() to libfin::              | NA      | NA      | NA      | NA      | NA      | NA      | NA      | *        | &#9745;  |

\* Note that Third Party Ownership - Host  model has payback cashflows but does not use the payback() function to calculate payback period.

## Question

That there are special `min()` and `max()` functions defined in lib_financial. I changed all calls to `min()` and `max()` in the financial cmods to `libfin::min()` and `libfin::max()` to avoid conflict, but should any of those be `std::min()` or `std::max()` instead?


shared/lib_financial.cpp
```diff
@@ -35,153 +35,240 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+                if ((initial_guess <= 0) || (initial_guess >= 1)) initial_guess = -0.5 * b + 0.5 * sqrt(b * b - 4.0 * c);
+            }
+        }
+        else if (initial_guess < 0) // first order
+        {
+            if (cf_vector.at(0) != 0) initial_guess = -(1.0 + cf_vector.at(1) / cf_vector.at(0));
+        }
+
+        double scale_factor = irr_scale_factor(cf_vector, count);
+        double residual = DBL_MAX;
```
sjanzou, line 177: DBL_MAX was not defined and is causing the Linux and MacOS tests to fail.

DBL_MAX is defined in float.h and was previously included at the top of each financial compute module as
#ifndef WIN32
#include <float.h>
#endif

sjanzou (changes requested):
For Mac and Linux GitHub Actions to pass, please add the following to the lib_financial.cpp file for DBL_MAX to be defined:


DBL_MAX is defined in float.h and was previously included at the top of each financial compute module as
#ifndef WIN32
#include <float.h>
#endif

cpaulgilman:
> For Mac and Linux GitHub Actions to pass, please add the following to the lib_financial.cpp file for DBL_MAX to be defined:
> 
> DBL_MAX is defined in float.h and was previously included at the top of each financial compute module as #ifndef WIN32 #include <float.h> #endif

@sjanzou Thanks for finding the cause of those tests failing.

shared/lib_financial.cpp
```diff
@@ -35,153 +35,244 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <limits>
 #include "lib_financial.h"
 
-using namespace libfin;
+#ifndef WIN32
+#include <float.h>
+#endif
+
+double libfin::min(double a, double b)
+{ // handle NaN
```
brtietz, line 43: Can you or @sjanzou expand upon this comment to include _why_ 0 is the correct answer when one of the inputs is NaN? My hunch based on reading the code is to make the cashflow readable, but it would be good to document that if true so future developers know whether to use this or std::min.

brtietz (approved):
Refactoring looks good! One request for a clarifying comment. Thanks!

shared/lib_financial.cpp
sjanzou: @brtietz , yes - to make the cashflow readable and that std::min and max have screwy results based on if a NaN is the first element in the comparison - see https://stackoverflow.com/questions/72296440/how-to-prevent-stdmin-and-max-to-return-nan-if-the-first-element-of-the-array

PR #1233, opened 2024-10-30
Noct taualpha

### LINKED ISSUE #917: Check NOCT model implementation
**Describe the bug**
A question on the implementation of the NOCT cell temeprature model equation used with the CEC performance models was presented in https://www.researchgate.net/publication/361463601_PV_Module_Operating_Temperature_Model_Equivalence_and_Parameter_Translation. Evaluate the implementation of the model based on the definitions presented in the SAM technical reference, the DeSoto model paper, and the Duffie and Beckman textbook. Confirm whether the POA irradiance used in the model should be before or after reflection effects, and consider how this definition effects the efficiency and absorptance-transmittance product variables used in the model.

mjprilliman:
-Use Geff_total in NOCT model rather than G_total, remove irradiance scaling of tau*alpha product
-Update tests
-Fixes #917


mjprilliman:
Need to discuss with @sjanzou how to update remaining tests, but otherwise should be good to review as is.

test/input_json/TechnologyModels/pvwattsv8/dc_ac_ratio_0.02_PVWatts_Single_Owner_cmod_pvwattsv8.json
```diff
@@ -1,5 +1,5 @@
 {
-	"solar_resource_file" : "C:/Projects/Github/NREL/working/SAM/deploy/solar_resource/phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv",
+	"solar_resource_file" : "C:/Users/mprillim/sam_dev/SAM/deploy/solar_resource/phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv",
```
janinefreeman, line 2: It looks like this line was specific to Steve's computer before, and is now specific to yours, but flagging in case it should be set to a common file... but I don't think there's an easy way to use environment variables in JSON, is there @sjanzou ? I don't require this to be fixed before merging though.

test/input_json/TechnologyModels/pvwattsv8/dc_ac_ratio_0.5_PVWatts_Single_Owner_cmod_pvwattsv8.json
```diff
@@ -1,5 +1,5 @@
 {
-	"solar_resource_file" : "C:/Projects/Github/NREL/working/SAM/deploy/solar_resource/phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv",
+	"solar_resource_file" : "C:/Users/mprillim/sam_dev/SAM/deploy/solar_resource/phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv",
```
janinefreeman, line 2: same comment as the other JSON file

janinefreeman (approved):
Looks good!

test/input_json/TechnologyModels/pvwattsv8/dc_ac_ratio_0.02_PVWatts_Single_Owner_cmod_pvwattsv8.json
sjanzou: > It looks like this line was specific to Steve's computer before, and is now specific to yours, but flagging in case it should be set to a common file... but I don't think there's an easy way to use environment variables in JSON, is there @sjanzou ? I don't require this to be fixed before merging though.

@janinefreeman ,  thanks for flagging this - the JSON generated always has the specific test machine path... However, in the test body, the weather file is updated to use 
image

and in the "Test" function, the weather file is set based on the SSCDIR/test/input_cases/general_data folder as
image

mjprilliman:
@sjanzou @brtietz is there known variance for ResidentialDCBatteryModelPriceSignalDispatch test across operating systems. I show this test as passing on Windows and Linux but failing on Mac for these temperature model changes. The battery capacity percent variable change on Mac, which I suppose is possible for different values in gen but I'm confused why it wouldn't show up across operating systems. 
image

PR #1234, opened 2024-10-30
Refactor adjust losses to use vartable instead of cmod 

brtietz:
Refactor adjust losses to make them more accessible to the battery. We expect no changes, but need to test to make sure this approach is threadsafe. My initial runs with parameterics looked reasonable.


cpaulgilman (approved):
Looks ok to me based on some basic tests in UI with parametrics and Loss Ajdustment widget.

dguittet (approved):

PR #1236, opened 2024-11-04
Consistent monthly and annual energy labels

cpaulgilman:
Fixes https://github.com/NREL/SAM/issues/1876

Use following labels as much as possible

`monthly_energy` "Monthly energy in Year 1"
`annual_energy` "Annual energy in Year 1"


ssc/cmod_pvwattsv8.cpp
```diff
@@ -203,16 +203,16 @@ static var_info _cm_vtab_pvwattsv8[] = {
...
-        { SSC_OUTPUT,       SSC_ARRAY,       "monthly_energy",                 "Monthly energy",                              "kWh",       "",                                             "Monthly",          "",                       "LENGTH=12",                          "" },
-        { SSC_OUTPUT,	    SSC_MATRIX,		 "annual_energy_distribution_time","Annual energy production as function of Time",				"",				"",				"Heatmaps",			"",						"",							"" },
+        { SSC_OUTPUT,       SSC_ARRAY,       "dc_monthly",                     "Monthly DC energy",                             "kWh",       "",                                             "Monthly",          "",                       "LENGTH=12",                          "" },
+        { SSC_OUTPUT,       SSC_ARRAY,       "ac_monthly",                     "Monthly AC energy",                            "kWh",       "",                                             "Monthly",          "",                       "LENGTH=12",                          "" },
```
mjprilliman, line 207: This is probably for a different issue, but I believe `ac_monthly` and `monthly_energy` will always be the same. Do we need both?

mjprilliman (approved):

ssc/cmod_pvwattsv8.cpp
cpaulgilman: > This is probably for a different issue, but I believe ac_monthly and monthly_energy will always be the same. Do we need both?

This is probably a vestige of the original implementation of PVWatts. I don't think we need both in SAM, but removing `ac_monthly` may break code out there in PySAM or other apps that call the PVWatts library.

PR #1239, opened 2024-11-07
Battery cycle-calendar life model updated to use sum not min

dguittet:
I modified the cycle calendar model to take the sum instead of the min. Here are the results of 25-yr battery capacity percent (replacements are off) across the different chemistries and models. Overall, there are 6675 cycles. Average DOD is ~50%
 
The table models have more degradation now.
 
However, whereas in the Old column, the NMC/Gr table and algebraic model both had more degradation than LFP, in the New version, the LFP table has less degradation than NMC/Gr Alg. The LMO/LTO table doesn’t change because the calendar loss is 0.
 
This is fine because the calendar-cycle models aren’t comparable to the algebraic ones—apples and oranges.

Model | Value | Old | New
-- | -- | -- | --
NMC/Gr Alg | 25-yr cap % | 76.0021 | 76.0021
NMC/Gr Table | 25-yr cap % | 65.4057 | 47.6304
LFP/Gr Table | 25-yr cap % | 82.2394 | 65.8467
LMO/LTO Alg | 25-yr cap % | 92.9983 | 92.9983
LMO/LTO Table | 25-yr cap % | 89.7067 | 89.7067

<p class="MsoNormal" style="margin: 0in; font-size: 16px; font-family: Aptos, sans-serif; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;"><span style="font-size: 11pt; line-height: 1.2;"> </span></p>

### DIFF
--- a/shared/lib_battery_lifetime_calendar_cycle.cpp
+++ b/shared/lib_battery_lifetime_calendar_cycle.cpp
@@ -263,6 +263,8 @@ double lifetime_cycle_t::average_range() { return state->average_range; }
 
 double lifetime_cycle_t::capacity_percent() { return state->cycle->q_relative_cycle; }
 
+double lifetime_cycle_t::capacity_loss_percent() { return fmax(0, 100. - state->cycle->q_relative_cycle); }
+
 void lifetime_cycle_t::resetDailyCycles() {
     state->cycle->DOD_min = -1;
     state->cycle->DOD_max = -1;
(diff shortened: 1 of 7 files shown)


dguittet:
@brtietz The tests are failing because the error between Old Mac/Linux/Windows vs New Mac is greater than the abs tolerance of 0.01 for some of the battery tests. Mind if I increase the tol for this particular test? `runTestCycleAt3C`

brtietz:
> @brtietz The tests are failing because the error between Old Mac/Linux/Windows vs New Mac is greater than the abs tolerance of 0.01 for some of the battery tests. Mind if I increase the tol for this particular test? `runTestCycleAt3C`

Increasing the tol seems good to me. Maybe 0.1?

dguittet:
> > @brtietz The tests are failing because the error between Old Mac/Linux/Windows vs New Mac is greater than the abs tolerance of 0.01 for some of the battery tests. Mind if I increase the tol for this particular test? `runTestCycleAt3C`
> 
> Increasing the tol seems good to me. Maybe 0.1?

Yep. Tests are passing now

brtietz (approved):
Results look good, tests all change in the direction that we expected. Do we need to feed any of this back to Kandler and Paul?

PR #1248, opened 2024-11-24
Fix SAM issue 1927

sjanzou:
Fixes missing value for PPA price in cents/MMBtu for analysis period year - see https://github.com/NREL/SAM/issues/1927
image


cpaulgilman (approved):
Thank you for finding and fixing that.

PR #1249, opened 2024-11-26
Update marine device cost curves

mjprilliman:
-Update marine cost curves based on additional data and inflation to 2023 dollars.


cpaulgilman (approved):
@mjprilliman Looks good! Changes are consistent with values in 2024 Model column of tables in Word doc you sent me. I don't see any changes in electrical infrastructure costs, so I assume those stay the same.

PR #1251, opened 2024-12-02
Create functions for battery parameters that were accessed in the code.

brtietz:
This prevents copying the entire params object multiple times each step. Improves battery runtimes, especially for PV smoothing.

Happy to add const pointer changes to this PR as well if we want to do that too.


dguittet (approved):

sjanzou (approved):
Looks good and thanks for adding access to only the necessary parameters - makes debugging easier for our future selves!

Any stats on the reduced memory or execution time?

brtietz:
> Any stats on the reduced memory or execution time?

Looks like this shaved 56 seconds off of the PVSmoothing tests. Develop:

image

This branch:

image

PR #1257, opened 2024-12-06
Update hybrid om vartable to input monthly_energy, not inout

mjprilliman:
-Allows AC monthly energy to not be labeled as AC for IPH models
-Still shows AC monthly energy for hybrid configs


taylorbrown75 (approved):
IPH monthly energy outputs now maintain their labels from their respective technology cmod vartables. Looks good.

PR #1258, opened 2024-12-06
Ssc 1256 better hybrid batteries

### LINKED ISSUE #1256: Battery charge and discharge doesn't get added to gen in hybrids compute modules
**Describe the bug**
Hybrids compute modules run the battery, but then those battery power values never effect gen, and therefore the financials

**To Reproduce**
1. Create a PVWatts-Wind-Battery Single Owner model
2. Run it
3. Look at gen and electricity to/from battery
4. Battery is dicharging at 7 pm but gen shows 0

**Expected behavior**
Battery outputs should be included in gen

**Screenshots**
image

 - OS: Windows
 - Version 2023.12.17, develop

brtietz:
Fixes #1256 

Pairs with https://github.com/NREL/SAM/pull/1958

Overwrites "gen" in the hybrids compute module with the battery's gen, since the battery gen is an inout and include the total powerflow from all components.

Calculates annual and monthly energy values after the battery runs to capture them in the values.

Updates the test to compare these energy metrics and make sure everything is included in the annual outputs.

### DIFF
--- a/test/ssc_test/cmod_hybrid_test.cpp
+++ b/test/ssc_test/cmod_hybrid_test.cpp
@@ -67,7 +67,9 @@ TEST_F(CmodHybridTest, PVWattsv8WindBatterySingleOwner) {
     {
         int len;
 
-        ssc_number_t pvannualenergy, windannualenergy, battannualenergy, npv;
+        ssc_number_t pvannualenergy, windannualenergy, battannualenergy, npv, total_energy;
+        ssc_number_t* battchargeenergy;
+        ssc_number_t* battdischargeenergy;
         auto outputs = ssc_data_get_table(dat, "output");
         auto inputs = ssc_data_get_table(dat, "input");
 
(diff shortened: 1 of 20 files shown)


mjprilliman (commented):
'Electricity net generation' now does not equal the generator annual energy values, due to efficiency losses of energy flowing through the battery. That's fine I think, then O&M production-based expense scales with net generation. Is that right? 

image

gen matches what is coming out of battery compute module now. 

image

brtietz:
> That's fine I think, then O&M production-based expense scales with net generation. Is that right?

Hybrid VOM is based on energy to grid, not net generation. I've set up a few test cases where the net generation is negative, do these make sense to you? 
[hybrid_vom_tests.zip](https://github.com/user-attachments/files/18044321/hybrid_vom_tests.zip)

mjprilliman:
> > That's fine I think, then O&M production-based expense scales with net generation. Is that right?
> 
> Hybrid VOM is based on energy to grid, not net generation. I've set up a few test cases where the net generation is negative, do these make sense to you? [hybrid_vom_tests.zip](https://github.com/user-attachments/files/18044321/hybrid_vom_tests.zip)

Yes that looks right.

mjprilliman (approved):
Approving, but would prefer if @sjanzou can take a look as well before merging.

test/ssc_test/cmod_hybrid_test.cpp
```diff
@@ -210,12 +244,17 @@ TEST_F(CmodHybridTest, CustomGenerationPVWattsWindFuelCellBatteryHybrid_SingleOw
         EXPECT_NEAR(genericannualenergy, 756864000., 1e6);
         EXPECT_NEAR(pvannualenergy, 211907456., 1e6);
         EXPECT_NEAR(windannualenergy, 366975552., 1e6);
+        EXPECT_NEAR(fuelcellannualenergy, 1561993, 1e6);
         EXPECT_NEAR(battannualenergy, 1331720000., 1e6);
+        EXPECT_NEAR(battchargeenergy[1], 55372248, 55372248 * 0.001);
```
sjanzou, line 249: Looks good and thanks for adding the additional outputs in the test!

sjanzou (approved):
Looks good and the results make more sense not - we may have to revisit the variable o and m value for the system in the future along with subcomponent incentives.

PR #1264, opened 2024-12-10
Allow for zero time series buy and sell rates

brtietz:
Use a boolean instead of tier_charge == 0 to determine whether time series rates or table rates should be used.

Fixes https://github.com/NREL/SAM/issues/1964


sjanzou (approved):
SAM issue 1964 resolved using this update!

PR #1271, opened 2025-01-03
add GeothermalCosts eqns

dguittet:
Moves the getem cost equation to be attached to geothermal_costs instead of geothermal.

@mjprilliman Please update `getem_om_cost_calc_doc` in `ssc/cmod_geothermal_costs_eqns.h`. I've made a skeleton of the documentation


sjanzou (approved):
Looks good, I will let Matt have final say with the getem_om_cost_calc_doc update

PR #1273, opened 2025-01-31
Address SAM issue 1990 - 5 minute weather data header info

sjanzou:
Addresses https://github.com/NREL/SAM/issues/1990

### DIFF
--- a/shared/lib_windfile.cpp
+++ b/shared/lib_windfile.cpp
@@ -115,7 +115,7 @@ static float col_or_nan(const std::string& s)
         }
     }
     else
-        return std::numeric_limits<float>::quiet_NaN();;
+        return std::numeric_limits<float>::quiet_NaN();
 }
 
 static int locate2(std::string buf, std::vector<std::string> &vstring, char delim)
@@ -479,7 +479,12 @@ bool windfile::open( const std::string &file )
         std::vector<std::string> hdr;
         int ncols = locate2(m_buf, hdr, ',');
 
-        for (size_t i = 0; (int)i < ncols; i++)
+        lat = std::numeric_limits<float>::quiet_NaN();
+        lon = std::numeric_limits<float>::quiet_NaN();
+        elev = std::numeric_limits<float>::quiet_NaN();
+
+//        for (size_t i = 0; (int)i < ncols; i++) SAM issue 1990
+        for (size_t i = 0; (int)i < ncols && (i+1) < hdr.size(); i++)
         {
             hdr_item = util::lower_case(trimboth(hdr[i]));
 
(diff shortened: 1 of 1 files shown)


cpaulgilman:
@sjanzou Thanks for tackling this one. There are a couple of things we need to sort out to fix this properly:

The Windpower model does not appear to use lat, lon, or elev, so I'm not sure the wind file reader needs to check those values.  (This data is not included in the new [BC-HRRR CONUS endpoint](https://developer.nrel.gov/docs/wind/wind-toolkit/wtk-bchrrr-v1-0-0-download/).)

The WIND Toolkit files include both a "site timezone" and a "data timezone." Some endpoints like `india-wind-download` deliver files with only "site time zone", while others like `wtk-pr100-5min-download` have only "data time zone".  SAM is currently checking that both values are the same to ensure that the time stamps in the weather file match the time stamp that SAM assigns to each simulation time step. This is important for financial models with time-dependent pricing (TOD PPA multipliers or TOU electricity rates). Note that the [format description for the SAM CSV Format for Wind](https://samrepo.nrelcloud.org/help/weather_format_csv_wind.html) defines Row 1 as being in the hour ending at 1 am on January 1 local time.

We also need to hear back from the WIND Toolkit team to know if there will be a single format for all endpoints or if we need to modify the wind file reader to handle a few format variations.

I'll put this on the agenda for this week's team meeting.

sjanzou:
> @sjanzou Thanks for tackling this one. There are a couple of things we need to sort out to fix this properly:
> 
> The Windpower model does not appear to use lat, lon, or elev, so I'm not sure the wind file reader needs to check those values. (This data is not included in the new [BC-HRRR CONUS endpoint](https://developer.nrel.gov/docs/wind/wind-toolkit/wtk-bchrrr-v1-0-0-download/).)
> 
> The WIND Toolkit files include both a "site timezone" and a "data timezone" field in the header. Some endpoints like `india-wind-download` deliver files with only "site time zone", while others like `wtk-pr100-5min-download` have only "data time zone". SAM is currently checking that both values are the same to ensure that the time stamps in the weather file match the time stamp that SAM assigns to each simulation time step. This is important for financial models with time-dependent pricing (TOD PPA multipliers or TOU electricity rates). Note that the [format description for the SAM CSV Format for Wind](https://samrepo.nrelcloud.org/help/weather_format_csv_wind.html) defines Row 1 as being in the hour ending at 1 am on January 1 local time.
> 
> We also need to hear back from the WIND Toolkit team to know if there will be a single format for all endpoints or if we need to modify the wind file reader to handle a few format variations.
> 
> I'll put this on the agenda for this week's team meeting.

Thanks for the details - let's discuss at the meeting on how to close issue https://github.com/NREL/SAM/issues/1990

cpaulgilman (approved):

PR #1274, opened 2025-02-05
Add steam heat sink for IPH models

tyneises:


taylorbrown75 (approved):
Tested on my machine, works well.

PR #1276, opened 2025-02-12
Remove duplicate utility bill variables

cpaulgilman:
Fixes https://github.com/NREL/sam/issues/1879

Goes with https://github.com/NREL/SAM/pull/2009, see that PR for details.


sjanzou (approved):
Sorry you had to change 47 test files!

brtietz:
I ran find and replace a few times to identify potential errors, and turned up the following:

image

(sys_system instead of just sys)

There were also a few hits in the PySAM examples:

image

What is the appropriate time to update those?

cpaulgilman:
> There were also a few hits in the PySAM examples:

This should be addressed in https://github.com/NREL/pysam/pull/185

cpaulgilman:
> (sys_system instead of just sys)

Good find. Thanks! Should be fixed in 0de61b7099cac82897330d38d45b64c4d4c77edd.

I also noticed one other set of variables that use "system" instead of "sys" in this context:

```
year1_monthly_peak_wo_system
year1_monthly_use_wo_system
year1_monthly_peak_w_system
year1_monthly_use_w_system
```

@brtietz, is it worth changing that? These variables seem to only be used as outputs to utilityrate5, so changing their names should have no impact on other SSC/SAM code and samples, but will affect any SDK/PySAM code out in the world that refers to those outputs.

Issue #1202, opened 2024-08-12
Underestimation in Battery power integration from voltage curve


dguittet:
Due to the way power is calculated as a coarse integration of the voltage curve, the power flowing into the battery for charging may be significantly more than the capacity of the battery.

The voltage model shows how the voltage changes as a function of capacity, Ah, in the battery. The quantity measured by SOC is the Ah in the battery. And the power charged or discharged is the delta Ah * voltage of the battery. This power is the rectangle formed by taking the area under the curve:
 
![image](https://github.com/user-attachments/assets/93eac1dd-8b21-43b0-8fa0-000e4c55a670)

When there are relatively large step sizes being taken by the charge and discharge current, the estimated area under the curve is under-estimating by quite a bit. This is worse using a 0-100% range of operation where the voltage curve is very nonlinear.

In this example, an LFP battery of 808 MWh is charged fairly rapidly from 0 to 100% SOC and the total DC power to the battery is 983MWh.

![image](https://github.com/user-attachments/assets/5c4bc3f2-ca6a-47a9-b2e4-85377758a673)

Here you can see the delta kWh batt of the battery, as calculated by battery voltage and delta Ah, is the same as the DC electricity to the battery. But the right two columns show how there’s a mismatch between the delta SOC by internal Ah and the delta SOC calculated by kWh. It takes much more change in kWh to make a change in internal SOC, because of how the internal voltage is being calculated.

A bandaid fix is to change the voltage model to the table model using a very flat voltage curve, and also to reduce the operating range.

![image](https://github.com/user-attachments/assets/db36ef14-0f2e-4228-9026-de4767079c09)

Now the DC electricity to charge the battery from 0 to 95% is a more reasonable 746 MWh. 

**Expected Behavior**
The power flow to the battery should more accurately reflect the change in stored energy of the battery. Currently, there is a lot of loss due to the coarse method of numerical integration.

@awabbes

brtietz:
Possibly related: https://github.com/NREL/ssc/issues/569

PR #1277, opened 2025-02-14
Incorperate inverter losses into grid limits code

brtietz:
The previous grid limits code did not properly account for inverter night time losses, this code corrects that issue.

Note that there are still some issues with power to grid below 10% battery SOC. Those might be better solved as a part of https://github.com/NREL/ssc/issues/1202

Fixes https://github.com/NREL/SAM/issues/1982


cpaulgilman (approved):
Looks good, thanks for fixing.

PR #1278, opened 2025-02-14
Avoid giving free energy at the end of a grid outage

### LINKED ISSUE #1270: Batteries get free energy at the end of a grid outage
**Describe the bug**
The battery uses a different minimum state of charge threshold during an outage versus when the grid is available. Typically the latter is higher. When the grid outage ends, the state of charge automatically resets to the minimum non-outage version without the battery charging being accounted for in the powerflow.

**To Reproduce**
1. Create a detailed PV-battery behind the meter case
2. Add a grid outage where the battery fully discharges to 10%
3. Simulate the case and look at the battery state of charge when the grid outage ends

**Expected behavior**
The battery model code should tolerate an SOC lower than the minimum until the dispatch code has a chance to charge.

**Screenshots**
image

 - Version 2023.12.17, 2024.12.12

brtietz:
The check_SOC() function in capacity_t was written before the grid outage code, and didn't expect the SOC limits to change dynamically during the simulation. Therefore at the end of the grid outage, the check_SOC() function granted the battery some free energy.

Changing the function means that the charge value is only adjusted when the current is adjusted, and the battery doesn't spontaneously gain energy without charging.

Test file here: [ssc_1270_test_case (2).zip](https://github.com/user-attachments/files/18828139/ssc_1270_test_case.2.zip)
Expectation is that the grid outage will end a few hours before the battery recharges above 10% (even though the minimum SOC is 30%.

Fixes #1270

### DIFF
--- a/shared/lib_battery_capacity.cpp
+++ b/shared/lib_battery_capacity.cpp
@@ -162,8 +162,11 @@ void capacity_t::check_SOC() {
 
             // do not switch to discharging
             state->cell_current = fmin(0, state->cell_current);
+
+            // Correct state given current change
+            state->q0 = q_upper;
         }
-        state->q0 = q_upper;
+        
     }
     else if (state->q0 < q_lower - tolerance) {
         // if undercharged then reduce discharhing
@@ -172,6 +175,8 @@ void capacity_t::check_SOC() {
 
             // do not switch to charging
             state->cell_current = fmax(0, state->cell_current);
+
+            // Correct state given current change
             state->q0 = q_lower;
         }
     }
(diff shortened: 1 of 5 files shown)


sjanzou (commented):
Maybe one more test to update - seems consistent on all platforms
image

brtietz:
Thanks for flagging! I also realized the max SOC code required the same change, so have updated the test file and the manual tests to deal with that case.

sjanzou (approved):
Thanks for updating the tests. The max SOC seems to work without the "free energy" results now!

dguittet (approved):
It makes sense that the charge in the battery doesn't change unless there's some charging or discharging

PR #1284, opened 2025-02-26
Fix wind adjustment factor double-counting

### LINKED ISSUE #1283: Windpower module applies HAF twice
I believe the second call to `haf.setup` here may be applying the hourly adjustment factors twice:

https://github.com/NREL/ssc/blob/82d9c9ebb98b556018e02ffa8464f53528be31d4/ssc/cmod_windpower.cpp#L569-L570

I noticed this when running PySAM 5+. I am inputting an hourly adjustment factor of 17% for a particular timestep, but SAM gives back a power generation value that is reduced by 31.11% (= 1 - 0.83 * 0.83).

Happy to provide more specific details if needed.

cpaulgilman:
Fix #1283

To test, run a default Wind / No Financial case with and without time series adjustment factors. The difference between the two runs should be the adjustment factor applied once, not twice as reported in #1283.

image

The adjustment factors should work correctly for the following combinations:

* Hourly weather file with hourly losses
* Subhourly weather file with subhourly losses
* Weibull distribution with hourly losses
* Wind resource probability table with hourly losses
* Constant loss wake model option with hourly or subhourly simulations/losses

A simulation error should be generated with the wrong combination of wind data time step and loss time step.

The attached file has .sam file with cases set up for these tests and a 15-minute wind file:  
[windpower-adj-test.zip](https://github.com/user-attachments/files/18974762/windpower-adj-test.zip)

This PR also fixes time series adjustment losses (`haf`) indexed to `hr` instead of time step `i`: Subhourly loss supported as of https://github.com/NREL/SAM/pull/1164.

### DIFF
--- a/ssc/cmod_windpower.cpp
+++ b/ssc/cmod_windpower.cpp
@@ -335,7 +335,7 @@ void cm_windpower::exec()
 	if (!wt.isInitialized())
 		throw exec_error("windpower", util::format("wind turbine class not properly initialized"));
 	if (wpc.nTurbines < 1)
-		throw exec_error("windpower", util::format("the number of wind turbines was zero."));
+		throw exec_error("windpower", util::format("the number of wind turbines was zero"));
 
     // check for maximum number of turbines
     int newMaxTurbines = 0;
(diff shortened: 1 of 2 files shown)


sjanzou (approved):
This works but the cmod_windpower compute module is extremely confusing... why is there a return statement in and if block on line 424? Seems like the Weibull model and the Wake model should be separated with an if else structure for easier debugging.

Also, the error message when using the 15min resource file and the hourly adjustment factors is a little confusing
image

Would it be clearer to use "resource file time steps" (for non-probability table inputs) instead of "simulation time steps"?

Also, same issue with the "wind resource probability table" case in the supplied "windpower-adj-test.sam" file:
image

There is an implied simulation time steps of 8760 using the wind resource probability table option...

janinefreeman (approved):

cpaulgilman:
> ...but the cmod_windpower compute module is extremely confusing... why is there a return statement in and if block on line 424? Seems like the Weibull model and the Wake model should be separated with an if else structure for easier debugging.

I agree. I implemented this solution in a way that did not require restructuring the compute module. I think the double-counting of adjustment factors was originally caused by the confusing structure.

janinefreeman:
All scenarios work well for me. Agree with @sjanzou  that changing the error label to "resource file time steps" might be helpful, but not totally opposed to keeping "simulation time steps" in those error messages. 

@sjanzou agree that the wind compute module can be confusing and might do good with a refactor, but unfortunately we don't have funding for that right now. I think the return in the if chunk on line 424 can stay for now. ;)

cpaulgilman:
> There is an explicit simulation time steps of 8760 using the wind resource probability table option (wind_resource_model_choice = 2) on line 488 after the adjustment factors are setup...

Yes. The number of time steps is set in several places and determined from the weather file in other places. Some functions appear to require 8760 time steps (e.g., Weibull distribution and wind resource probability table and possibly some wake model options), and others work with either hourly or subhourly time steps. The compute module should be restructured so that all functions support subhourly time steps to work with hourly and subhourly adjustment losses, and to work with subhourly financial model features like market prices, etc.

cpaulgilman:
> Would it be clearer to use "resource file time steps" (for resource file inputs (wind_resource_model_choice = 0) instead of "simulation time steps"?

I worded those error messages that are associated with calls to `haf.setup()` to help with debugging -- they help me see where in the code the error was generated. I'll revise them to make them less confusing for users.

PR #1286, opened 2025-02-27
Add Github Action to Export Test Times

dguittet:
Add step in existing GitHub workflow to upload an artifact containing CSVs of elapsed test times

In a future branch, will add a new Workflow to compare CSVs across branches. Need to merge this one first so other branches upload the artifacts


sjanzou (approved):
Great! this will be useful going forward.

PR #1287, opened 2025-02-28
Code Coverage workflow added

sjanzou:
Code coverage workflow added coverage.yml
Runs on Linux only

Please check the Readme.md file and click on badge for coverage results
image

Note that this is a separate workflow file (coverage.yml) from the existing ci.yml and we can decide when to run.

Let me know if you have any issues reaching the badge link with the coverage results at https://coveralls.io/github/NREL/ssc


.github/workflows/ci.yml
```diff
@@ -1,246 +0,0 @@
-name: CI
```
brtietz, line 1: Why was this deleted? Are there restrictions on GitHub that would prevent us from running both workflows?

brtietz (changes requested):
Great job getting this working! Is there a way to fix the "source not available" issue in the coverage results? It's really useful if we can see which lines are and aren't covered, if possible.

image

janinefreeman:
Coverage viewing worked great for me! I defer to Brian on other feedback :P Thank you @sjanzou !

sjanzou:
> Great job getting this working! Is there a way to fix the "source not available" issue in the coverage results? It's really useful if we can see which lines are and aren't covered, if possible.
> 
> image

Researching issue for paths on coveralls.io, since the source code viewing is working locally on Ubuntu development machine
image

sjanzou:
Requested changes made:
1. ci.yml added back from latest patch branch
2. source files now viewable from coverage report showing uncovered lines in red, e.g.  cmod_cashloan:
image

sjanzou:
Coverage badge now shows latest coverage percentage (3% was with only the cashloan and singleowner tests running)
image

sjanzou:
Somewhat useful looking at the line covered in pvsamv1:
image

brtietz:
> Somewhat useful looking at the line covered in pvsamv1: 

Fantastic! It looks like an additional requirement is that Coveralls gives the "source not available" to a logged out user, but after I logged in with GitHub I get this view. Thank you!

brtietz (approved):

CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ option(SAMAPI_EXPORT "Export of ssc binaries to the SAM_api directory; for Unix,
 #   environment vars LK_LIB and LKD_LIB can be used to specify where to find those libraries
 #
 
+# if using same custome command for several subfolders - will not work for VS or XCode generation
+#set_property(GLOBAL PROPERTY ALLOW_DUPLICATE_CUSTOM_TARGETS 1)
+
```
dguittet, line 26: Is this commented out code needed?

CMakeLists.txt
```diff
@@ -92,13 +95,21 @@ function(set_default_compile_options target)
         endif()
         if(CMAKE_BUILD_TYPE STREQUAL "Debug")
             set(MAIN_CFLAGS "${MAIN_CFLAGS} -D_DEBUG")
+# code coverage generation using gcov and lcov - linux only
+            if (NOT APPLE AND ENABLE_COVERAGE)
+#                enable_testing()
+                # set compiler flags
+#                set(MAIN_CFLAGS "${MAIN_CFLAGS}-g -O0 -fprofile-arcs -ftest-coverage")
+                set(MAIN_CFLAGS " -D__UNIX__ -D__64BIT__ -g -O0 --coverage")
+           endif()
```
dguittet, line 104: `D__UNIX__` is already set in line 91/94, `D__64BIT__` in line 78/81, do they need to be repeated?

Also, are the commented out code needed?

sjanzou: For code coverage, the previously defined ${MAIN_CFLAGS} are overwritten with optimization removed, so the compiler flags -D__UNIX__ -D__64BIT__  are repeated for the coverage build.

sjanzou:
> Looks great and super useful feature! Just a couple of clean up questions

Thanks for reviewing and comments have been removed and the code coverage cflags are updated as noted.

PR #1288, opened 2025-02-28
Improve error handling for forecast price signals object

brtietz:
Handle previously unhanded error reports from the forecast price signal object. Now throws an exec error instead of crashing.

This is not testable with the GUI, only SDK, PySAM, or gtest. I can provide PySAM code to reproduce if desired, otherwise the new tests are likely sufficient.


sjanzou:
> Handle previously unhanded error reports from the forecast price signal object. Now throws an exec error instead of crashing.
> 
> This is not testable with the GUI, only SDK, PySAM, or gtest. I can provide PySAM code to reproduce if desired, otherwise the new tests are likely sufficient.

The changes look good... Could you please provide the pysam test script? Thanks!

brtietz:
Please download: [BatteryMerchantPlantPricingTest.zip](https://github.com/user-attachments/files/19059752/BatteryMerchantPlantPricingTest.zip) for a PySAM script that causes the error. In PySAM 6.0.1 this crashes with no output. On this branch it should produce an error about matrix dimensions.

Note that the MerchantPlant can use the inputs in this format, so I'm open to an alternative solution that brings those matrix transformations to bear (but think this error handling code is helpful regardless)

sjanzou (approved):
Agreed, the message is much better than the no message before ;-)
image

Should the BatteryMerchantPlantPricingTest.py (or something to trigger the message) be added to the pysam tests? Should this be tied or referenced to the pysam repo?

test/ssc_test/cmod_battery_test.cpp
```diff
@@ -186,3 +186,55 @@ TEST_F(CMBattery_cmod_battery, ResilienceMetricsFullLoadLifetime){
     }
     EXPECT_EQ(max_indices[0], 2);
 }
+
+/// Test standalone battery compute modeule with a input lifetime generation and merchant plant prices
+TEST_F(CMBattery_cmod_battery_fom, MerchantPlantLifetimeAutomatic) {
```
dguittet, line 191: Was lifetime model crucial to the case being fixed here?

brtietz: Good call. I converted the test to a single year.

dguittet (approved):
LGTM

PR #1293, opened 2025-03-10
Throw an error if all billing demand periods are zero

brtietz:
Fixes https://github.com/NREL/SAM/issues/1985

To test:

1. Create a PVWatts commercial case
2. Enable billing demand lookback
3. Run the case - case should run successfully
4. Set all "Included in billing demand" values under "billing demand by time-of-use period for demand charges" values to zero
5. Run the case - new error message should appear.

### DIFF
--- a/shared/lib_utility_rate_equations.cpp
+++ b/shared/lib_utility_rate_equations.cpp
@@ -914,7 +914,7 @@ void rate_data::setup_demand_charges(ssc_number_t* dc_weekday, ssc_number_t* dc_
 	}
 }
 
-void rate_data::setup_ratcheting_demand(ssc_number_t* ratchet_percent_matrix, ssc_number_t* bd_tou_period_matrix)
+bool rate_data::setup_ratcheting_demand(ssc_number_t* ratchet_percent_matrix, ssc_number_t* bd_tou_period_matrix)
 {
     // Error checked in SSC variables
     size_t nrows = 12;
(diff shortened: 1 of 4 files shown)


sjanzou (approved):
Error message looks good and is repeated twice for emphasis ;-)
image

As far as should the lookback periods look at energy charge periods for a case like this? Or do we want to force the user to set up demand charge periods and then set the demand charges to zero?

Maybe put the onus on the user - if they are using lookback periods, then they should be able to add detailed inputs - @cpaulgilman ?

cpaulgilman:
> As far as should the lookback periods look at energy charge periods for a case like this? Or do we want to force the user to set up demand charge periods and then set the demand charges to zero?
> 
> Maybe put the onus on the user - if they are using lookback periods, then they should be able to add detailed inputs - @cpaulgilman ?

I agree with Steve. If billing demand lookback is enabled, then the user should set up the lookback periods correctly (with at least one "Included in Billing Demand (0/1)" set to 1 as per the Help note mentioned in https://github.com/NREL/SAM/issues/1985).

cpaulgilman (approved):

brtietz:
Thanks, I've adjusted the test data to reflect that change. Do you recommend any adjustments to the error message (or help) to reflect the change as well? A small number of users will have files that were valid in 2024.12.12 that will not run with this change, and I'm not sure versions.lk is the right way to solve that (too many edge cases, and too much risk of overwriting valid data)

sjanzou:
> Thanks, I've adjusted the test data to reflect that change. Do you recommend any adjustments to the error message (or help) to reflect the change as well? A small number of users will have files that were valid in 2024.12.12 that will not run with this change, and I'm not sure versions.lk is the right way to solve that (too many edge cases, and too much risk of overwriting valid data). "Wait for the forum question" might also be an acceptable answer given how small the number of users might be.

@brtietz ,  I think the error message is fine now and I think the "Wait for the forum question" is appropriate for the small number of users affected.  However, I defer to @cpaulgilman for the final verdict.

PR #1294, opened 2025-03-10
Enforce non-negative batt to load numbers

brtietz:
Fixes https://github.com/NREL/sam/issues/1914 - see issue for test file.


cpaulgilman (approved):
Results look good in SAM. Thanks for fixing.

PR #1295, opened 2025-03-11
Update cmod_communitysolar to address https://github.com/NREL/SAM/issues/1867

sjanzou:
To test, run the attached PVWatts / Community Solar SAM project file with and without subscriber upfront fees and check cashflow subscriber benefits and subscriber NPV values in results metric tables
[SAM_1867.zip](https://github.com/user-attachments/files/19179956/SAM_1867.zip)


cpaulgilman (approved):
Looks good, thanks for fixing. Should this go into Develop instead of Patch given our current release plan?

sjanzou:
> Looks good, thanks for fixing. Should this go into Develop instead of Patch given our current release plan?

I think patch for now and then will be merged into develop - @brtietz  - correct?

PR #1297, opened 2025-03-13
Move five more identified cmods to the attic

brtietz:


sjanzou (approved):
I was not sure about cmod_fossilgen but it seems like it was only used temporarily back in 2011 

image

Issue #1230, opened 2024-10-28
Add battery dispatch periods to outputs


brtietz:
Add an output similar to TOD factors or TOU periods to the battery outputs when using manual dispatch - each hour should have a number 1 to 6 so users can easily identify behavior without having to squint at which hour it is currently.

Double check that this is done consistently for day/week inputs throughout the model.

PR #1298, opened 2025-03-14
Add new dispatch period output for manual dispatch

brtietz:
New variable "Battery manual dispatch period in this timestep" should appear for manual dispatch (both front of meter and behind the meter) but not for other dispatch modes.

Chose to do a lifetime variable instead of year 1 since I think we're trying to move in that direction.

Fixes https://github.com/NREL/ssc/issues/1230


dguittet (approved):
LGTM

PR #1299, opened 2025-03-17
Old CSP code to attic

tyneises:


test/ssc_test/cmod_json_comparison_test.h
tyneises, line 1: @sjanzou some cmods use "file_name" instead of "solar_resource_file", so I added a line to define "file_name" the same way as "solar_resource_file".

sjanzou: Thanks for updating!

sjanzou (approved):
Thanks for adding tests and increasing code coverage while removing compute modules and tests!

sjanzou (approved):
Looks good with a bump to code coverage!

sjanzou (approved):
Hmmm... historical record of removed equations?

Issue #1301, opened 2025-03-21
Typo in batt_adjust_constant


brtietz:
![Image](https://github.com/user-attachments/assets/faa5c4bd-9e20-48d2-ae98-19880414d6d5)

PR #1302, opened 2025-03-21
Fix batt adjust constant typo

brtietz:
Fixes https://github.com/NREL/ssc/issues/1301


cpaulgilman (approved):

sjanzou (approved):
Is only VS had a spell check ;-) - maybe co-pilot...

PR #1303, opened 2025-03-25
Retail rates dispatch grid export

brtietz:
Add logic to retail rates dispatch that will discharge power to the grid if the checkbox is checked and if a time of use period (or time series sell rate) has a sell rate greater than the buy rate that day.

Test file: 
[nbt_sam_analysis_pr_test.zip](https://github.com/user-attachments/files/19451630/nbt_sam_analysis_pr_test.zip)
Expect to see a meaningful amount (~29 kWh) of battery to grid export in August, but not other months. Feel free to experiment with time of use rates to create more substantial grid discharge.

Pairs with https://github.com/NREL/SAM/pull/2041

Fixes https://github.com/NREL/SAM/issues/2040


cpaulgilman (commented):
Thanks for adding this. I'm having a hard time coming up with a scenario where battery to grid adds a lot of value, but I am seeing the battery discharge to grid.

For the attached file, I would expect battery to discharge to grid during TOU period 2 (Noon-3pm) with a sell rate of $1/kWh, but the battery only discharges to grid at a very low rate when it is also discharging to the load ("Discharge battery only when load exceeds system output" is unchecked). Is this the desired behavior?

image

[batt-to-grid-test.zip](https://github.com/user-attachments/files/19457301/batt-to-grid-test.zip)

brtietz:
@cpaulgilman That's a good test - thanks for including it. This is not the intended behavior and highlights a bug in the lib_utility_rate_equations rate_data::get_composite_tou_sell_rate function - where these sell rates are being interpreted as 0 even though they shouldn't be. I'll push a solution to this branch so the battery discharges with the high sell rate you've specified in the case.

sjanzou:
The following four tests are now failing on Ubuntu

[  FAILED  ] 4 tests, listed below:
[  FAILED  ] lib_utility_rate_test.test_tiered_tou_cost_estimates
[  FAILED  ] lib_utility_rate_test.test_tiered_sell_rates
[  FAILED  ] lib_utility_rate_test.test_simple_demand_charges
[  FAILED  ] lib_utility_rate_test.test_demand_monthly_tou_demand_charges

cpaulgilman:
> I'll push a solution to this branch so the battery discharges with the high sell rate you've specified in the case.

Thanks!

cpaulgilman (approved):

shared/lib_battery_dispatch.cpp
```diff
@@ -1153,3 +1153,12 @@ bool byLowestMarginalCost::operator() (grid_point const& a, grid_point const& b)
     return a.MarginalCost() < b.MarginalCost();
 
 }
+
+bool byExportPerKWh::operator() (grid_point const& a, grid_point const& b)
+{
+    if (a.ExportPerKWh() == b.ExportPerKWh())
+    {
+        return a.ExportPerKWh() > b.ExportPerKWh();
```
sjanzou, line 1161: @brtietz , can you explain the logic here? If the two grid points are equal then return false - always?

sjanzou (commented):
Are we holding off merging this until the Mac tolerance is at a value for the tests to pass?

sjanzou (approved):
With latest commit all tests (except landbosse - since I do not have installed for release build) are passing on windows - not sure why the Windows runners are failing...

image

dguittet:
Trying to fix windows runner... bash commands on the Windows bash shell seem to not always be the same as for unix...

shared/lib_battery_dispatch.cpp
brtietz: This is a typo. I corrected the tiebreaker to ExportPrice (total paid by utility to discharge at full power) Good catch!

cpaulgilman:
PR reviewed for Help revisions -- I made a few minor edits, but no major revisions required.

PR #1304, opened 2025-03-27
Throw error whenever cumulative energy exceeds maximum tier usage for a given period.

sjanzou:
Addressed SAM issue https://github.com/NREL/SAM/issues/1834

Throws error message rather than producing erroneous results.

To test follow the steps:
Load a behind the meter model
Set the compensation for distributed generation to net billing
Set a maximum tier to 100 kWh (or something lower than the month's usage
Run simulation
The bill savings are negative

The following file can be used to test a valid case and a case with an invalid maximum usage for the maximum tier.

[SAM_1834.zip](https://github.com/user-attachments/files/19483818/SAM_1834.zip)

### DIFF
--- a/ssc/cmod_utilityrate5.cpp
+++ b/ssc/cmod_utilityrate5.cpp
@@ -2366,6 +2366,16 @@ public:
 
 
 								// cumulative energy used to determine tier for credit of entire surplus amount
+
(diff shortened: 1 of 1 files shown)


brtietz (approved):
This works well! I additionally tested a two tiered rate to make sure it still worked, and I got the expected errors.

PR #1327, opened 2025-05-21
ssc test speed up

sjanzou:
Low hanging fruit in the Google Doc completed.

The following results from running locally on Ubuntu machine with debug build for coverage results
Coverage reduced by 43 lines 63,733 to 63,690 (same percentage 54.7%)
Tests reduced from 776 to 690
Time reduced from 4,555.25s (75.92m) to 2,239.21s (37.32m)
Gory details at https://github.com/NREL/SAM-documentation/blob/main/Unit%20Testing/ssc%20test%20speedup/SSC%20Test%20speedup.docx

### DIFF
--- a/test/input_cases/geothermal_common_data.h
+++ b/test/input_cases/geothermal_common_data.h
@@ -58,7 +58,7 @@ static void geothermal_singleowner_default(ssc_data_t &data)
 	ssc_data_set_number(data, "resource_temp", 200);
 	ssc_data_set_number(data, "resource_depth", 2000);
     ssc_data_set_number(data, "geotherm.cost.inj_prod_well_ratio", 50);
-	ssc_data_set_number(data, "geothermal_analysis_period", 25);
+	ssc_data_set_number(data, "geothermal_analysis_period", 2);
 	ssc_data_set_number(data, "model_choice", 0);
 	ssc_data_set_number(data, "specified_pump_work_amount", 0);
 	ssc_data_set_number(data, "nameplate", 30000);
@@ -145,7 +145,7 @@ static void geothermal_singleowner_default(ssc_data_t &data)
 	set_array(data, "grid_curtailment", geothermal_curtailment_path, 8760);
 	ssc_data_set_number(data, "grid_interconnection_limit_kwac", 100000);
 	
-	ssc_data_set_number(data, "analysis_period", 25);
+	ssc_data_set_number(data, "analysis_period", 2);
 
 	ssc_number_t p_federal_tax_rate[1] = { 21 };
 	ssc_data_set_array(data, "federal_tax_rate", p_federal_tax_rate, 1);
--- a/test/input_cases/geothermal_common_data.h
+++ b/test/input_cases/geothermal_common_data.h
@@ -145,8 +145,7 @@ static void geothermal_singleowner_default(ssc_data_t &data)
 	set_array(data, "grid_curtailment", geothermal_curtailment_path, 8760);
 	ssc_data_set_number(data, "grid_interconnection_limit_kwac", 100000);
 	
-//	ssc_data_set_number(data, "analysis_period", 25);
-    ssc_data_set_number(data, "analysis_period", 2); // Speedy Gonzales 
+	ssc_data_set_number(data, "analysis_period", 25);
 
 	ssc_number_t p_federal_tax_rate[1] = { 21 };
 	ssc_data_set_array(data, "federal_tax_rate", p_federal_tax_rate, 1);
--- a/test/input_cases/geothermal_common_data.h
+++ b/test/input_cases/geothermal_common_data.h
@@ -145,7 +145,9 @@ static void geothermal_singleowner_default(ssc_data_t &data)
 	set_array(data, "grid_curtailment", geothermal_curtailment_path, 8760);
 	ssc_data_set_number(data, "grid_interconnection_limit_kwac", 100000);
 	
-	ssc_data_set_number(data, "analysis_period", 25);
+//	ssc_data_set_number(data, "analysis_period", 25);
+    ssc_data_set_number(data, "analysis_period", 2); // Speedy Gonzales 
+
 	ssc_number_t p_federal_tax_rate[1] = { 21 };
 	ssc_data_set_array(data, "federal_tax_rate", p_federal_tax_rate, 1);
 	ssc_number_t p_state_tax_rate[1] = { 7 };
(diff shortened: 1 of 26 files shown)


janinefreeman:
@tyneises @cpaulgilman @brtietz @dguittet @mjprilliman We discussed having all of us review on an extended timeline. Please review the **text** changes for technologies relevant to you in the test changes, and approve the PR when you've done so, but we do not all need to actually run the tests locally. Please aim for COB 5/30 for review. @cpaulgilman could you please be the one to run the actual test suite locally on your machine and make sure they pass?

tyneises (approved):

dguittet (approved):
Awesome such simple changes made a big difference. Please remove all the comments though

test/ssc_test/cmod_pvsamv1_test.cpp
```diff
@@ -666,7 +670,7 @@ TEST_F(CMPvsamv1PowerIntegration_cmod_pvsamv1, NoFinancialModelSystemDesign)
                                 192329, 187710, 185732, 197565, 201021, 192329, 187710
     };
 
-    for (size_t i = 0; i != annual_energy_expected.size(); i++)
+    for (size_t i = 57; i < annual_energy_expected.size(); i++)
```
brtietz: I'm looking at coveralls and trying to find the coverage decrease, and given that the mismatch losses are no longer covered I think this a likely candidate:

image

Mismatch losses appear to be case i = 13. Can you:

(a) Edit the arrays above to only include the cases we want to keep? 
(b) Ensure mismatch losses are included in those cases?

brtietz (commented):
Looks good overall, this is a very impressive speedup! I requested one change where I think the coverage loss is significant.

test/input_cases/geothermal_common_data.h
```diff
@@ -146,6 +146,7 @@ static void geothermal_singleowner_default(ssc_data_t &data)
 	ssc_data_set_number(data, "grid_interconnection_limit_kwac", 100000);
 	
 	ssc_data_set_number(data, "analysis_period", 25);
+
```
janinefreeman, line 149: Was there a reason this one is still set to 25? Maybe doesn't actually really affect test time if geothermal doesn't run in lifetime mode, but worth a comment if that's the case

janinefreeman (approved):
Thanks Steve! Only one comment from me.

test/ssc_test/cmod_pvsamv1_test.cpp
```diff
@@ -145,11 +145,13 @@ TEST_F(CMPvsamv1PowerIntegration_cmod_pvsamv1, DefaultResidentialModel)
 
         ssc_number_t lcoe_nom;
         ssc_data_get_number(data, "lcoe_nom", &lcoe_nom);
-        EXPECT_NEAR(lcoe_nom, 7.03, m_error_tolerance_lo) << "Levelized COE (nominal)";
+//        EXPECT_NEAR(lcoe_nom, 7.03, m_error_tolerance_lo) << "Levelized COE (nominal)";
+        EXPECT_NEAR(lcoe_nom, -26.45, m_error_tolerance_lo) << "Levelized COE (nominal)"; // Speedy Gonzales 25yr to 2yr
```
mjprilliman: Negative LCOE? Might want to look at defaults for 2 year case, just to keep a meaningful value to check in the future. For 2 year runs, I set itc_fed_percent and itc_sta_percent to = [0] in pvsamv1_common_data.cpp and got a more reasonable lcoe (but a negative npv).

test/ssc_test/cmod_pvsamv1_test.cpp
```diff
@@ -145,11 +145,13 @@ TEST_F(CMPvsamv1PowerIntegration_cmod_pvsamv1, DefaultResidentialModel)
...
 
         ssc_number_t lcoe_real;
         ssc_data_get_number(data, "lcoe_real", &lcoe_real);
-        EXPECT_NEAR(lcoe_real, 5.65, m_error_tolerance_lo) << "Levelized COE (real)";
+//        EXPECT_NEAR(lcoe_real, 5.65, m_error_tolerance_lo) << "Levelized COE (real)";
+        EXPECT_NEAR(lcoe_real, -25.50, m_error_tolerance_lo) << "Levelized COE (real)"; // Speedy Gonzales 25yr to 2yr
```
mjprilliman: Same comment as above

test/ssc_test/cmod_singleowner_test.cpp
```diff
@@ -102,7 +102,7 @@ TEST_F(CmodSingleOwnerTest, ETES) {
 
     Test("singleowner", file_inputs, file_outputs, compare_number_variables, compare_array_variables);
 }
-
+/*
 TEST_F(CmodSingleOwnerTest, PV) {
```
mjprilliman: As PV-SingleOwner is likely one of the most modeled technologies, I might consider keeping this one and commenting out others. Although I see PV Battery is still being run, so there may be enough coverage already. Same thought for PVWatts-SingleOwner

mjprilliman (approved):
Just a few nitpicks, thanks for the great work on this!

test/ssc_test/cmod_pvsamv1_test.cpp
sjanzou: lcoe changed from -26.45 to 12.05 with the itc_fed_percent = 0 and itc_sta_percent = 0 in latest commit

test/ssc_test/cmod_pvsamv1_test.cpp
sjanzou: lcoe changed from -25.5 to 11.61 with the itc_fed_percent = 0 and itc_sta_percent = 0 in latest commit

npv changed from 5685.3 to -358.2

test/ssc_test/cmod_singleowner_test.cpp
sjanzou: The ETES test was kept because it uses an additional o and m value that the others single owner tests do not. Same with the battery and fuel cell. The other tests were selected based on inputs to single owner compute module being changed. The PV and PVwatts single owner tests were added back (no coverage change and only an additional 40ms added)

test/ssc_test/cmod_pvsamv1_test.cpp
sjanzou: Great catch - running i=13 gives the same coverage as running all 64 configurations

base run (no changes from the i=57 to annual_energy_expected.size()
image

i=13 results - coverage increased and time reduced
image

enabling all tests does not increase coverage and a 2.5 times increase in elapsed time
image

Updated in latest commit and commented code to be removed after one additional test

brtietz (approved):
Thanks for updating the pvsamv1 tests!

test/input_cases/geothermal_common_data.h
sjanzou: @janinefreeman, thanks for the feedback - originally left at 25 because the speed impact was low
image

Updated to 2 years for consistency. However, this caused a failure in the cmod_grid 
image

The issue was when the "analysis_period" is changed from 25 to 2, then the "geothermal_analysis_period" has to also be changed from 25 to 2
image

Then everything runs
image

sjanzou:
Latest coverage results from local Ubuntu VM
image

PR #1328, opened 2025-05-22
Ci updates

dguittet:
Compare Test Time Elapsed workflow is currently failing during the `get_workflow_artifact_branch` step. The list of artifacts being returned has a limited length and if the required artifact is older then it may not show up in the list. Add a loop to step through all the pages of the artifacts from the API until one is found.

Also, Compare Test Time Elapsed can fail if the Google Test run didn't complete successfully to produce a test time log. So first check that it was successful.

### DIFF
--- a/.github/workflows/test_times.yml
+++ b/.github/workflows/test_times.yml
@@ -55,6 +55,12 @@ jobs:
       run: |
         echo "WORKFLOW_ID=${{ steps.workflow_id.outputs.result }}" >> $GITHUB_ENV
 
+    - name: Check Test-Running Workflows were Successfully Completed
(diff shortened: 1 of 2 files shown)


dguittet:
@sjanzou This should fix the issue you brought up yesterday

sjanzou:
> @sjanzou This should fix the issue you brought up yesterday

Awesome, thanks for updating!

There seems to be a file missing with Ty's latest commit https://github.com/NREL/ssc/actions/runs/15198270315/job/42747343063

image

Maybe we need to get everyone to merge in the latest patch branch with the fix to their respective feature branches

PR #1331, opened 2025-06-02
Fix overflow in Detailed PV Heat Transfer Method when using array dimensions

berg-michael:
### Description

This PR fixes an overflow that led to incorrect results when using the detailed PV Heat Transfer Method when using array dimensions.

The root cause of the error was that the Length and Width parameters of the mcsp_celltemp_t class were being scaled by Nrows, Ncols on each iteration. The intended behavior was to scale them a single time. 

A secondary issue was that the result of the temperature model was not checked. The temperature model takes in a temperature variable by reference and modifies it upon return. In cases where the temperature model fails (including the overflow condition above), the function fails and returns false. Because this return is not checked, and because the temperature variable is initialized to ambient temp, the error state is not detected. This results in the behavior described in [nrel/SAM #2081](https://github.com/NREL/SAM/issues/2081).

This PR should solve these issues. 

1. It removes all modifications of class variables, and instead modifies local copies. This removes the overflow.
2. It adds additional error checking in cmod_pvsamv1.cpp.
3. It adds additional error messages in the mcsp_celltemp_t operator() function. 

One additional change I made is const-qualifying the pvcelltemp_t and pvmodule_t operator() methods, and cosnt-qualifying additional methods that are downstream of these operator() methods and therefore must be const as well. This required no changes in code besides those I already made in mcsp_celltemp_t to remove the overflow. Const-qualifying these methods should prevent errors of this class from arising in the future. 

To test, follow @cpaulgilman's steps to reproduce the bug from the initial issue:

> Steps to reproduce the behavior:
> 1. Create a default Detailed PV / No Financial case.
> 2. On the Module page, choose **Heat transfer method** for the temperature correction option.
> 3. Choose **Array Dimensions** for the heat transfer dimensions option.
> 4. Run a simulation.
> 5. On the Data Tables tab of the Results page, display "Subarray 1 Cell temperature (C)" and "Weather file ambient temperature (C)". Note that starting the last daylight time step of January 16, the cell temperature and ambient temperature are the same.

**Observe that now the cell temperature does not revert to the ambient temperature on January 16**. 

image

Fixes # (issue(s)):  [nrel/SAM #2081](https://github.com/NREL/SAM/issues/2081)

This PR is an SSC only PR. There are no specific branches of other repositories that this PR depends upon. 

### Unit Test Impact:

No tests were changed. If tests exist which assumed the correctness of the detailed PV heat transfer method and used array dimensions, the output of those tests may now change. 

### Checklist
- [ ] requires help revision and I added that label
- [ ] changes defaults

### DIFF
--- a/shared/lib_cec6par.cpp
+++ b/shared/lib_cec6par.cpp
@@ -216,7 +216,7 @@ bool cec6par_module_t::operator() ( pvinput_t &input, double TcellC, double opvo
  *********************************************************************************************
  *********************************************************************************************/
 
-bool noct_celltemp_t::operator() ( pvinput_t &input, pvmodule_t &module, double , double &Tcell )
+bool noct_celltemp_t::operator() ( pvinput_t &input, pvmodule_t &module, double , double &Tcell ) const
 {
 	double G_total, Geff_total;
 	double tau_al = std::abs(TauAlpha);
@@ -362,7 +362,7 @@ static double channel_free_194( double W_gap, double SLOPE, double TA, double T_
 	return  Nu*k_air/W_gap;
 }
 
-bool mcsp_celltemp_t::operator() ( pvinput_t &input, pvmodule_t &module, double opvoltage, double &Tcell ) 
+bool mcsp_celltemp_t::operator() ( pvinput_t  &input, pvmodule_t &module, double opvoltage, double &Tcell ) const
 {	
     m_err = "Populating a default error message";
 	if ( input.Ibeam + input.Idiff + input.Ignd < 1 )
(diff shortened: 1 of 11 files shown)


brtietz (approved):
Looks good from my perspective, thanks for cleaning up all of the 'const' definitions!

I think @mjprilliman should potentially take a look at this as he's more familiar with these models.

cpaulgilman (approved):
This looks good to me. The cell temperature values look more reasonable, i.e., not equal to ambient temperature. And, comparing results for "module dimensions" to "array dimensions," the latter temperatures are slightly higher, which is what I expect intuitively.

I agree it would be good for @mjprilliman to take a look.

mjprilliman (approved):

PR #1335, opened 2025-07-03
SSC: Add builtin snow depth array, support varying periods for this array. 

berg-michael:
### Description

The PR enables the builtin snow depth array to which the nohrsc_snow_data branch of SAM writes to. It allows the user to choose which source of snow data they wish to use: weather file provided or builtin array. 

If the snow data is provided via the weather file, as before, there was an implicit guarantee that the snow depth data was of the same period as the weather file. This guarantee is lost when using a user-input array. 

Thus, this PR also includes code the verify that the snow depth array can be aligned with the weather file data. Periods of 5 minutes and 10 minutes, for example are compatible, whereas periods of 5 and 12 minutes are not because there are timesteps that will not line up. In the case of incompatible periods, the simulation fails. In the case of matching periods, the simulation succeeds without any massaging of the snow depth data. In the case of non-matching but compatible periods, the snow data will be up- or down-sampled to align with the period of the weather file data. 

To review, build against the nohrsc_snow_data branch of sam. Verify that snow data input into the array is used and has behavior that aligns with entering snow data into the weather file via the snow macro. 

branch:[ sam/nohrsc_snow_data](https://github.com/NREL/SAM/tree/nohrsc_snow_data)

https://github.com/NREL/SAM/pull/2089
https://github.com/NREL/SAM-private/pull/128

### Unit Test Impact:

New tests written to exercise the snow array. They check:
- using the snow_array compared to weather file data produces the same results
- Upsampling the snow_array to match with a weather file with shorter period
- Downsampling the snow_array to  

### Checklist
- [x] requires help revision and I added that label
- [x] adds, removes, modifies, or deletes variables in existing compute modules
- [x] changes defaults (the corresponding SAM PR adds them)


berg-michael:
Recent commit fixes the issue where PV + Financial Model simulations led to errors.

cpaulgilman (approved):

sjanzou (approved):
Working and approved if SAM 2089 GitHub actions are working

PR #1339, opened 2025-07-09
Update to use the same Windows runner as the SAM repo

brtietz:


sjanzou (approved):
Thanks for updating to a non-retired runner!

PR #1345, opened 2025-08-04
Fixes CI test failures and updates GTest reference

sjanzou:
Addresses intermittent CI test failures and improves test reliability:

- Adds pipe exit code to properly catch googletest failures
- Modifies the `general_error` constructor to use `std::move` for efficiency.
- Corrects an issue in the `UserArraySnowModel` test related to tilt angle and expected results, ensuring the test passes consistently.


brtietz (approved):
Thanks for cleaning these up!

PR #1351, opened 2025-08-21
Update release.yml

dguittet:
Update release.yml to include more description


brtietz (approved):
Thanks for updating the explanation!

PR #1362, opened 2025-11-14
Battery Charging Bug

### LINKED ISSUE #1361: Issue in Battery
**Describe the bug**
The `calculate_current_for_power_kw` function reduces the battery's target power and current to what is available given SOC. There's a bug in the charging portion where the reduced power and current should have been negative. 

Thanks to @jaredthomas68 for tracking this down

dguittet:
The calculate_current_for_power_kw function reduces the battery's target power and current to what is available given SOC. Fix the bug in the charging portion where the reduced power and current should have been negative.

Fix #1361


brtietz:
Thanks for looking into this! A few requests:

1. Can you add a test that establishes this sign convention explicitly? There are a few tests in lib_resilience_test that are close, but I'm getting lost with multiple negative sign conventions. Bonus points for establishing the convention in the discharging function as well.
2. Can you review other uses of the function to ensure the sign convention is correct? I'm seeing some additional inconsistencies in the below.

image

dguittet:
Thanks Brian for the good catch. Turns out it was just a few edge cases here and there where `calculate_max_discharge_kw` and `calculate_max_charge_kw` could return values that were outside of their expected bounds so I added limit checks and removed the erroneous sign change in `calculate_current_for_power_kw`

brtietz (approved):
Thanks for incorporating the fixes!

PR #1366, opened 2025-12-12
Lengthen analysis period and max loan term variables

brtietz:
### Description

Change ssc vartable limits to allow analysis periods of up to 100 years. 

Fixes # https://github.com/NREL/SAM/issues/2140

The SAM GUI has several seconds of "not responding" when attempting to shorten a long analysis period, but this also occurs on the released version (say changing 50 to 45) so I haven't attempted to address it here.

Develop of wex, lk, sam-private

SAM PR https://github.com/NREL/SAM/pull/2139 resolves some related callback issues that make this easier to test, but these are technically independent.

### Unit Test Impact:

Open to suggestions on unit test impact

### Checklist
- [x] requires help revision and I added that label
- [ ] changes defaults
- [x] I've tagged this PR to a milestone

### DIFF
--- a/ssc/common.cpp
+++ b/ssc/common.cpp
@@ -40,7 +40,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "lib_util.h"
 
 var_info vtab_standard_financial[] = {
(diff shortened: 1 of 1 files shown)


mjprilliman (approved):
Works with https://github.com/NREL/SAM/pull/2139 checked out.

PR #1367, opened 2026-02-06
OR Tools

sjanzou:
https://github.com/NatLabRockies/SAM-private/pull/140


.github/workflows/ci.yml
```diff
@@ -49,27 +61,23 @@ jobs:
           export 
           mkdir ${GTEST}/build
           cd ${GTEST}/build
-          cmake -DCMAKE_CXX_FLAGS=-std=c++11 ..
+          cmake -DCMAKE_CXX_FLAGS=-std=c++17 ..
```
brtietz, line 64: Flagging another c++17 - Is this intentional for gtest?

CMakeLists.txt
```diff
@@ -25,6 +37,16 @@ if(APPLE)
     set(CMAKE_OSX_DEPLOYMENT_TARGET "12" CACHE STRING "Minimum OS X deployment version")
 endif()
 
+if (UNIX AND NOT CMAKE_C_COMPILER)
+    set(CMAKE_C_COMPILER gcc)
+    set(CMAKE_CXX_COMPILER g++)
+endif()
+if(MSVC)
+    set(CMAKE_CXX_STANDARD 20)
```
brtietz: Having this inside of an if statement feels off to me. Mind adding a comment as to why we have different versions by platform?

brtietz (changes requested):
Please ensure that new files get the appropriate copyright notices. Are lib_ortools.h/cpp and lob_ptes_chp_dispatch NLR code or third party?

Additionally, please update the README to reflect what libraries we're using now. (ortools, did anything get removed?)

CMakeLists.txt
sjanzou: Thanks for flagging the inconsistency.

Based on OR Tools v9.14, all CMakeLists.txt have been updated to match
if(MSVC)
set(CMAKE_CXX_STANDARD 20)
else()
set(CMAKE_CXX_STANDARD 17)
endif()

Note that Kestrel (maybe still on RHEL 8.8 - can someone verify?) supports c++17 only.
image

image



Also, note EOL for RHEL
https://access.redhat.com/support/policy/updates/errata#:~:text=8.6%20(ends%20May%2031%2C%202026),Maintenance%20Support%20Phase)%20policy%20above.

brtietz (approved):

shared/lib_ortools.cpp
```diff
@@ -0,0 +1,59 @@
+#include "ortools/linear_solver/linear_solver.h"
```
mjprilliman, line 1: Is this file supposed to be in the final build? I just built with the latest instructions and it wasn't included.

PR #1368, opened 2026-02-13
Alliance name change update

mjprilliman:
### Description

License update for Alliance name change


brtietz (approved):
Thanks for updating!

PR #1370, opened 2026-03-09
Split ITC basis by tech part 1a: refactor single owner

brtietz:
### Description

This is the first PR in the process of updating the ITC and depreciation bases to account for OBBBA and split things by technology. This PR adds unit tests, but is otherwise just a refactor of the other code.

The near term goal is to put ITC and depreciation calculations in a class in common financial to share code between financial models prior to making further changes to behavior. Single Owner has been refactored using the new code. The other financial models have had the depreciation code extracted from their IRR target & DSCR loop to show that this code is independent of those loops (this is ["new" circa 2024](https://github.com/NatLabRockies/SAM/issues/1803) - older versions of SAM this code needed to be in the loop).

Questions for this PR include:
- Do the new variable and class names make sense?
- Do the function scope and class scope make sense?
- Is there anything in the wrong spot (e.g. should be in the new class, but is in SO or vice versa)?
- Do the unit tests give you confidence that the refactor didn't break anything?
- Any other notes before applying this to other financial models?

Happy to discuss merge strategy. (merging this to develop shouldn't break anything, but there may be a stage at which the merge makes more sense than now)

Develop for other branches

### Unit Test Impact:

Added JSON financial model unit tests that look at the combination of IRR target mode and dscr mode.

### Checklist
- [x] I've tagged this PR to a milestone


brtietz:
Tests added in the latest round are below. These were prioritized by commonly used features for all financial models, and less commonly used features only in single owner. Results are based off of develop as of this morning.

**All Relevant financial models**
- Fixed debt percentage with and without IRR target
- PBI for debt service
- Bonus depreciation

**Only Single Owner**
- Custom depreciation
- Loan moratorium
- Receivables Reserve
- Major equipment deprecation besides 5-year MACRS
- IBI/CBI (compliments host developer)

Additionally, added a test for Single Owner Heat.

test/ssc_test/cmod_singleowner_test.cpp
```diff
@@ -133,3 +133,113 @@ TEST_F(CmodSingleOwnerTest, PVWatts) {
...
+    file_inputs += "/test/input_json/FinancialModels/singleowner/so_pbi_dscr_targets_PVWatts_Single_Owner_cmod_singleowner.json";
+    std::string file_outputs = SSCDIR;
+    file_outputs += "/test/input_json/FinancialModels/singleowner/so_pbi_dscr_targets_PVWatts_Single_Owner_cmod_singleowner_outputs.json";
+    std::vector<std::string> compare_number_variables = { "ppa", "project_return_aftertax_npv", "lcoe_real", "lppa_nom", "min_dscr", "project_return_aftertax_irr" };
+    std::vector<std::string> compare_array_variables = { "cf_project_return_aftertax", "cf_annual_costs", "cf_pretax_dscr", "cf_debt_balance" };
+
+    Test("singleowner", file_inputs, file_outputs, compare_number_variables, compare_array_variables);
+}
+
+TEST_F(CmodSingleOwnerTest, CustomDepr) {
```
sjanzou, line 192: Thanks for adding all of these!

sjanzou (approved):
Great refactoring! The cmod_singleowner is much cleaner and the tests seem to cover everything and match with develop (before changes. 

As far as GitHub Actions - I think the failing Windows test CMPvsamv1PowerIntegration_cmod_pvsamv1.UserArraySnowModel should be made in a separate pull request as this one continues to pass/fail on Windows intermittently. Works on other runners consistently.

I also, went through and updated the SAM itc_basis_updates branch to work with the ortools merge and updated defaults and results and went through to check for other tests to add for cmod_singleowner - I did not find any.

Great work - very meticulous!

brtietz:
> CMPvsamv1PowerIntegration_cmod_pvsamv1.UserArraySnowModel

Thanks for the review! I created an issue for this here: https://github.com/NatLabRockies/ssc/issues/1372

PR #1376, opened 2026-03-23
Split ITC basis by tech part 1b: Refactor other financial modules

brtietz:
Apply https://github.com/NatLabRockies/ssc/pull/1370 to all of the financial models that use the same structure.

No behavior should change with this PR, this is refactor-only.

I'll admit that my ordering conventions for CBI/IBI and reserves were a bit inconsistent. If this is bothersome I can put in a little work to standardize.

Next step will be adding SSC inputs and allowing for the new basis calcs.

After that this structure will be applied to other compute modules, like cashloan. (if we still think that's a good idea)


sjanzou (approved):
Looks good and the tests you added from #1370 for the other financial models are still passing. 

It would be good to standardize the CBI/IBI and reserves ordering cointerventions; however, that can be another refactoring when time (and funding) permits.

The failing GitHub Actions Windows runner should be fixed when #1377 is merged.

PR #1386, opened 2026-04-11
Ssc 1300 wind lifetime mode

brtietz:
### Description

Adds lifetime mode to windpower. In addition to the obvious for loops, required updating the data provider with a rewind function. This reads in the file at the start of the simulation and references arrays, more similar to PV. AI disclosure: Claude Sonnet assisted with the rewind function.

Note that the weather file mode for wind is slower by a linear factor of analysis period length. Open to changing that prior to merging if we prefer a different approach to applying degradation. This would impact year 1 vs lifetime outputs.

Fixes https://github.com/NatLabRockies/ssc/issues/1300

SAM PR tbd. Develop of other repos.

### Unit Test Impact:

New wind lifetime tests. No other changes expected

### Checklist
- [ ] requires help revision and I added that label
- [x] changes defaults

### Reminders- this section can be deleted
[Checking for PySAM Incompatible API Changes]
(https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files).

[When do the PySAM files need to be regenerated?]
(https://github.com/NREL/SAM/wiki/PySAM-Incompatible-API-Changes-&-Regenerating-PySAM-Files#when-do-the-pysam-files-need-to-be-regenerated-via-export_config)

### DIFF
--- a/shared/lib_battery.cpp
+++ b/shared/lib_battery.cpp
@@ -537,6 +537,8 @@ double battery_t::calculate_max_charge_kw(double *max_current_A) {
         thermal->updateTemperature(current, state->last_idx + 1);
         qmax = capacity->qmax() * thermal->capacity_percent() * 0.01 * SOC_ratio;
     }
+    current = std::min(0.0, current);
+    power_W = std::min(0.0, power_W);
     if (max_current_A)
         *max_current_A = current;
     *thermal->state = thermal_initial;
@@ -556,6 +558,8 @@ double battery_t::calculate_max_discharge_kw(double *max_current_A) {
         thermal->updateTemperature(current, state->last_idx + 1);
         qmax = capacity->qmax() * thermal->capacity_percent()  * 0.01;
     }
+    current = std::max(0.0, current);
+    power_W = std::max(0.0, power_W);
     if (max_current_A)
         *max_current_A = current;
     *thermal->state = thermal_initial;
--- a/shared/lib_pv_io_manager.cpp
+++ b/shared/lib_pv_io_manager.cpp
@@ -140,7 +140,7 @@ Irradiance_IO::Irradiance_IO(compute_module* cm, std::string cmName)
         if (weatherDataProvider->has_message()) cm->log(weatherDataProvider->message(), SSC_WARNING);
     }
     else {
-        throw exec_error(cmName, "No weather data supplied");
+        throw exec_error(cmName, "No weather data supplied.");
     }
 
     // assumes instantaneous values, unless hourly file with no minute column specified
@@ -166,7 +166,7 @@ Irradiance_IO::Irradiance_IO(compute_module* cm, std::string cmName)
         tsShiftHours = 0.5;
     }
     else
-        throw exec_error(cmName, "subhourly and non-annual weather files must specify the minute for each record");
+        throw exec_error(cmName, "Subhourly and non-annual weather files must specify the minute for each record.");
 
     weatherDataProvider->header(&weatherHeader);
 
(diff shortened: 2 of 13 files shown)


cpaulgilman (changes requested):
Degradation does not appear to be applied correctly.

Results for default Wind Power / Single Owner case with 5% AC degradation rate lifetime simulation:

Image

Compare to SAM 2025.4.15 r1 single year simulation:

Image

brtietz:
> Degradation does not appear to be applied correctly.

Thanks for flagging this. I grabbed the code from PVWatts, which applies the degradation linearly: https://github.com/NatLabRockies/ssc/blob/2ffaf5715758084b6d9118de1272577162869df0/ssc/cmod_windpower.cpp#L378

The code in the released version applies degradation exponentially: https://github.com/NatLabRockies/ssc/blob/2ffaf5715758084b6d9118de1272577162869df0/ssc/cmod_singleowner.cpp#L1155

Looks like I used the exponential convention for custom generation profile as well: https://github.com/NatLabRockies/ssc/blob/2ffaf5715758084b6d9118de1272577162869df0/ssc/cmod_custom_generation.cpp#L138

Just to confirm/establish a convention here: DC degradation for PV is linear, but for all other models AC degradation is exponential. Is that what we want? (if so I'll want to adjust some version upgrade messaging around the PVWatts changes for this version)

brtietz:
I pulled up some papers to help inform this decision. The curves in https://www.sciencedirect.com/science/article/pii/S0360544225017840 and https://papers.phmsociety.org/index.php/phme/article/download/3328/1939 look pretty linear. ATB assumes that O&M will counteract degradation, which is unhelpful here. 

@janinefreeman is there an NLR wind person you'd recommend asking about this?

janinefreeman:
I've messaged a wind person to ask for the right name and I'll send you a Teams message. It does seem weird to me that DC degradation would be linear but it also rings a bell, is there an old issue that might illuminate that decision somewhere?

brtietz:
> I've messaged a wind person to ask for the right name and I'll send you a Teams message. It does seem weird to me that DC degradation would be linear but it also rings a bell, is there an old issue that might illuminate that decision somewhere?

Thanks! See https://github.com/NatLabRockies/ssc/issues/503 for some of the PV background.

cpaulgilman:
> > I've messaged a wind person to ask for the right name and I'll send you a Teams message. It does seem weird to me that DC degradation would be linear but it also rings a bell, is there an old issue that might illuminate that decision somewhere?
> 
> Thanks! See #503 for some of the PV background.

Regarding PV degradation, I did a quick search on the PVPMC Modeling Steps and pvlib websites and did not see mention of annual DC degradation there.

brtietz:
https://www.sciencedirect.com/science/article/pii/S0960148113005727#fig9 is a good citation for linear degradation for wind. The recent changes add a page to explain linear versus compounding by model using AC degradation.

I've also updated startup.lk to get all of the hybrids in lifetime mode, which should solve the issues caught in https://github.com/NatLabRockies/SAM/pull/2166

Outstanding issues include:

1. Version upgrade doesn't work for hybrids. In the 5 tech configuration, custom generation profile degradation makes it through the upgrade, but PVWatts and wind keep their defaults
2. The wind default degradation has some annoying floating point issues. It should be 0.64 but is just below that.

Any suggestions on those two prior to a more formal re-review?

mjprilliman (approved):

cpaulgilman:
I found a couple more instances of `generic_degradation` in SSC (none in SAM/SAM_api):

### [\ssc\test\input_cases\general_data\pvsmoothing_Generic_alloptions.json](https://github.com/NatLabRockies/ssc/blob/ssc_1300_wind_lifetime_mode/test/input_cases/general_data/pvsmoothing_Generic_alloptions.json) 
* Line  11: `"generic_degradation" : [ 0 ]`

### [\ssc\ssc\cmod_generic_system.cpp](https://github.com/NatLabRockies/ssc/blob/ssc_1300_wind_lifetime_mode/ssc/cmod_generic_system.cpp)
* Line  66: `{ SSC_INPUT, SSC_ARRAY, "generic_degradation", "Annual AC degradation", "%/year",   "", "Lifetime", "system_use_lifetime_output=1",  "", "" }`
* Line 145: degrad = `as_array("generic_degradation", &count_degrad);`

cpaulgilman (approved):

PR #1396, opened 2026-05-11
Add net power control option for MSPT

tyneises:
This pull request merges in-progress code to develop CSP hybrid models in order to add Beta version of new "net power" control option for the MSPT model. The default original control method targets thermal power input to the power cycle. This approach is easier to converge and was maintained from early version of CSP annual timeseries modeling. The new Beta model controls HTF mass flow rate in order to achieve a system net power output, where net is the cycle output less cooling parasitics less other plant electricity consumption like pump and heliostat power. 
		
The net power control option is available on the System Design page. When it is selected, the model uses either the tabular or hourly target output fractions and applies it relative to the design point "estimated net output at design (nameplate)" to calculate the timestep target net power.
		
Currently, we hardcode heliostat and heater startup energy requirements to 0 in the mspt cmod if net power control mode is selected because it results in power spikes during short timesteps.

The pull request also brings in some minor bug fixes and improvements to plant control and convergence that are relevant for both controller modes.


taylorbrown75 (approved):
Branch builds on my machine and successfully targets the net power output in the default case.

qualand (approved):
Merged other net power branch (merge_net_power_control) which included updates to dispatch and a simple addition of PV within the salt tower cmod.

PR #1399, opened 2026-05-22
SDKtool CMakeLists capitalization error

sjanzou:
Fix CMake generation error on MacOS


cpaulgilman (approved):
Looks good. I searched for other instances of "SDKTool" but didn't find any that aren't in comments.

PR #1416, opened 2026-07-08
Fix links and update version in README.md

brtietz:
Post-release readme updates. Merge these to main, then pull through to patch and develop.


cpaulgilman (approved):
Looks good. Thanks.

Do we need to add OR-Tools to the list of dependencies?

PR #1419, opened 2026-07-17
Remove refs to gtest since we no longer need to pin to that version

brtietz:
Remove the pinned version of gtest to match coveralls to the main test script.


sjanzou (approved):
Looks good and macOS runners passing at this point.

PR #1420, opened 2026-07-21
Alma linux paths for the source built or-tools

brtietz:
Add eigen dependency directory so ssc builds with or-tools on alma linux

### DIFF
--- a/splinter/CMakeLists.txt
+++ b/splinter/CMakeLists.txt
@@ -4,7 +4,7 @@
 #
 #####################################################################################################################
 
-include_directories(. include src $ENV{ORTOOLSDIR}/include/eigen3 ../ssc)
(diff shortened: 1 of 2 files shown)


sjanzou (commented):
What do you have ORTOOLSDIR set to?

Did you run install after you built or-tools from source?

Image

The {ORTOOLSDIR}/include/eigen3 should have all the /_deps/eigen3-src items that you are missing

Image

brtietz:
It looks like the install command installed to /usr/local by default. Is there an alternative command that sets up the folder structure you have above?

brtietz:
Instructions that get the built or-tools installed to the appropriate directory:

image

Then set or-tools to the installed target. Closing since these are not needed with the (soon to be) updated build instructions.

